PR #12, opened 2021-11-04
Add Device pages to frontend

davidebriani:
Add first draft for:
- Devices page to list available devices
- Device page to display info about a device

Devices page:
Devices

Device page:
Device


frontend/src/components/Image.tsx
```diff
@@ -0,0 +1,37 @@
...
+        className="rounded border"
+        fluid
+        src={imageSrc}
+        onError={handleError}
+      />
+    </Figure>
+  );
+};
+
+export default Image;
```
Pavinati, line 37: Should we call this component `Figure` instead of `Image`? When using an image, you may not expect borders

frontend/src/pages/Devices.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const devicesData = usePreloadedQuery(GET_DEVICES_QUERY, getDevicesQuery);
+
+  // TODO: handle readonly type without mapping to mutable type
+  const devices = useMemo(
+    () => devicesData.devices.map((device) => ({ ...device, online: true })),
+    [devicesData]
+  );
+
+  return (
+    <div className="p-4">
```
Pavinati, line 40: Let's avoid bootsrap classes in pages.
In this specific instance i'd use a `Page` or `PageWrapper` component

frontend/src/pages/Devices.tsx
```diff
@@ -0,0 +1,100 @@
...
+          className="h-100 d-flex flex-column justify-content-center align-items-center"
+          data-testid="page-loading"
+        >
+          <Spinner />
+        </div>
+      }
+    >
+      <ErrorBoundary
+        FallbackComponent={(props) => (
+          <div className="h-100 d-flex flex-column justify-content-center align-items-center">
```
Pavinati, line 75: This will be reused often. We should make a component for that

frontend/src/pages/Device.tsx
```diff
@@ -16,19 +16,318 @@
...
-      <main className="mt-4"></main>
+      <main className="mt-4">
+        <Row>
+          <Col sm="4" lg="3">
+            <div>
+              <Image alt={device.name} />
+            </div>
+          </Col>
+          <Col sm="8" lg="9">
+            <Form className="ms-3">
```
Pavinati, line 241: bootstrap classes in pages

Pavinati (changes requested):

Pavinati (approved):

frontend/src/pages/Device.tsx
davidebriani: Rethinking this piece in terms of layout and components is not straightforward. We can deliver the code in this simple form now and let a good pattern emerge with future PRs

bettio (approved):

PR #39, opened 2021-11-10
Implement pages for hardware types

davidebriani:
Add forms and logic to create and update Hardware Types in the frontend.

https://user-images.githubusercontent.com/60540759/141147473-7a859ab4-7711-4adb-bd65-9e29ed3d4a1e.mp4

### DIFF
--- a/frontend/src/components/HardwareTypesTable.tsx
+++ b/frontend/src/components/HardwareTypesTable.tsx
@@ -1,0 +1,91 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2021 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import React from "react";
+import { FormattedMessage } from "react-intl";
+
+import Table from "components/Table";
+import type { Column } from "components/Table";
+import { Link, Route } from "Navigation";
+
+type HardwareTypeProps = {
+  id: string;
+  handle: string;
+  name: string;
+  partNumbers: string[];
+};
+
+const columns: Column<HardwareTypeProps>[] = [
+  {
+    accessor: "name",
+    Header: (
+      <FormattedMessage
+        id="components.HardwareTypesTable.nameTitle"
+        defaultMessage="Hardware Type Name"
+      />
+    ),
+    Cell: ({ row, value }) => (
+      <Link
+        route={Route.hardwareTypesEdit}
+        params={{ hardwareTypeId: row.original.id }}
+      >
+        {value}
+      </Link>
+    ),
+  },
+  {
+    accessor: "handle",
+    Header: (
+      <FormattedMessage
+        id="components.HardwareTypesTable.handleTitle"
+        defaultMessage="Handle"
+      />
+    ),
+    Cell: ({ value }) => <span className="text-nowrap">{value}</span>,
+  },
+  {
+    accessor: "partNumbers",
+    Header: (
+      <FormattedMessage
+        id="components.HardwareTypesTable.partNumbersTitle"
+        defaultMessage="Part Numbers"
+      />
+    ),
+    Cell: ({ value }) =>
+      value.map((partNumber, index) => (
+        <React.Fragment key={partNumber}>
+          {index > 0 && ", "}
+          <span className="text-nowrap">{partNumber}</span>
+        </React.Fragment>
+      )),
+    disableSortBy: true,
+  },
+];
+
+interface Props {
+  className?: string;
+  data: HardwareTypeProps[];
+}
+
+const HardwareTypesTable = ({ className, data }: Props) => {
+  return <Table className={className} columns={columns} data={data} />;
+};
+
+export type { HardwareTypeProps };
+
+export default HardwareTypesTable;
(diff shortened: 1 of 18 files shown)


frontend/src/forms/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  unique: {
+    id: "validation.unique",
+    defaultMessage: "Duplicate value.",
+  },
+  arrayMin: {
+    id: "validation.arrayMin",
+    defaultMessage: "Does not have enough values.",
+  },
+  partNumberFormat: {
+    id: "validation.partNumberFormat",
```
Pavinati, line 36: what about `validation.array.min` and `validation.partNumber.format`?
We might have multiple validations for arrays and common types (ex: `array.max` or `partNumber.alreadyUsed`)

frontend/src/forms/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    required: messages.required.id,
+  },
+  array: {
+    min: messages.arrayMin.id,
+  },
+});
+
+const hardwareTypeHandleSchema = yup
+  .string()
+  .matches(/^[a-z\d-]+$/, messages.partNumberFormat.id);
```
Pavinati, line 53: this only checks for the char-set used.
we might add an other match clause making sure it starts with a character and does not end with a hyphen?
Or maybe update the regex but i cannot think of a simple message to describe why the input is not valid
`/^[a-z][a-z\d-][a-z\d]$/`

frontend/src/pages/HardwareTypeCreate.tsx
```diff
@@ -16,11 +16,91 @@
...
+          } else {
+            navigate({ route: Route.hardwareTypes });
+          }
+        },
+        onError(error) {
+          setErrorFeedback(
+            <FormattedMessage
+              id="pages.HardwareTypeCreate.creationErrorFeedback"
+              defaultMessage="Could not create the hardware type, please try again."
+              description="Feedback for unknown creation error in the HardwareTypeCreate page"
```
Pavinati, line 81: i would drop descriptions for now, just to keep translations easy to mange once we think of how to index them across components and pages

frontend/src/forms/UpdateHardwareType.tsx
```diff
@@ -0,0 +1,248 @@
...
+          <Stack gap={3}>
+            {partNumbers.fields.map((partNumber, index) => (
+              <Stack direction="horizontal" gap={3} key={partNumber.id}>
+                <Stack>
+                  <Form.Control
+                    {...register(`partNumbers.${index}.value`)}
+                    isInvalid={!!(errors.partNumbers || [])[index]}
+                  />
+                  <Form.Control.Feedback type="invalid">
+                    {(errors.partNumbers || [])[index]?.value?.message && (
```
Pavinati, line 204: you can use optional chaining with indexes too `errors?.partNumbers?.[index]?.value?.message`

frontend/src/forms/UpdateHardwareType.tsx
```diff
@@ -0,0 +1,248 @@
...
+  });
+
+  const onFormSubmit = useCallback(
+    (data: FormData) => {
+      onSubmit(transformOutputData(data));
+    },
+    [onSubmit]
+  );
+
+  const partNumbers = useFieldArray({
```
Pavinati, line 129: i would define this closed to the `useForm` definition, to prevent confusion with many surounding callbacks

Pavinati (changes requested):

frontend/src/mocks/relay.ts
```diff
@@ -35,6 +35,13 @@ const relayMockResolvers: MockPayloadGenerator.MockResolvers = {
       memoryTotalBytes: 409600,
     };
   },
+  HardwareType() {
+    return {
+      handle: "esp32",
+      name: "ESP32",
+      partNumbers: ["PN_0000001"],
```
Pavinati, line 42: this does not match the part number regexp validation

Pavinati (changes requested):

frontend/src/mocks/relay.ts
davidebriani: There is no regexp validation on partNumbers, they just need to be non-empty strings.

frontend/src/forms/index.ts
davidebriani: Let's stick with the same regex used in the backend for now.

frontend/src/mocks/relay.ts
Pavinati: my bad, i confused it with the `handle` property

frontend/src/forms/index.ts
Pavinati: Ok, but now the feedback field doesn't match what the regexp does. Starting the handle with a `-` would give an error but the user won't understand why.
We can have 2 regex, one for characters and one to say the handle cannot start with `-`, otherwise we have to update out feedback message to reflect the updated check.

davidebriani: Thanks, you're right. I updated the error message to `The part number must start with a letter and only contain lower case characters, numbers or the hyphen symbol -`, so now it matches the RegExp

frontend/src/components/HardwareTypesTable.tsx
```diff
@@ -0,0 +1,91 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2021 SECO Mind
```
Pavinati, line 4: Missing Srl

frontend/src/i18n/langs/en.json
```diff
@@ -1 +1,187 @@
...
+    "defaultMessage": "Login"
+  },
+  "pages.Logout.title": {
+    "defaultMessage": "Logout"
+  },
+  "validation.array.min": {
+    "defaultMessage": "Does not have enough values."
+  },
+  "validation.partNumber.format": {
+    "defaultMessage": "Only lower case characters, numbers and the hypen symbol are allowed."
```
Pavinati, line 179: this must be updated

Pavinati (changes requested):

Pavinati (approved):

bettio (approved):

PR #41, opened 2021-11-10
Validate format for 'handle' of Hardware Types

davidebriani:
Pattern match 'handle' property of Hardware Types to only allow lower case ASCII letters (from a to z), digits and -


rbino (approved):

backend/lib/edgehog/appliances/hardware_type.ex
```diff
@@ -38,5 +38,8 @@ defmodule Edgehog.Appliances.HardwareType do
     |> validate_required([:name, :handle])
     |> unique_constraint([:name, :tenant_id])
     |> unique_constraint([:handle, :tenant_id])
+    |> validate_format(:handle, ~r/^[a-z\d\-]+$/,
```
bettio, line 41: `[a-z][a-z\d\-]*`

bettio (changes requested):

bettio (approved):

PR #54, opened 2021-11-16
Display device geolocation

davidebriani:
Show a Geolocation tab in the Device page displaying the last known position of the device on a map, if available.

https://user-images.githubusercontent.com/60540759/142637928-3f7407a5-1a6d-4d8a-b2e0-f2e7d33789c0.mp4


frontend/public/index.html
```diff
@@ -25,6 +25,12 @@
     <meta name="description" content="Edgehog" />
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
+    <link
```
Pavinati, line 28: why importing it from a CDN compared to importing it from the npm package inside the Map component?
https://github.com/Leaflet/Leaflet/blob/master/dist/leaflet.css

davidebriani: Importing it via Webpack requires either a custom loaders configuration for Webpack or custom code to also import each image/asset file needed besides the CSS.
I think that linking it in the `index.html` is a simple and more maintainable approach for now.

frontend/src/components/Map.tsx
```diff
@@ -0,0 +1,48 @@
...
+type Props = {
+  latitude: number;
+  longitude: number;
+  popup?: React.ReactNode;
+};
+
+const Map = ({ latitude, longitude, popup, ...props }: Props) => {
+  return (
+    <MapContainer
+      style={{ minHeight: 300 }}
```
Pavinati, line 30: is it possible to use css classes instead of inline styling?

frontend/src/components/Icon.tsx
```diff
@@ -40,6 +41,7 @@ const icons = {
   circle: faCircle,
   delete: faTrash,
   devices: faTabletAlt,
+  locationMarker: faMapMarkerAlt,
```
Pavinati, line 44: doesn't look like we're using this one yet

frontend/src/pages/Device.tsx
```diff
@@ -193,6 +209,61 @@ const DeviceHardwareInfo = ({ deviceRef }: DeviceHardwareInfoProps) => {
...
+          />
+        </p>
+        <Map
+          latitude={location.latitude}
+          longitude={location.longitude}
+          popup={
+            <div>
+              <p>{location.address}</p>
+              <p>
+                {location.latitude}°N, {location.longitude}°E
```
szakhlypa, line 257: Latitudes and longitudes also can be negative.
Negative latitudes represent the southern hemisphere (so with latitude < 0 N -> S)
And negative longitudes represent the western hemisphere (so E -> W)

davidebriani: Nice catch. The convention tells to use either:
- a decimal representation with pure numbers
- a N/E/S/W representation with non-negative degrees
Let's stick with the same decimal format returned from the API for now.

frontend/src/components/Icon.tsx
```diff
@@ -25,6 +25,7 @@ import {
   faCircle,
   faCompactDisc,
   faPlus,
+  faMapMarkerAlt,
```
Pavinati, line 28: you're still importing the image

Pavinati (approved):

bettio (approved):

PR #57, opened 2021-11-18
Add Documentation

bettio:
Add an Elixir exdoc based project for generating documentation for the main doc site.


.github/workflows/docs-workflow.yaml
```diff
@@ -0,0 +1,64 @@
...
+        path: edgehog
+    # Checkout the docs repository
+    - uses: actions/checkout@v2
+      with:
+        repository: edgehog-device-manager/docs
+        path: docs
+    - uses: erlef/setup-beam@v1.7.0
+      with:
+        otp-version: "23.2"
+        elixir-version: "1.11.4"
```
davidebriani, line 33: Should we match the versions with the ones defined in the repo's `.tool-versions`?

```
erlang 24.1.4
elixir 1.12.3-otp-24
```

doc/mix.exs
```diff
@@ -0,0 +1,35 @@
+defmodule Doc.MixProject do
+  use Mix.Project
+
+  def project do
+    [
+      app: :doc,
+      version: "0.1.0",
+      elixir: "~> 1.11",
```
Pavinati, line 8: We're using elixir 1.12.3

rbino (approved):

PR #60, opened 2021-11-19
Add descriptions to fields of the GraphQL schema

davidebriani:
Add `@desc` annotations to document the meaning and purpose of fields in the GraphQL schema.

### DIFF
--- a/backend/lib/edgehog_web/schema/appliances_types.ex
+++ b/backend/lib/edgehog_web/schema/appliances_types.ex
@@ -22,71 +22,132 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
 
   alias EdgehogWeb.Resolvers
 
+  @desc """
+  Denotes a type of hardware that devices can have.
+
+  It refers to the physical components embedded in an appliance.
+  This can represent, e.g., multiple revisions of a PCB (each with a \
+  different part number) which are functionally equivalent from the appliance \
+  point of view.
+  """
   node object(:hardware_type) do
+    @desc "The display name of the hardware type."
     field :name, non_null(:string)
+
+    @desc "The identifier of the hardware type."
     field :handle, non_null(:string)
 
+    @desc "The list of part numbers associated with the hardware type."
     field :part_numbers, non_null(list_of(non_null(:string))) do
       resolve &Resolvers.Appliances.extract_hardware_type_part_numbers/3
     end
   end
 
+  @desc """
+  Represents a specific appliance model.
+
+  An appliance model corresponds to what the users thinks as functionally \
+  equivalent appliances (e.g. two revisions of an appliance containing two \
+  different embedded chips but that have the same enclosure and the same \
+  functionality).
+  Each ApplianceModel must be associated to a specific HardwareType.
+  """
   node object(:appliance_model) do
+    @desc "The display name of the appliance model."
     field :name, non_null(:string)
+
+    @desc "The identifier of the appliance model."
     field :handle, non_null(:string)
+
+    @desc "The type of hardware that can be plugged into the appliance model."
     field :hardware_type, non_null(:hardware_type)
 
+    @desc "The list of part numbers associated with the appliance model."
     field :part_numbers, non_null(list_of(non_null(:string))) do
       resolve &Resolvers.Appliances.extract_appliance_model_part_numbers/3
     end
   end
 
   object :appliances_queries do
+    @desc "Fetches the list of all hardware types."
     field :hardware_types, non_null(list_of(non_null(:hardware_type))) do
       resolve &Resolvers.Appliances.list_hardware_types/3
     end
 
+    @desc "Fetches a single hardware type."
     field :hardware_type, :hardware_type do
+      @desc "The ID of the hardware type."
       arg :id, non_null(:id)
+
       middleware Absinthe.Relay.Node.ParseIDs, id: :hardware_type
       resolve &Resolvers.Appliances.find_hardware_type/2
     end
 
+    @desc "Fetches the list of all appliance models."
     field :appliance_models, non_null(list_of(non_null(:appliance_model))) do
       resolve &Resolvers.Appliances.list_appliance_models/3
     end
 
+    @desc "Fetches a single appliance model."
     field :appliance_model, :appliance_model do
+      @desc "The ID of the appliance model."
       arg :id, non_null(:id)
+
       middleware Absinthe.Relay.Node.ParseIDs, id: :appliance_model
       resolve &Resolvers.Appliances.find_appliance_model/2
     end
   end
 
   object :appliances_mutations do
+    @desc "Creates a new hardware type."
     payload field :create_hardware_type do
       input do
+        @desc "The display name of the hardware type."
         field :name, non_null(:string)
+
+        @desc """
+        The identifier of the hardware type.
+
+        It should start with a lower case ASCII letter and only contain \
+        lower case ASCII letters, digits and the hyphen - symbol.
+        """
         field :handle, non_null(:string)
+
+        @desc "The list of part numbers associated with the hardware type."
         field :part_numbers, non_null(list_of(non_null(:string)))
       end
 
       output do
+        @desc "The created hardware type."
         field :hardware_type, non_null(:hardware_type)
       end
 
       resolve &Resolvers.Appliances.create_hardware_type/3
     end
 
+    @desc "Updates a hardware type."
     payload field :update_hardware_type do
       input do
+        @desc "The ID of the hardware type to be updated."
         field :hardware_type_id, non_null(:id)
+
+        @desc "The display name of the hardware type."
         field :name, :string
+
+        @desc """
+        The identifier of the hardware type.
+
+        It should start with a lower case ASCII letter and only contain \
+        lower case ASCII letters, digits and the hyphen - symbol.
+        """
         field :handle, :string
+
+        @desc "The list of part numbers associated with the hardware type."
         field :part_numbers, list_of(non_null(:string))
       end
 
       output do
+        @desc "The updated hardware type."
         field :hardware_type, non_null(:hardware_type)
       end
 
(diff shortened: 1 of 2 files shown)


backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -23,75 +23,156 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
...
+  When multiple filters are specified, they are applied in an AND fashion to further refine the results.
+  """
   input_object :device_filter do
+    @desc "Whether to return devices connected or not to Astarte."
     field :online, :boolean
+
+    @desc "A string to match against the device ID. The match is case-insensitive and tests whether the string is included in the device ID."
     field :device_id, :string
+
+    @desc "A string to match against the part number of the device's appliance model. The match is case-insensitive and tests whether the string is included in the part number of the device's appliance model."
```
Pavinati, line 38: Those lines are pretty long. Is there a way to split the desc code in multiple lines without changing the resulting doc string?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,154 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
       resolve &Resolvers.Appliances.update_hardware_type/3
     end
 
+    @desc "Creates a new appliance model."
     payload field :create_appliance_model do
       input do
+        @desc "The display name of the appliance model."
         field :name, non_null(:string)
+
+        """
```
rbino, line 156: Missing `@desc`?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -111,15 +178,29 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
+    @desc "Updates an appliance model."
     payload field :update_appliance_model do
       input do
+        @desc "The ID of the appliance model to be updated."
         field :appliance_model_id, non_null(:id)
+
+        @desc "The display name of the appliance model."
         field :name, :string
+
+        """
```
rbino, line 190: Missing `@desc`?

backend/lib/edgehog_web/schema/astarte_types.ex
rbino: It's possible to do something like
```
@desc "A long string " <>
  "on multiple " <>
  "lines"
```
but it becomes ugly quickly: it's a mess to maintain if you have to add something to a line, you have to remember the spaces at the end of the line etc.

Imho we either accept the long lines or use heredocs (accepting the newlines in the resulting doc) where needed

rbino (changes requested):

backend/lib/edgehog_web/schema/astarte_types.ex
Pavinati: That's a good point.
In this case, I personally vouch for heredocs.
I think documentation in code should be readable without having to scroll horizontally or configuring the editor to wrap long lines.

davidebriani: Actually using heredocs produces the correct output, adopting the `\` character to break lines.
So for shorter lines we can use `@desc "some desc"` and for the longer ones use:

```elixir
    [...]
    field :device_id, :string

    @desc """
    A string to match against the part number of the device's appliance model.
    
    The match is case-insensitive and tests whether the string is included in \
    the part number of the device's appliance model.
    """
    field :appliance_model_part_number, :string
    
    [...]
```

This way the leading whitespace is stripped from each, based on the trailing space of the last line
```
    """
```
and lines broken by `\` will produce a single line without line-breaks.

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,159 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
     field :handle, non_null(:string)
 
+    @desc "The list of part numbers associated with the hardware type."
     field :part_numbers, non_null(list_of(non_null(:string))) do
       resolve &Resolvers.Appliances.extract_hardware_type_part_numbers/3
     end
   end
 
+  @desc """
+  Denotes the type of appliance that a device's hardware can be plugged to.
```
bettio, line 44: `plugged into` maybe?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,159 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
+
+  It refers to the functional components that the device can be integrated \
+  with.
+  Each ApplianceModel must be associated to a specific HardwareType.
+  """
   node object(:appliance_model) do
+    @desc "The display name of the appliance model."
     field :name, non_null(:string)
+
+    @desc "The reference name of the appliance model."
```
bettio, line 54: `The identifier` maybe?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,159 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
+  @desc """
+  Denotes a type of hardware that devices can have.
+
+  It refers to the physical components embedded in a device's board.
+  """
   node object(:hardware_type) do
+    @desc "The display name of the hardware type."
     field :name, non_null(:string)
+
+    @desc "The reference name of the hardware type."
```
bettio, line 34: `The identifier` maybe?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,159 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
 
   object :appliances_mutations do
+    @desc "Creates a new hardware type."
     payload field :create_hardware_type do
       input do
+        @desc "The display name of the hardware type."
         field :name, non_null(:string)
+
+        @desc """
+        The reference name of the hardware type.
```
bettio, line 104: `The identifier` maybe?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,159 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
     payload field :update_hardware_type do
       input do
+        @desc "The ID of the hardware type to be updated."
         field :hardware_type_id, non_null(:id)
+
+        @desc "The display name of the hardware type."
         field :name, :string
+
+        @desc """
+        The reference name of the hardware type.
```
bettio, line 133: `The identifier` maybe?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,159 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
     end
 
+    @desc "Creates a new appliance model."
     payload field :create_appliance_model do
       input do
+        @desc "The display name of the appliance model."
         field :name, non_null(:string)
+
+        @desc """
+        The reference name of the appliance model.
```
bettio, line 160: `The identifier` maybe?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -111,15 +183,29 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
     payload field :update_appliance_model do
       input do
+        @desc "The ID of the appliance model to be updated."
         field :appliance_model_id, non_null(:id)
+
+        @desc "The display name of the appliance model."
         field :name, :string
+
+        @desc """
+        The reference name of the appliance model.
```
bettio, line 196: `The identifier` maybe?

bettio (changes requested):

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,160 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
 
   alias EdgehogWeb.Resolvers
 
+  @desc """
+  Denotes a type of hardware that devices can have.
+
+  It refers to the physical components embedded in a device's board.
```
bettio, line 28: I would change to "It refers to the physical components embedded in an appliance."

bettio: @rbino any comment here?

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,160 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
     field :handle, non_null(:string)
 
+    @desc "The list of part numbers associated with the hardware type."
     field :part_numbers, non_null(list_of(non_null(:string))) do
       resolve &Resolvers.Appliances.extract_hardware_type_part_numbers/3
     end
   end
 
+  @desc """
+  Denotes the type of appliance that a device's hardware can be plugged into.
```
bettio, line 44: I was thinking about this twice, and it sound a bit weird. @rbino any comment here?

bettio (changes requested):

backend/lib/edgehog_web/schema/appliances_types.ex
rbino: I'd go with your proposal, maybe adding "This can represent, e.g., multiple revisions of a PCB (each with a different part number) which are functionally equivalent from the appliance point of view."

backend/lib/edgehog_web/schema/appliances_types.ex
```diff
@@ -22,87 +22,160 @@ defmodule EdgehogWeb.Schema.AppliancesTypes do
...
+
+  It refers to the functional components that the device can be integrated \
+  with.
```
rbino, line 47: @bettio proposal for replacement of the highlighted lines:

Represents a specific appliance model. An appliance model corresponds to what the users thinks as functionally equivalent appliances (e.g. two revisions of an appliance containing two different embedded chips but that have the same enclosure and the same functionality).

backend/lib/edgehog_web/schema/appliances_types.ex
rbino: See other comment

rbino (changes requested):

bettio (approved):

PR #61, opened 2021-11-19
Display appliance model and hardware type for devices

davidebriani:
Fetch and display the appliance model and the associated hardware type for each device.

Screenshot 2021-11-19 at 16-54-20 Edgehog

### DIFF
--- a/frontend/src/components/DevicesTable.tsx
+++ b/frontend/src/components/DevicesTable.tsx
@@ -32,6 +32,12 @@ type DeviceProps = {
   lastDisconnection: string | null;
   name: string;
   online: boolean;
+  applianceModel: {
+    name: string;
+    hardwareType: {
+      name: string;
+    };
+  } | null;
 };
 
 const columns: Column<DeviceProps>[] = [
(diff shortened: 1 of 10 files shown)


frontend/src/components/DevicesTable.tsx
```diff
@@ -77,6 +83,28 @@ const columns: Column<DeviceProps>[] = [
     ),
     sortType: "basic",
   },
+  {
+    id: "applianceModel",
+    accessor: (device) => device.applianceModel?.name,
+    Header: (
+      <FormattedMessage
```
Pavinati, line 90: we already have `Device.applianceModel` for that very same string. Shouldn't we reuse it?

frontend/src/components/DevicesTable.tsx
```diff
@@ -77,6 +83,28 @@ const columns: Column<DeviceProps>[] = [
...
+        description="Title for the Appliance Model column of the devices table"
+      />
+    ),
+  },
+  {
+    id: "hardwareType",
+    accessor: (device) => device.applianceModel?.hardwareType.name,
+    Header: (
+      <FormattedMessage
+        id="components.DevicesTable.hardwareTypeTitle"
```
Pavinati, line 102: same here with `Device.hardwareType`

frontend/src/components/DevicesTable.tsx
davidebriani: Yep, that's fine. Fixed.

rbino (approved):

Pavinati (approved):

bettio (approved):

PR #64, opened 2021-11-22
Display Storage Usage

### LINKED ISSUE #35: Display Device Local Storage Status
Device page should display a tab with information about local storage status (e.g. avail / used disk space) for each disk.

davidebriani:
Fetch Storage data from AppEngine and display the info in a tab on the Device page.

Storage tab displaying some storage units:

Screenshot 2021-11-24 at 12-25-29 Edgehog

Storage tab with no storage unit found:

Screenshot 2021-11-24 at 12-28-41 Edgehog

Closes #35

### DIFF
--- a/backend/lib/edgehog/astarte/device/hardware_info.ex
+++ b/backend/lib/edgehog/astarte/device/hardware_info.ex
@@ -13,6 +13,9 @@ defmodule Edgehog.Astarte.Device.HardwareInfo do
   @interface "io.edgehog.devicemanager.HardwareInfo"
 
   def get(%AppEngine{} = client, device_id) do
+    # TODO: right now we request the whole interface at once, so `memory_total_bytes` can't
+    # be requested as string (see https://github.com/astarte-platform/astarte/issues/630).
+    # Request it as string as soon as that issue is solved.
     with {:ok, %{"data" => data}} <-
            AppEngine.Devices.get_properties_data(client, device_id, @interface) do
       hardware_info = %HardwareInfo{
--- a/backend/config/test.exs
+++ b/backend/config/test.exs
@@ -33,6 +33,7 @@ config :tesla, adapter: Tesla.Mock
 
 # Astarte mocks for tests
 config :edgehog, :astarte_device_status_module, Edgehog.Astarte.Device.DeviceStatusMock
+config :edgehog, :astarte_storage_usage_module, Edgehog.Astarte.Device.StorageUsageMock
 config :edgehog, :astarte_wifi_scan_result_module, Edgehog.Astarte.Device.WiFiScanResultMock
 
 config :edgehog,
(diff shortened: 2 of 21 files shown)


rbino (changes requested):

bettio (approved):

backend/lib/edgehog/astarte/device/storage.ex
```diff
@@ -0,0 +1,68 @@
...
+# limitations under the License.
+#
+
+defmodule Edgehog.Astarte.Device.Storage do
+  @enforce_keys [:label]
+  defstruct [
+    :label,
+    :total_bytes,
+    :free_bytes
+  ]
```
rbino, line 25: Following the changes in the GraphQL API, we could put the struct in a separate module here too (i.e. `Storage.get` returns a list of `Storage.StorageUnit`)

backend/lib/edgehog/astarte/device/hardware_info.ex
```diff
@@ -13,6 +13,9 @@ defmodule Edgehog.Astarte.Device.HardwareInfo do
   @interface "io.edgehog.devicemanager.HardwareInfo"
 
   def get(%AppEngine{} = client, device_id) do
+    # TODO: right now we request the whole interface at once, so `memory_total_bytes` can't
```
rbino, line 16: You should copy this TODO also in the `Storage.get` function

backend/lib/edgehog/astarte/device/storage.ex
```diff
@@ -0,0 +1,68 @@
...
+  alias Astarte.Client.AppEngine
+  alias Edgehog.Astarte.Device.Storage
+
+  @interface "io.edgehog.devicemanager.Storage"
+
+  def get(%AppEngine{} = client, device_id) do
+    with {:ok, %{"data" => data}} <-
+           AppEngine.Devices.get_datastream_data(client, device_id, @interface) do
+      storage_list =
+        data["storage"]
```
rbino, line 44: This will crash if the interface is empty (so there's no `storage` key in `data`), since you will call `Enum.map(nil, fun)`.
You have to decide what you want to return in case of an empty interface.
- If you want to return an empty list, then you can do
```elixir
Map.get(data, "storage", %{})
|> Enum.map(...)
```
- If you want to return `{:ok, nil}`, you will probably get away with an `if`.
Imho the first approach makes sense (and it's the behaviour you're declaring in the typespec), but if that's the case then the GraphQL type wille become `non_null(list_of(non_null(:storage_unit)))` because you're effectively always returning a (possibly empty) list

davidebriani: I went with `Enum.map(data, ...)` so it should avoid the `nil` case since `data` should not be empty.
This way we get:
- an empty list when the device supports the interface but has no storage units
- `nil` if the device does not support the interface, which will result in the frontend not showing the Storage tab

bettio (approved):

rbino (approved):

PR #70, opened 2021-11-29
Display system status for devices

davidebriani:
Fetch SystemStatus data from AppEngine and display the info in a tab on the Device page

Screenshot 2021-11-29 at 16-39-13 Edgehog

### DIFF
--- a/backend/lib/edgehog_web/resolvers/astarte.ex
+++ b/backend/lib/edgehog_web/resolvers/astarte.ex
@@ -44,6 +44,13 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     end
   end
 
+  def fetch_system_status(%Device{} = device, _args, _context) do
+    case Astarte.fetch_system_status(device) do
+      {:ok, system_status} -> {:ok, system_status}
+      _ -> {:ok, nil}
+    end
+  end
+
   def fetch_wifi_scan_results(%Device{} = device, _args, _context) do
     case Astarte.fetch_wifi_scan_results(device) do
       {:ok, wifi_scan_results} -> {:ok, wifi_scan_results}
--- a/backend/lib/edgehog_web/schema/astarte_types.ex
+++ b/backend/lib/edgehog_web/schema/astarte_types.ex
@@ -141,6 +141,26 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
   end
 
   @desc """
+  Describes the current status of the operating system of a device.
+  """
+  object :system_status do
+    @desc "The identifier of the performed boot sequence."
+    field :boot_id, :string
+
+    @desc "The number of free bytes of memory."
+    field :memory_free_bytes, :integer
+
+    @desc "The number of running tasks on the system."
+    field :task_count, :integer
+
+    @desc "The number of milliseconds since the last system boot."
+    field :uptime_milliseconds, :integer
+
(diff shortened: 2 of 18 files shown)


backend/lib/edgehog_web/resolvers/astarte.ex
```diff
@@ -44,6 +44,13 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     end
   end
 
+  def fetch_system_status(%Device{} = device, _args, _context) do
+    case Astarte.fetch_system_status(device) do
+      {:ok, [system_status | _tail]} -> {:ok, system_status}
```
rbino, line 49: I'd move the list unpacking in `fetch_system_status/1`. The fact that you obtain a list of SystemStatus is an implementation detail imho, and we're only interested in the last one right now.

backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -140,6 +140,26 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
...
+    field :memory_free_bytes, :integer
+
+    @desc "The number of running tasks on the system."
+    field :task_count, :integer
+
+    @desc "The number of milliseconds since the last system boot."
+    field :uptime_milliseconds, :integer
+
+    @desc "The date at which the system status was read."
+    field :timestamp, non_null(:datetime)
```
rbino, line 160: Why only this is `non_null`? Since it's an object aggregation I expect that either we have all fields or none (so the `non_null` would be one layer above)

rbino (changes requested):

backend/lib/edgehog_web/schema/astarte_types.ex
davidebriani: Last time I checked I could send object aggregated with null values. So I'd take that into account. The `timestamp` however will be present for sure.

rbino: Makes sense

frontend/src/pages/Device.tsx
```diff
@@ -161,85 +175,105 @@ const DeviceHardwareInfo = ({ deviceRef }: DeviceHardwareInfoProps) => {
...
+        <FormattedMessage
+          id="pages.Device.hardwareInfoTab"
+          defaultMessage="Hardware Info"
+        />
+      }
+    >
+      <div className="mt-3">
+        <Stack gap={3}>
+          {hardwareInfo.cpuArchitecture != null && (
+            <FormRow
```
Pavinati, line 190: we have loads of form rows that are all the same.
I think we can create an array of objects and map over them
```
[
  {
    value: hardwareInfo.cpuArchitecture,
    type: "text",
    label: <FormattedMessage id="Device.hardwareInfo.cpuArchitecture" defaultMessage="CPU architecture" />
  },
  {
    value: hardwareInfo.cpuModel,
    type: "text",
    label: <FormattedMessage id="Device.hardwareInfo.cpuModel" defaultMessage="CPU model code" />
  },
].map(ReadOnlyValue)
```
```
const ReadOnlyValue = ({ value, type, label }) => {
  if (!value) {
    return null;
  }

  if (type === "numeric") {
    [...]
  }
  // text
  [...]
}
```

rbino (approved):

PR #71, opened 2021-11-29
[backend] Device battery status

### LINKED ISSUE #37: Implement API for Fetching Device Battery and Power Supply Status
An API for fetching data about all batteries connected to the device should be implemented.

szakhlypa:
Fetch BatteryStatus data from AppEngine and expose it via GraphQL

Closes #37

### DIFF
--- a/backend/lib/edgehog_web/schema/astarte_types.ex
+++ b/backend/lib/edgehog_web/schema/astarte_types.ex
@@ -181,6 +181,48 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
   end
 
   @desc """
+  The current status of the battery.
+  """
+  enum :battery_status do
+    @desc "The battery is charging."
+    value :charging
+    @desc "The battery is discharging."
+    value :discharging
+    @desc "The battery is idle."
+    value :idle
+
+    @desc """
+    The battery is either in a charging or in an idle state, \
+    since the hardware doesn't allow to distinguish between them.
+    """
+    value :either_idle_or_charging
+    @desc "The battery is in a failed state."
+    value :failure
+    @desc "The battery is removed."
+    value :removed
+    @desc "The battery status cannot be determined."
+    value :unknown
+  end
+
+  @desc "Describes a battery slot of a device."
+  object :battery_slot do
+    @desc "The identifier of the battery slot."
+    field :slot, non_null(:string)
+
+    @desc "Battery level estimated percentage [0.0%-100.0%]"
+    field :level_percentage, :float
+
+    @desc "Battery level measurement absolute error [0.0-100.0]"
+    field :level_absolute_error, :float
+
+    @desc "The current status of the battery."
+    field :status, :battery_status do
+      resolve &Resolvers.Astarte.battery_status_to_enum/3
+      middleware Middleware.ErrorHandler
+    end
+  end
+
+  @desc """
   Denotes a device instance that connects and exchanges data.
 
   Each Device is associated to a specific ApplianceModel, which in turn is \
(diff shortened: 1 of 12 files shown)


backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -149,6 +149,23 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
...
+    field :slot, non_null(:string)
+
+    @desc "Battery level estimated percentage [0.0%-100.0%]"
+    field :level_percentage, :float
+
+    @desc "Battery level measurement absolute error [0.0-100.0]"
+    field :level_absolute_error, :float
+
+    @desc "Battery status string, any of: Charging, Discharging, Idle, EitherIdleOrCharging, Failure, Removed, Unknown"
+    field :status, non_null(:string)
```
davidebriani, line 166: It's better to wrap lines at 80 characters to improve readability.
You can write the @desc using heredocs notations, such as:
```elixir
    @desc """
    The current status of the battery.
    
    It can be any of: Charging, Discharging, Idle, EitherIdleOrCharging, \
    Failure, Removed, Unknown.
    """
```

Using heredocs produces the correct output, adopting the \ character to break lines.
So for shorter lines we can use @desc "some desc" and for the longer ones use the multiline @desc.
This way the leading whitespace is stripped from each line, based on the trailing space of the last line
```elixir
    """
```
and lines broken by `\` will produce a single line without line-breaks.

backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -193,6 +210,12 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
       resolve &Resolvers.Astarte.fetch_wifi_scan_results/3
       middleware Middleware.ErrorHandler
     end
+
+    @desc "The list of battery slots with status of the device."
```
davidebriani, line 214: Maybe just `@desc "The status of the battery slots of the device."`

davidebriani (changes requested):

davidebriani (approved):

backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -149,6 +149,28 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
...
+    @desc "Battery level measurement absolute error [0.0-100.0]"
+    field :level_absolute_error, :float
+
+    @desc """
+    The current status of the battery.
+
+    It can be any of: Charging, Discharging, Idle, EitherIdleOrCharging, \
+    Failure, Removed, Unknown.
+    """
+    field :status, non_null(:string)
```
rbino, line 171: Let's go the extra mile here and expose the battery status as a GraphQL enum.
Here's a sketch of what needs to be done

In `AstarteTypes`:
```elixir
  enum :battery_status do
    value :charging
    value :discharging
    ...
  end
```
(note: use `snake_case` for GraphQL enum values)
and
```
   field :status, non_null(:battery_status) do
     resolve &Resolvers.Astarte.battery_status_to_enum/2
   end
```

In `Resolvers.Astarte`:
```elixir
  def battery_status_to_enum(status, _context) do
    case status do
      "Charging" -> {:ok, :charging}
      "Discharging" -> {:ok, :discharging}
      ...
      _other -> {:error, :invalid_battery_status}
    end
  end
```

rbino (changes requested):
@szakhlypa the PR is good, I just suggested an improvement that fits nicely here. Also please rebase it on main since it's conflicting now that I merged device status PR.

rbino (changes requested):

backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -180,6 +180,39 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
...
+    value :discharging, description: "The battery is discharging."
+    value :idle, description: "The battery is idle."
+
+    value :either_idle_or_charging,
+      description:
+        "The battery is either in a charging or in an idle state, since the hardware doesn't allow to distinguish between them."
+
+    value :failure, description: "The battery is in a failed state."
+    value :removed, description: "The battery is removed."
+    value :unknown, description: "The battery status cannot be determined."
```
rbino, line 195: Let's be consistent with everything else:
```
@desc "The battery is charging"
value :charging
...
```

rbino (changes requested):

rbino (approved):

PR #72, opened 2021-11-30
Handle optional configuration for geolocation providers

davidebriani:
Use an optional and general config to define geolocation/geocoding providers.
- The list of providers to use is directly defined via Config
- Providers are enabled only if properly configured via environment variables, otherwise they are not used.
- a Device instance is given directly to Geolocation providers and they are free to implement the logic to fetch device's info and compute its location

### DIFF
--- a/backend/config/runtime.exs
+++ b/backend/config/runtime.exs
@@ -45,19 +45,42 @@ if config_env() == :prod do
     ],
     secret_key_base: secret_key_base
 
+  freegeoip_api_key = System.get_env("FREEGEOIP_API_KEY")
+
+  freegeoip_provider = %{
+    module: Edgehog.Geolocation.Providers.FreeGeoIp,
+    enabled?: !is_nil(freegeoip_api_key),
+    config: [api_key: freegeoip_api_key]
+  }
+
+  google_geolocation_api_key = System.get_env("GOOGLE_GEOLOCATION_API_KEY")
+
+  google_geolocation_provider = %{
+    module: Edgehog.Geolocation.Providers.GoogleGeolocation,
+    enabled?: !is_nil(google_geolocation_api_key),
+    config: [api_key: google_geolocation_api_key]
+  }
+
+  google_geocoding_api_key = System.get_env("GOOGLE_GEOCODING_API_KEY")
+
+  google_geocoding_provider = %{
+    module: Edgehog.Geolocation.Providers.GoogleGeocoding,
+    enabled?: !is_nil(google_geocoding_api_key),
+    config: [api_key: google_geocoding_api_key]
+  }
+
+  geolocation_providers = [google_geolocation_provider, freegeoip_provider]
+  geocoding_providers = [google_geocoding_provider]
+
   config :edgehog,
-    ip_geolocation_provider: Edgehog.Geolocation.Providers.FreeGeoIp,
-    wifi_geolocation_provider: Edgehog.Geolocation.Providers.GoogleGeolocation,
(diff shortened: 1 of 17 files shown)


backend/config/runtime.exs
```diff
@@ -45,19 +45,48 @@ if config_env() == :prod do
...
+  google_geocoding_api_key = System.get_env("GOOGLE_GEOCODING_API_KEY")
 
-  config :edgehog, Edgehog.Geolocation.Providers.FreeGeoIp,
-    api_key: System.fetch_env!("FREEGEOIP_API_KEY")
+  google_geocoding_provider = %{
+    module: Edgehog.Geolocation.Providers.GoogleGeocoding,
+    enabled?: !is_nil(google_geocoding_api_key),
+    config: %{
+      api_key: google_geocoding_api_key
+    }
```
rbino, line 75: Why do you provide the config as a map just to convert it to a Keyword list in `Enum.each`? Just use a Keyword list here directly

backend/lib/edgehog/geolocation/geolocation_provider.ex
```diff
@@ -16,14 +16,15 @@
...
+defmodule Edgehog.Geolocation.GeolocationProvider do
+  alias Edgehog.Astarte.Device
 
   @type coordinates :: %{
           latitude: float,
           longitude: float,
-          accuracy: number | nil
+          accuracy: number | nil,
+          timestamp: DateTime.t()
         }
```
rbino, line 27: We should probably convert this to a struct since this is returned by a behaviour (and a struct gives you stronger guarantees when you build it about not using the wrong key names etc).
Something like `Edgehog.Geolocation.Position`

backend/lib/edgehog/geolocation/providers/google_geolocation.ex
```diff
@@ -17,16 +17,48 @@
...
+  @impl Edgehog.Geolocation.GeolocationProvider
+  def geolocate(%Device{} = device) do
+    with {:ok, wifi_scan_results} <- Astarte.fetch_wifi_scan_results(device),
+         {:ok, wifi_scan_results} <- filter_latest_wifi_scan_results(wifi_scan_results),
+         {:ok, coordinates} <- geolocate_wifi(wifi_scan_results) do
+      timestamp =
+        case Enum.empty?(wifi_scan_results) do
+          false -> List.first(wifi_scan_results).timestamp
+          true -> DateTime.utc_now()
+        end
```
rbino, line 40: When dealing with booleans just use an `if`
```elixir
if Enum.empty?(...) do
...
else
...
end
```

rbino (changes requested):

backend/lib/edgehog/geolocation/providers/google_geolocation.ex
```diff
@@ -42,7 +42,10 @@ defmodule Edgehog.Geolocation.Providers.GoogleGeolocation do
     latest_scan = Enum.max_by(wifi_scan_results, & &1.timestamp, DateTime)
 
     latest_wifi_scan_results =
-      Enum.filter(wifi_scan_results, &(&1.timestamp == latest_scan.timestamp))
+      Enum.filter(
+        wifi_scan_results,
+        &(DateTime.diff(latest_scan.timestamp, &1.timestamp, :second) < 1)
```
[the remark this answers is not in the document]
davidebriani: `latest_scan.timestamp` is the latest timestamp of the list, since we already computed the max a few lines before.
Therefore, the diff cannot be negative.

Pavinati (approved):

bettio (approved):

PR #84, opened 2021-12-03
Fix backend modules that fetch and parse Astarte data

davidebriani:
Fix parsing of Astarte data for the modules:
- StorageUsage: correctly parse data that are wrapped in an array. Use `limit: 1` query option to be sure to fetch the latest data
- SystemStatus: use `limit: 1` query option to be sure to fetch the latest data
- WiFiScanResult: use `limit: 1000` query option to avoid heavy results and to ensure the results are already DESC-ordered by timestamp

### DIFF
--- a/backend/lib/edgehog/astarte/device/system_status.ex
+++ b/backend/lib/edgehog/astarte/device/system_status.ex
@@ -47,22 +47,20 @@ defmodule Edgehog.Astarte.Device.SystemStatus do
     # type Object Aggregrate.
     # For details, see https://github.com/astarte-platform/astarte/issues/630
     with {:ok, %{"data" => data}} <-
-           AppEngine.Devices.get_datastream_data(client, device_id, @interface) do
-      system_status_list =
-        data["systemStatus"]
-        |> Enum.map(fn ss ->
-          %SystemStatus{
-            boot_id: ss["bootId"],
-            memory_free_bytes: parse_longinteger(ss["availMemoryBytes"]),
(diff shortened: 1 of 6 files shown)


backend/lib/edgehog/astarte/device/system_status.ex
```diff
@@ -47,22 +47,20 @@ defmodule Edgehog.Astarte.Device.SystemStatus do
...
-          %SystemStatus{
-            boot_id: ss["bootId"],
-            memory_free_bytes: parse_longinteger(ss["availMemoryBytes"]),
-            task_count: ss["taskCount"],
-            uptime_milliseconds: parse_longinteger(ss["uptimeMillis"]),
-            timestamp: parse_datetime(ss["timestamp"])
-          }
-        end)
+           AppEngine.Devices.get_datastream_data(client, device_id, @interface, limit: 1) do
+      case List.first(data["systemStatus"]) do
```
rbino, line 51: You're assuming there's a `systemStatus` key and checking if there's at least one entry for it, but if there's no data there actually won't even be a `systemStatus` key. So this should be:
```
case Map.fetch(data, "systemStatus") do
  :error ->
    {:error, :system_status_not_found
  {:ok, [status]} ->
    # Do stuff
end
```
Note that I explicitly match on a single value since if the `systemStatus` key is there, the list will have a single value by definition.

rbino (changes requested):

rbino (approved):

PR #86, opened 2021-12-03
Fix BatteryStatus module to correctly parse Astarte data

szakhlypa:
- Add fix to properly parse data fetched from Astarte
- Add a test to ensure the correct behavior

Signed-off-by: Sergey Zakhlypa <sergey.zakhlypa@secomind.com>

### DIFF
--- a/backend/test/edgehog/astarte/device/battery_status_test.exs
+++ b/backend/test/edgehog/astarte/device/battery_status_test.exs
@@ -1,0 +1,87 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2021 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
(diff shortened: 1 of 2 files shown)


backend/test/edgehog/astarte/device/battery_status_test.exs
```diff
@@ -0,0 +1,85 @@
...
+            }
+          ],
+          "slot2" => [
+            %{
+              "levelPercentage" => 70.0,
+              "levelAbsoluteError" => 100.0,
+              "status" => "EitherIdleOrCharging"
+            }
+          ]
+        }
```
davidebriani, line 59: I would add the timestamps to each battery slot returned by Astarte, so we have the same exact format for data as the one specified in the interface definition.
The fact that the `BatterySlot` module discards the timestamps is fine, but this way we can be sure it will correctly parse real data, without crashing because it didn't expect extra properties from Astarte.

backend/test/edgehog/astarte/device/battery_status_test.exs
```diff
@@ -0,0 +1,85 @@
...
+      }
+
+      mock(fn
+        %{method: :get, url: _api_url} ->
+          json(response)
+      end)
```
rbino, line 65: Why is Tesla.Mock needed here? This should be mocked at the `Mox` level

rbino (changes requested):

backend/test/edgehog/astarte/device/battery_status_test.exs
davidebriani: We're using `Mox` to have mocks of our own modules, but here we're testing the inner workings of the module.

We can either:
1. use `Tesla.Mock` to mock the network call's response
2. use `Mox` to mock the entire AppEngine client used by the module, but AppEngine is an external module and we don't have access to its behavior to stub it with a mock

I think option 1 is the easiest to carry on the fix.

rbino (approved):

PR #96, opened 2021-12-15
[backend] Device OS info 

### LINKED ISSUE #93: Add support to backend for OS Info

szakhlypa:
Closes #93

### DIFF
--- a/backend/lib/edgehog/astarte/device/os_info.ex
+++ b/backend/lib/edgehog/astarte/device/os_info.ex
@@ -16,35 +16,35 @@
 # limitations under the License.
 #
 
-defmodule Edgehog.Astarte.Device.HardwareInfo do
-  defstruct [
-    :cpu_architecture,
-    :cpu_model,
-    :cpu_model_name,
-    :cpu_vendor,
-    :memory_total_bytes
-  ]
+defmodule Edgehog.Astarte.Device.OSInfo do
+  @type t :: %__MODULE__{
+          name: String.t() | nil,
+          version: String.t() | nil
+        }
+
+  @enforce_keys [:name, :version]
+  defstruct @enforce_keys
+
+  @behaviour Edgehog.Astarte.Device.OSInfo.Behaviour
 
   alias Astarte.Client.AppEngine
-  alias Edgehog.Astarte.Device.HardwareInfo
 
-  @interface "io.edgehog.devicemanager.HardwareInfo"
+  @interface "io.edgehog.devicemanager.OSInfo"
 
+  @impl true
   def get(%AppEngine{} = client, device_id) do
-    # TODO: right now we request the whole interface at once, so `memory_total_bytes` can't
-    # be requested as string (see https://github.com/astarte-platform/astarte/issues/630).
-    # Request it as string as soon as that issue is solved.
     with {:ok, %{"data" => data}} <-
            AppEngine.Devices.get_properties_data(client, device_id, @interface) do
-      hardware_info = %HardwareInfo{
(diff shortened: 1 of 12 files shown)


backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -140,6 +140,15 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
     field :timestamp, non_null(:datetime)
   end
 
+  @desc "Describes an operating system of a device."
+  object :os_info do
+    @desc "The name of the operating system."
+    field :os_name, :string
```
szakhlypa, line 146: Should we step back from interface naming and expose it as name and version?
os_name -> name
os_version -> version

backend/lib/edgehog/astarte/device/os_info.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Edgehog.Astarte.Device.OSInfo do
+  @type t :: %__MODULE__{
+          os_name: String.t() | nil,
+          os_version: String.t() | nil
+        }
+
+  @enforce_keys [:os_name, :os_version]
+  defstruct @enforce_keys
```
davidebriani, line 26: You are enforcing the presence of the two keys but you have declared them as `nil`lable in the type.
We can opt to keep them nullable, removing `@enforce_keys`:
```
defstruct [:os_name, :os_version]
```

backend/lib/edgehog/astarte/device/os_info/behaviour.ex
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+defmodule Edgehog.Astarte.Device.OSInfo.Behaviour do
+  alias Astarte.Client.AppEngine
+
+  @callback get(client :: AppEngine.t(), device_id :: String.t()) ::
+              {:ok, term()} | {:error, term()}
```
davidebriani, line 23: In the return type you can use the correct one instead of `term()`.
```elixir
alias Astarte.Client.AppEngine
alias Edgehog.Astarte.Device.OSInfo

@callback get(client :: AppEngine.t(), device_id :: String.t()) ::
              {:ok, OSInfo.t()} | {:error, term()}
```

backend/lib/edgehog/astarte/device/os_info.ex
```diff
@@ -0,0 +1,34 @@
+defmodule Edgehog.Astarte.Device.OSInfo do
```
Pavinati, line 1: missing license header

szakhlypa: Thanks, added headers for OSInfo and HardwareInfo

backend/lib/edgehog_web/schema/astarte_types.ex
davidebriani: Yes I think we could drop the prefix, it is still clear that `name` and `version` refer to the OS.
You could start dropping the prefix since the `parseData` method in the OSInfo module.

davidebriani (changes requested):

backend/lib/edgehog/astarte/device/os_info.ex
davidebriani: Sorry, I confused enforcing the presence of the keys with the validation of their values. Your code here is perfectly fine

davidebriani (approved):

rbino (approved):

PR #126, opened 2022-02-07
Add tenant login page

davidebriani:
Add the authentication form in the Login page to authenticate against a specific tenant via `tenantSlug` + `authToken`.
Once authenticated, the backend endpoint is set up to hit the APIs for the selected tenant.
The auth data is persisted via cookies for the session, or for a year if the user selects `Keep me logged in` upon login.
The Login page accepts the URL query params `tenantSlug` and `authToken` to perform automatic authentication via URL.

Screenshot 2022-02-07 at 16-59-18 Edgehog


frontend/src/api/index.ts
```diff
@@ -25,23 +25,24 @@ import {
...
 const fetchGraphQL = async (
   query: string | null | undefined,
-  variables: Record<string, unknown>
+  variables: Record<string, unknown>,
+  authConfig: AuthConfig | null
 ) => {
   const userLanguage = navigator.language; // TODO allow users to overwrite this
-  const authToken = loadAuthToken();
-  const response = await fetch(backendUrl, {
+  const apiUrl = new URL(`tenants/${authConfig?.tenantSlug}/api`, backendUrl);
```
Pavinati, line 41: `authConfig` may be null. What should we do in this case?
If `authConfig` is required. It should never be null, at the cost of crashing the application earlier.
Otherwise we should provide a meaningful default

frontend/src/contexts/Auth.tsx
```diff
@@ -0,0 +1,184 @@
...
+  }
+`;
+
+const AUTH_CONFIG_VERSION = 1;
+
+function saveAuthConfig(authConfig?: AuthConfig | null): void {
+  if (!authConfig) {
+    localStorage.removeItem("authConfig");
+  } else {
+    localStorage.setItem(
```
Pavinati, line 52: We've been relying heavily on local storage for persistentcy across tabs/sessions. For users that don't want to persist their session or in general to lay out the basis for the GDPR handling, I think it's time to use cookies.

Pavinati (approved):

PR #127, opened 2022-02-07
Add basic authentication to GraphQL API

rbino:
Uses the `e_tga` claim encoded in the JWT. Right now the auth mechanism just verifies the correctness of the signature and the existence of the claim, without using its value.


backend/priv/repo/migrations/20220202171522_add_public_key_to_tenants.exs
```diff
@@ -0,0 +1,9 @@
+defmodule Edgehog.Repo.Migrations.AddPublicKeyToTenants do
+  use Ecto.Migration
+
+  def change do
+    alter table(:tenants) do
+      add :public_key, :text, null: false
```
Pavinati, line 6: `null: false` will make this fail for all existing deployments.
That got me thinking, could the authentication be enabled/disabled per tenant?
Instead of relying on a single env variable, could the absence of public key on the tenant be our flag for enabling/disabling authentication?

Pavinati (approved):

davidebriani (approved):

szakhlypa (approved):

bettio (approved):

PR #128, opened 2022-02-07
[Frontend] Manual OTA update improvements

Pavinati:
- reuse generated relay types
- hide full images URLs in favor of just the image name


frontend/src/components/OperationTable.tsx
```diff
@@ -16,71 +16,92 @@
...
+        <FormattedMessage
+          id={messages[status]?.id || messages.UnknownStatus.id}
+        />
       </span>
     </div>
   );
 };
 
-const columns: Column<OTAOperation>[] = [
+type TableRecord = OperationTable_otaOperations["otaOperations"][0];
```
davidebriani, line 102: We could just call it `OtaOperation` instead of `TableRecord`

frontend/src/components/OperationTable.tsx
```diff
@@ -113,7 +134,7 @@ const columns: Column<OTAOperation>[] = [
     ),
     Cell: ({ value }) => (
       <FormattedDate
-        value={value}
+        value={new Date(value)}
```
davidebriani, line 137: `FormattedDate` will happily accepts the value without pre-parsing it into a `Date`

frontend/src/components/OperationTable.tsx
```diff
@@ -145,17 +166,41 @@ const columns: Column<OTAOperation>[] = [
...
 };
 
 const initialSortedColumns = [{ id: "updatedAt", desc: true }];
 
-const OperationTable = ({ className, data }: OperationTableProps) => {
+const OperationTable = ({ className, deviceRef }: OperationTableProps) => {
+  const data = useFragment(OPERATION_TABLE_FRAGMENT, deviceRef);
+
+  // react table requires mutable data
+  // TODO change this only a solution has been found
```
davidebriani, line 178: change this **once** a solution has been found

frontend/src/components/OperationTable.tsx
```diff
@@ -16,71 +16,92 @@
...
 import Icon from "components/Icon";
 import Table from "components/Table";
 import type { Column } from "components/Table";
-import type { OTAOperation, OTAOperationStatus } from "types/OTAUpdate";
 
-type OperationStatusProps = {
-  status: OTAOperationStatus | null;
-};
+const OPERATION_TABLE_FRAGMENT = graphql`
+  fragment OperationTable_otaOperations on Device {
```
davidebriani, line 33: This can be fine but we would have to revert it if we added a page with the list of all ota operations, since the fragment would not be on `Device` anymore.
Just adding this note on the general pattern of mixing GraphQL with inner React components

davidebriani (changes requested):

frontend/src/components/OperationTable.tsx
Pavinati: If the fragment comes from a different node, for example the list of all operations, it may have access to additional information, like the target device. In general over generalisation may have other problems. I think it's best to evaluate what to do case by case instead of walking the preemptive optimization path

frontend/src/components/OperationTable.tsx
davidebriani: Alright, let's settle with that

davidebriani (approved):

bettio (approved):

bettio (approved):

PR #137, opened 2022-02-21
WIP: Add update device api

davidebriani:
Add a GraphQL mutation to update info for a device. As of now, only the name property can be updated.
Using this `updateDevice` mutation, the frontend Device page is modified to be able to update the name of devices.


frontend/src/pages/Device.tsx
```diff
@@ -983,6 +1003,59 @@ const DeviceContent = ({ getDeviceQuery }: DeviceContentProps) => {
     [deviceData.device]
   );
 
+  const [deviceDraft, setDeviceDraft] = useState(_.pick(device, ["name"]));
```
Pavinati, line 1006: why not just `useState(device.name)`?

davidebriani: Having just a single property to track, we can do that. Here I kept the same pattern we used in the other similar situations, where we keep a single object to track changes to the resource to be updated. Adding a new property to track would be trivial this way. I don't have a preference though, you tell me

Pavinati: Oh right, this is an object with a field `name`, not just a string.
I suppose in the near future we're likely to add more fields to update so i guess that's ok.

backend/lib/edgehog_web/resolvers/astarte.ex
```diff
@@ -47,6 +47,16 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     {:ok, devices}
   end
 
+  def update_device(_parent, %{device_id: id} = attrs, %{context: context}) do
+    device = Astarte.get_device!(id)
+    attrs = Map.delete(attrs, :device_id)
+
+    with {:ok, device} <- Astarte.update_device(device, attrs) do
+      device = device |> preload_system_model_for_device(context)
```
rbino, line 55: Style: no need to pipe for a single function call
```elixir
device = preload_system_model_for_device(device, context)
```

Pavinati (approved):

backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -480,5 +480,24 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
...
+        field :name, :string
+      end
+
+      output do
+        @desc "The updated device."
+        field :device, non_null(:device)
+      end
+
+      middleware Absinthe.Relay.Node.ParseIDs, device_id: :device
+      resolve &Resolvers.Astarte.update_device/3
```
rbino, line 500: Like @szakhlypa noted, use `update_device/2` and drop the first argument if you're not using the parent

backend/lib/edgehog_web/resolvers/astarte.ex
```diff
@@ -47,6 +47,16 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     {:ok, devices}
   end
 
+  def update_device(_parent, %{device_id: id} = attrs, %{context: context}) do
+    device = Astarte.get_device!(id)
+    attrs = Map.delete(attrs, :device_id)
```
rbino, line 52: You shouldn't manually filter stuff from `attrs` here, that is the responsibility of the `changeset` function.
If your intention is "I don't want the user to be able to change the `device_id` when updating the device", then you need to separate changeset for creation and update (calling the right one for each situation in the `Astarte` context).

rbino (changes requested):

rbino (approved):

szakhlypa (approved):

PR #143, opened 2022-02-23
[frontend] Device Led behavior

### LINKED ISSUE #105: Add Button to Frontend for Blinking the Indicator Led
Add a button that makes the onboard indicator led blink. The button should be hidden or not visible when the led interface is not available.
The purpose of this feature is to allow finding/identify a specific board among a number of boards using a visual feedback, and to perform a quick connectivity test.

szakhlypa:
Closes #105 

Adds `Identify` dropdown on Device page.
device_connected

Disables dropdown for disconnected device.
device_disconnected

Shows tooltip
tooltip

Disables dropdown while mutation in flight and maybe shows error

https://user-images.githubusercontent.com/26611935/156815870-3fcff0e3-118c-4e05-a012-e0ee46a76543.mp4

Shows selected LED behavior after mutation success for 10 seconds.

https://user-images.githubusercontent.com/26611935/156816027-3febd1fc-8c04-4af3-a8a0-35ee53aae9a4.mp4

### DIFF
--- a/frontend/src/components/LedBehaviorDropdown.tsx
+++ b/frontend/src/components/LedBehaviorDropdown.tsx
@@ -1,0 +1,130 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2022 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+*/
+
+import { useCallback } from "react";
+import graphql from "babel-plugin-relay/macro";
+import { useMutation } from "react-relay/hooks";
+import { defineMessages, FormattedMessage, useIntl } from "react-intl";
+import type { MessageDescriptor } from "react-intl";
+
+import ButtonGroup from "react-bootstrap/ButtonGroup";
+import Dropdown from "react-bootstrap/Dropdown";
+
+import Spinner from "components/Spinner";
+
+import type { LedBehaviorDropdown_setLedBehavior_Mutation } from "api/__generated__/LedBehaviorDropdown_setLedBehavior_Mutation.graphql";
+
+const SET_LED_BEHAVIOR_MUTATION = graphql`
+  mutation LedBehaviorDropdown_setLedBehavior_Mutation(
+    $input: SetLedBehaviorInput!
+  ) {
+    setLedBehavior(input: $input) {
+      behavior
+    }
+  }
+`;
+
+const SUPPORTED_LED_BEHAVIORS = [
+  "BLINK",
+  "DOUBLE_BLINK",
+  "SLOW_BLINK",
+] as const;
+type SupportedLedBehavior = typeof SUPPORTED_LED_BEHAVIORS[number];
+
+const supportedBehaviorMessages: Record<
+  SupportedLedBehavior,
+  MessageDescriptor
+> = defineMessages({
+  BLINK: {
+    id: "components.LedBehaviorDropdown.behavior.blinkLED",
+    defaultMessage: "Blink LED",
+  },
+  DOUBLE_BLINK: {
+    id: "components.LedBehaviorDropdown.behavior.doubleBlinkLED",
+    defaultMessage: "Double blink LED",
+  },
+  SLOW_BLINK: {
+    id: "components.LedBehaviorDropdown.behavior.slowBlinkLED",
+    defaultMessage: "Slow blink LED",
+  },
+});
+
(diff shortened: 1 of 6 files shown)


frontend/src/components/LedBehaviorDropdown.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const [setLedBehavior, isSettingLedBehavior] =
+    useMutation<LedBehaviorDropdown_setLedBehavior_Mutation>(
+      SET_LED_BEHAVIOR_MUTATION
+    );
+
+  const handleSetLedBehavior = useCallback(
+    (ledBehavior: string | null) => {
+      if (ledBehavior == null || !isLedBehavior(ledBehavior)) {
+        return;
+      }
```
davidebriani, line 63: Maybe you can avoid the `null` check here by accepting `unknown` in `isLedBehavior`:
```
function isLedBehavior(value: unknown): value is LedBehavior {
```

Pavinati (approved):

frontend/src/components/LedBehaviorDropdown.tsx
szakhlypa: You're right, thanks

frontend/src/components/LedBehaviorDropdown.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+`;
+
+interface Props {
+  deviceId: string;
+  onError: (error: React.ReactNode) => void;
+}
+
+function isLedBehavior(value: unknown): value is LedBehavior {
```
Pavinati, line 49: `isLedBehavior` is a bit misleading considering future added values.
If we add a new value to the backend enum, this fail to check the newly added value.
I suggest renaming this function to `isSupportedLedBehavior` or something else like:
```typescript
const supportedLedBehaviors: LedBehavior[] = ["BLINK", "DOUBLE_BLINK", "SLOW_BLINK"];

const parseLedBehaviorInput = (value: string | null): LedBehavior | null => {
  return (value && supportedLedBehaviors.includes(value) ? value : null);
};
```
i would avoid reflection (`typeof`) and guards when not needed
@davidebriani what do you think?

davidebriani: Yep, I am fine with the renaming `isLedBehavior` -> `isSupportedLedBehavior`.
However in both the code snippets you have the parsed value typed as `LedBehavior`, which contains future added values. Without going crazy, I'd prefer the `isSupportedLedBehavior` over `parseLedBehaviorInput`, since the function is used to validate a condition and exit early rather than using the returned value typed as `LedBehavior`.

Pavinati: makes sense :+1:

davidebriani (approved):

frontend/src/components/LedBehaviorDropdown.tsx
```diff
@@ -0,0 +1,190 @@
...
+    id: "components.LedBehaviorDropdown.behavior.slowBlinkLED",
+    defaultMessage: "Slow blink LED",
+  },
+});
+
+function isSupportedLedBehavior(value: unknown): value is SupportedLedBehavior {
+  return (
+    typeof value === "string" &&
+    SUPPORTED_LED_BEHAVIORS.includes(value as SupportedLedBehavior)
+  );
```
Pavinati, line 75: here we don't know if `value` is of type `SupportedLedBehavior` yet. We should cast it to `string` instead.
Maybe the TypeScript compiler/linter is intelligent enough to deduce it from the `typeof value === "string"` check.
Otherwise you can use `string | null` in the function signature. We won't get other values from GraphQL responses. That is also checked by relay typescript compiler, so if for any reason value could take values other than string, we will know where to handle them.

frontend/src/components/LedBehaviorDropdown.tsx
```diff
@@ -0,0 +1,190 @@
...
+          );
+        },
+      });
+    },
+    [setLedBehavior, deviceId, onError, setCurrentBehavior]
+  );
+
+  if (currentBehavior) {
+    return (
+      <Button variant="success" active>
```
Pavinati, line 147: This button has no `onClick` handler. Should it display as disabled? Or maybe with a <div> instead of a button?
Having a clickable button that doesn't do anything when clicked could be a usability issue (minor).
Also the `:active` pseudo-class is only used when the button is currently being pressed (with the primary mouse button, keyboard or some Assistive Technology) by the user. It doesn't represent the button being connected to some backend operation.

Pavinati (commented):
I left a few nitpicking comments, but overall looks good to me.

frontend/src/components/LedBehaviorDropdown.tsx
szakhlypa: Typescript doesn't allow `ReadonlyArray<SupportedLedBehavior>.includes(string)` [link](https://github.com/microsoft/TypeScript/issues/14520#issuecomment-853946018) 

We have two options:
1. `typeof value === "string" && SUPPORTED_LED_BEHAVIORS.includes(value as SupportedLedBehavior)`
2. `typeof value === "string" && (SUPPORTED_LED_BEHAVIORS as readonly string[]).includes(value)`

Both transpiled to `typeof value === "string" && SUPPORTED_LED_BEHAVIORS.includes(value)` in JS.

I have no bias what to cast. Do you have a preference?

PR #149, opened 2022-03-08
Handle device missing interfaces with device capabilities query

### LINKED ISSUE #112: Handle Missing Interfaces
Edgehog should work also when some interfaces (or most of them) are not implemented on the device side.
The frontend should show less tabs in that case.

Pavinati:
Allow the frontend or other GrahQL clients to query the device capabilities. This is a useful information if you want to know what the device actually supports and act accordingly.

Closes #112

Signed-off-by: Mattia Pavinati <mattia.pavinati@secomind.com>

### DIFF
--- a/backend/lib/edgehog/astarte.ex
+++ b/backend/lib/edgehog/astarte.ex
@@ -26,6 +26,7 @@ defmodule Edgehog.Astarte do
 
   alias Astarte.Client.AppEngine
   alias Edgehog.Astarte.Cluster
+  alias Edgehog.Astarte.InterfaceID
 
   alias Edgehog.Astarte.Device.{
     BaseImage,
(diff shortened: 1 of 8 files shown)


szakhlypa (changes requested):

backend/lib/edgehog/astarte.ex
```diff
@@ -90,6 +90,47 @@ defmodule Edgehog.Astarte do
...
+    ],
+    storage: [
+      %{name: "io.edgehog.devicemanager.StorageUsage", major: 0, minor: 1}
+    ],
+    battery_status: [
+      %{name: "io.edgehog.devicemanager.BatteryStatus", major: 0, minor: 1}
+    ],
+    cellular_connection: [
+      %{name: "io.edgehog.devicemanager.CellularConnectionStatus", major: 0, minor: 1},
+      %{name: "io.edgehog.devicemanager.CellularConnectionProperties", major: 0, minor: 1}
```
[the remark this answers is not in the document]
Pavinati: I confronted with @FrancescoVaiani to have a clearer view on what those interfaces so, and we agreed it doesn't make much sense to have a device that gives you hardware information about its modem but doesn't say whether is connected or no to something. The devices should have both interfaces to support said capability, and if disconnected it won't publish data on the status one.

backend/lib/edgehog_web/resolvers/astarte.ex
```diff
@@ -73,6 +73,13 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     Astarte.preload_system_model_for_device(target, preload: preload)
   end
 
+  def fetch_device_capabilities(%Device{} = device, _args, _context) do
+    case Astarte.fetch_device_introspection(device) do
+      {:ok, introspection} -> {:ok, Astarte.get_device_capabilities(introspection)}
+      _ -> {:ok, []}
+    end
+  end
+
```
rbino, line 82: Since this function cannot fail by definition and it always returns a list, it's more akin to a `list` function rather than a `fetch` function, so I would call it `list_device_capabilities` and just return a list without the wrapping tuple

backend/lib/edgehog/astarte.ex
```diff
@@ -683,4 +744,34 @@ defmodule Edgehog.Astarte do
...
+          [capability | acc]
+        else
+          acc
+        end
+      end)
+
+    # TODO add checks on device privacy settings and geolocation providers
+    capabilities = [:geolocation | capabilities]
+
+    capabilities
```
rbino, line 761: Just
```
[:geolocation | capabilities]
```

backend/test/edgehog/astarte_test.exs
```diff
@@ -602,5 +602,70 @@ defmodule Edgehog.AstarteTest do
...
+      }
+
+      partial_introspection_2 = %{
+        "io.edgehog.devicemanager.OTAResponse" => %{"major" => 0, "minor" => 1}
+      }
+
+      assert not Enum.member?(
+               Astarte.get_device_capabilities(partial_introspection_1),
+               :software_updates
+             )
```
rbino, line 658: You can do
```elixir
assert :software_updates not in Astarte.get_device_capabilities(partial_introspection_1)
```
it reads more fluently

rbino (changes requested):

backend/lib/edgehog_web/resolvers/astarte.ex
```diff
@@ -73,6 +73,13 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     Astarte.preload_system_model_for_device(target, preload: preload)
   end
 
+  def list_device_capabilities(%Device{} = device, _args, _context) do
+    case Astarte.fetch_device_introspection(device) do
+      {:ok, introspection} -> Astarte.get_device_capabilities(introspection)
+      _ -> []
```
rbino, line 79: Should this be `[:geolocation]` even without an introspection since it is added unconditionally in `get_device_capabilities`?

Pavinati: my reasoning was: If i'm able to obtain the introspection, i won't get the last seen IP too, hence no geolocalization

rbino (approved):

szakhlypa (approved):

bettio (approved):

PR #171, opened 2022-03-16
Add dialyzer checks to backend workflow

### LINKED ISSUE #169: Add support to dialyzer to backend

Pavinati:
Use dialyzer to statically analyze the elixir backend code
Closes #169


Pavinati:
After an internal discussion, we agreed to use `dialyxir` instead of `dialyzex`.
The key reasons in favor of dialyxir are
- it now supports global LPT cache
- appears to be more popular or more active on GitHub

bettio (approved):

.github/workflows/backend-test.yaml
```diff
@@ -38,6 +38,60 @@ env:
...
+    # Cache key based on Elixir & Erlang version (also useful when running in matrix)
+    - name: Restore PLT cache
+      uses: actions/cache@v2
+      id: plt_cache
+      with:
+        path: backend/priv/plts
+        key: |
+          ${{ runner.os }}-${{ env.elixir_version }}-${{ env.otp-version }}-plt
+        restore-keys: |
+          ${{ runner.os }}-${{ env.elixir_version }}-${{ env.otp-version }}-plt
```
rbino, line 83: Let's be consistent with the other stuff and use
```yaml
        key: |
          ${{ runner.os }}-dialyzer-plt-${{ env.otp_version }}-${{ env.elixir_version }}-${{ github.sha }}
        restore-keys: |
          ${{ runner.os }}-dialyzer-plt-${{ env.otp_version }}-${{ env.elixir_version }}-
```

rbino (changes requested):

backend/mix.exs
```diff
@@ -37,6 +37,9 @@ defmodule Edgehog.MixProject do
         "coveralls.detail": :test,
         "coveralls.post": :test,
         "coveralls.html": :test
+      ],
+      dialyzer: [
+        plt_file: {:no_warn, "priv/plts/dialyzer.plt"}
```
szakhlypa, line 42: Should we also add `plt_core_path: "priv/plts",` to store PLTs of core Erlang and Elixir files in cache?

szakhlypa (changes requested):

szakhlypa (approved):

backend/mix.exs
```diff
@@ -37,6 +37,10 @@ defmodule Edgehog.MixProject do
         "coveralls.detail": :test,
         "coveralls.post": :test,
         "coveralls.html": :test
+      ],
+      dialyzer: [
+        plt_core_path: "priv/plts",
```
rbino, line 42: As discussed by voice, this should be treated like we treat [`dialyzer_cache_directory`](https://github.com/astarte-platform/astarte/blob/master/apps/astarte_appengine_api/mix.exs#L36) in Astarte, because outside of the CI we want core PLTs to actually be shared in Mix home

rbino (changes requested):

backend/config/ci.exs
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+use Mix.Config
```
rbino, line 20: This should be `import Config`

rbino (changes requested):

backend/config/ci.exs
rbino: Also: there shoulde be an empty line between the last comment line in the header and the first code line

rbino (approved):

bettio (approved):

PR #191, opened 2022-05-18
Add support for device tags

rbino:
Allow tagging a device, showing its tags when querying it and filtering for a specific tag when requesting the device list.

### DIFF
--- a/backend/lib/edgehog/astarte.ex
+++ b/backend/lib/edgehog/astarte.ex
@@ -27,9 +27,11 @@ defmodule Edgehog.Astarte do
   alias Edgehog.Repo
 
   alias Astarte.Client.AppEngine
+  alias Ecto.Multi
   alias Edgehog.Astarte.Cluster
   alias Edgehog.Astarte.InterfaceID
   alias Edgehog.Astarte.InterfaceVersion
+  alias Edgehog.Devices
 
   alias Edgehog.Astarte.Device.{
     BaseImage,
@@ -518,9 +520,26 @@ defmodule Edgehog.Astarte do
 
   """
   def update_device(%Device{} = device, attrs) do
-    device
-    |> Device.update_changeset(attrs)
-    |> Repo.update()
+    Multi.new()
+    |> Devices.ensure_tags_exist_multi(attrs)
+    |> Multi.update(:update_device, fn
+      %{ensure_tags_exist: nil} ->
+        device
+        |> Device.update_changeset(attrs)
+
+      %{ensure_tags_exist: tags} when is_list(tags) ->
+        device
+        |> Device.update_changeset(attrs)
+        |> Ecto.Changeset.put_assoc(:tags, tags)
+    end)
+    |> Repo.transaction()
+    |> case do
+      {:ok, %{update_device: device}} ->
+        {:ok, Repo.preload(device, :tags)}
+
+      {:error, _failed_operation, failed_value, _progress_so_far} ->
+        {:error, failed_value}
+    end
   end
 
   @doc """
--- a/backend/lib/edgehog/astarte.ex
+++ b/backend/lib/edgehog/astarte.ex
@@ -378,6 +378,7 @@ defmodule Edgehog.Astarte do
     filters
     |> Enum.reduce(Device, &filter_with/2)
     |> Repo.all()
+    |> Repo.preload(:tags)
   end
 
   defp filter_with(filter, query) do
@@ -472,6 +473,7 @@ defmodule Edgehog.Astarte do
   """
   def get_device!(id) do
     Repo.get!(Device, id)
+    |> Repo.preload(:tags)
   end
 
   @doc """
@@ -487,9 +489,13 @@ defmodule Edgehog.Astarte do
 
   """
   def create_device(%Realm{} = realm, attrs \\ %{}) do
-    %Device{realm_id: realm.id, tenant_id: Repo.get_tenant_id()}
-    |> Device.changeset(attrs)
-    |> Repo.insert()
+    changeset =
+      %Device{realm_id: realm.id, tenant_id: Repo.get_tenant_id()}
+      |> Device.changeset(attrs)
+
+    with {:ok, device} <- Repo.insert(changeset) do
+      {:ok, Repo.preload(device, :tags)}
+    end
   end
 
   @doc """
@@ -585,7 +591,7 @@ defmodule Edgehog.Astarte do
   """
   def fetch_realm_device(%Realm{id: realm_id}, device_id) do
     case Repo.get_by(Device, realm_id: realm_id, device_id: device_id) do
-      %Device{} = device -> {:ok, device}
+      %Device{} = device -> {:ok, Repo.preload(device, :tags)}
       nil -> {:error, :device_not_found}
     end
   end
--- a/backend/lib/edgehog/astarte.ex
+++ b/backend/lib/edgehog/astarte.ex
@@ -414,6 +414,20 @@ defmodule Edgehog.Astarte do
       {:hardware_type_name, name} ->
         from [hardware_type: ht] in ensure_hardware_type(query),
           where: ilike(ht.name, ^"%#{name}%")
+
+      {:tag, tag} ->
+        # Need to filter tenant explicitly since prepare_query is not executed for subqueries
+        tenant_id = Repo.get_tenant_id()
+
+        device_ids_matching_tag =
+          from t in Devices.Tag,
+            where: ilike(t.name, ^"%#{tag}%") and t.tenant_id == ^tenant_id,
+            join: dt in Devices.DeviceTag,
+            on: t.id == dt.tag_id,
+            select: dt.device_id
+
+        from q in query,
+          where: q.id in subquery(device_ids_matching_tag)
     end
   end
 
--- a/backend/lib/edgehog/devices.ex
+++ b/backend/lib/edgehog/devices.ex
@@ -32,6 +32,7 @@ defmodule Edgehog.Devices do
   alias Edgehog.Devices.SystemModelPartNumber
   alias Edgehog.Devices.HardwareType
   alias Edgehog.Devices.HardwareTypePartNumber
+  alias Edgehog.Devices.Tag
   alias Edgehog.Assets
 
   @doc """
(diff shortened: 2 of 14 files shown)


backend/priv/repo/migrations/20220517063737_create_devices_tags.exs
```diff
@@ -22,16 +22,23 @@ defmodule Edgehog.Repo.Migrations.CreateDevicesTags do
...
-      add :tag_id, references(:tags, on_delete: :nothing)
-      add :device_id, references(:devices, on_delete: :nothing)
+    create table(:devices_tags, primary_key: false) do
+      add :tenant_id, references(:tenants, column: :tenant_id, on_delete: :delete_all),
+        null: false,
+        primary_key: true
 
-      timestamps()
+      add :tag_id,
+          references(:tags, with: [tenant_id: :tenant_id], match: :full, on_delete: :nothing),
```
davidebriani, line 31: I think we can use `on_delete: :delete_all`. It doesn't make sense to keep the Tag -> Device association when the referenced tag doesn't exist.
Or, if you want to use custom logic when deleting a used tag, we can use `on_delete: :restrict`

backend/lib/edgehog/devices/device_tag.ex
```diff
@@ -22,12 +22,11 @@ defmodule Edgehog.Devices.DeviceTag do
...
   import Ecto.Changeset
 
+  @primary_key false
   schema "devices_tags" do
-    field :tenant_id, :id
-    field :tag_id, :id
-    field :device_id, :id
-
-    timestamps()
+    field :tenant_id, :integer, autogenerate: {Edgehog.Repo, :get_tenant_id, []}
```
davidebriani, line 27: Shouldn't `:tenant_id` be an `:id` instead of `:integer`?

backend/lib/edgehog/astarte.ex
```diff
@@ -518,9 +520,26 @@ defmodule Edgehog.Astarte do
 
   """
   def update_device(%Device{} = device, attrs) do
-    device
-    |> Device.update_changeset(attrs)
-    |> Repo.update()
+    Multi.new()
+    |> Devices.ensure_tags_multi(attrs)
+    |> Multi.update(:device, fn
+      %{tags: nil} ->
```
davidebriani, line 546: Naming each Multi step with something that matches the performed action would make the logic easier to follow.

For instance, the step that updates the device could be named `:update_device`, so we can maintain a uniform naming between `update_device/1`, `update_device_multi/2` and the `:update_device`  step that would contain the end result of the Multi.
With such convention, I could already expect `ensure_tags_multi()` to contain a step `:ensure_tags` with the end result for that Multi.

backend/lib/edgehog/devices.ex
```diff
@@ -420,4 +421,56 @@ defmodule Edgehog.Devices do
   def change_system_model(%SystemModel{} = system_model, attrs \\ %{}) do
     SystemModel.changeset(system_model, attrs)
   end
+
+  @doc """
+  Inserts the tags passed in attrs within a multi transaction, normalizing them.
+
+  Returns the updated `%Ecto.Multi{}`.
+  """
+  def ensure_tags_multi(multi, %{tags: _tags} = attrs) do
```
davidebriani, line 430: `ensure_tags_multi` does not convey a clear meaning for this function, we could use something like `upsert_tags_multi`.

backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -406,6 +413,11 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
     @desc "The system model of the device."
     field :system_model, :system_model
 
+    @desc "The tags of the device"
+    field :tags, non_null(list_of(non_null(:string))) do
+      resolve &Resolvers.Devices.extract_device_tags/3
```
davidebriani, line 418: What about just supporting the serialization of Tags to string instead of manually using `extract_device_tags/3`:

```elixir
defimpl String.Chars, for: Edgehog.Devices.Tag do
  def to_string(%Edgehog.Devices.Tag{name: name}), do: name
end
```
This way you won't need to replicate the serialization should you wish to expose tags on some other queries.

davidebriani (changes requested):

backend/lib/edgehog/devices/device_tag.ex
rbino: Explicit `autogenerate` tuples do not work with `id` fields, so this has to be an `integer`.
This is the new status quo to populate the `tenant_id` (I want to gradually introduce it everywhere (it's already used in `OtaOperation`), because it has the advantage of allowing application code to override the `tenant_id` if there's the need (e.g. in tests) by passing a struct with the `tenant_id` already populated while still automatically populating it with the saved `tenant_id` in the average case

backend/lib/edgehog/devices/device_tag.ex
```diff
@@ -0,0 +1,38 @@
...
+    field :tag_id, :id, primary_key: true
+    field :device_id, :id, primary_key: true
+  end
+
+  @doc false
+  def changeset(device_tag, attrs) do
+    device_tag
+    |> cast(attrs, [])
+    |> validate_required([])
```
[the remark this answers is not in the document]
rbino: No, in fact we might as well remove the whole changeset since this is just used as a support schema for associations and we never create it explicitly.

backend/lib/edgehog/astarte.ex
rbino: I checked and it seems there's no established convention for naming multi steps, what you proposed makes total sense, changed the code to reflect it

backend/lib/edgehog/devices.ex
rbino: Actually I think that `ensure` is better than `upsert` here: there's actually no update happening, what this steps are doing is that if tags are not there, they insert them. It's akin to what `ensure_device_exists` is doing in some other part of the codebase, hence the similar name.
If you have another verb that conveys the same meaning I'm open to suggestions.

backend/lib/edgehog/astarte.ex
davidebriani: Another reason for adopting a naming convention would be that you cannot have steps with the same name in a Multi. So following a convention with specific names for specific actions can save you from conflicts when merging Multis

backend/lib/edgehog/devices.ex
davidebriani: You're right. Indeed the meaning is clear if you use the verb too, i.e. `ensure_tags_exist_multi`

backend/lib/edgehog/devices/tag.ex
```diff
@@ -0,0 +1,39 @@
...
+
+    timestamps()
+  end
+
+  @doc false
+  def changeset(tag, attrs) do
+    tag
+    |> cast(attrs, [:name])
+    |> validate_required([:name])
+    |> unique_constraint(:name)
```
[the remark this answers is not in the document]
rbino: Correct

backend/lib/edgehog_web/schema/astarte_types.ex
rbino: Explicit is better than implicit, the explicit resolver is still reusable with other queries and it doesn't force the same Tag string representation application-wide

backend/priv/repo/migrations/20220517063737_create_devices_tags.exs
rbino: I'll use `:restrict`. Tags in the `tags` table are basically thought to be append-only (i.e. only the association is deleted if the tag is deleted), so I'd rather disallow them from being deleted than cascade (since they should normally not be deleted anyway).

backend/test/edgehog/astarte_test.exs
```diff
@@ -460,10 +460,25 @@ defmodule Edgehog.AstarteTest do
...
+    end
+
+    test "update_device/2 normalizes and deduplicates tags", %{realm: realm} do
+      device = device_fixture(realm)
+      update_attrs = %{tags: ["sTRANGE", "taGs   ", "  DUPlicate", "DUPLICATE"]}
+      assert {:ok, %Device{} = device} = Astarte.update_device(device, update_attrs)
+
+      assert ["strange", "tags", "duplicate"] == Enum.map(device.tags, & &1.name)
+    end
+
```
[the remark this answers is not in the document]
rbino: Fixed adding `on_replace: :delete` to the association, I also added your new tests to the suite

backend/lib/edgehog/devices/tag.ex
```diff
@@ -0,0 +1,39 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Devices.Tag do
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  schema "tags" do
+    field :name, :string
+    field :tenant_id, :id
```
davidebriani, line 27: Could we use :autogenerate here too?
```elixir
field :tenant_id, :integer, autogenerate: {Edgehog.Repo, :get_tenant_id, []}
```

rbino: I'll add it for consistency, but currently tags are inserted using `Repo.insert_all` which doesn't populate `autogenerate` fields, so we still have to populate it manually in `ensure_tags_exist_multi`

backend/test/edgehog/astarte_test.exs
```diff
@@ -460,10 +460,49 @@ defmodule Edgehog.AstarteTest do
...
+    test "update_device/2 removes tags", %{realm: realm} do
+      device = device_fixture(realm)
+      update_attrs = %{tags: ["some", "tags"]}
+
+      assert {:ok, %Device{} = device} = Astarte.update_device(device, update_attrs)
+      assert ["some", "tags"] == Enum.map(device.tags, & &1.name)
+
+      update_attrs = %{tags: ["new"]}
+      assert {:ok, %Device{} = device} = Astarte.update_device(device, update_attrs)
+      assert ["new"] == Enum.map(device.tags, & &1.name)
```
szakhlypa, line 487: Should we also check that unused tags were deleted?
It seems they are not.
`assert length(Repo.all(Edgehog.Devices.Tag)) == 1` fails with 3 != 1

szakhlypa (changes requested):

backend/test/edgehog/astarte_test.exs
rbino: Unused tags shouldn't get deleted - only their association. Basically the `tags` table is append-only. This is how it's usually done in all examples I've found.

davidebriani (approved):

backend/lib/edgehog/astarte.ex
```diff
@@ -411,6 +414,17 @@ defmodule Edgehog.Astarte do
       {:hardware_type_name, name} ->
         from [hardware_type: ht] in ensure_hardware_type(query),
           where: ilike(ht.name, ^"%#{name}%")
+
+      {:tag, tag} ->
+        device_ids_matching_tag =
+          from t in Devices.Tag,
+            where: ilike(t.name, ^"%#{tag}%"),
```
szakhlypa, line 421: Just for my understanding, does [`prepare_query`](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/repo.ex#L39) works for subquery too?
Or should we also filter Devices.Tag with `tenant_id` in subquery?

szakhlypa (approved):

backend/lib/edgehog/astarte.ex
rbino: Good catch, I checked the generated query and indeed `prepare_query` was not executed for the subquery, fixed this by explicitly matching `tenant_id` in the subquery.

szakhlypa (approved):

davidebriani (approved):

bettio (changes requested):
Let's add something to the changelog

PR #193, opened 2022-05-26
Accentuate last connected WiFi AP

### LINKED ISSUE #192: Display connected WiFi network

szakhlypa:
Closes #192


bettio:
What does it happen if a device has not the latest interface version and it doesn't send the connected information? Are we handling this scenario properly?

szakhlypa:
`wifiScanResults` GraphQL object has nullable `connected` field, so if device has `io.edgehog.devicemanager.WiFiScanResults` interface without `connected` endpoint, no one from scan results will be accentuated with bold.

rbino (approved):

PR #201, opened 2022-06-06
Fix queries and mutations for translated descriptions

davidebriani:
Currently, saving translated descriptions for System Models doesn't work when the browser displaying the Edgehog's frontend has a preferred locale that differs from the default one configured for the tenant. Indeed, browsers can query descriptions for a custom locale, but they can only save descriptions in the tenant's default locale.

Sample steps to reproduce the issue:
1. The browser queries a System Model asking for its translated description with, an `it-IT` locale.
2. The backend returns the System Model without a description because there is none.
3. The browser tries to update the System Model with a description in the tenant's default locale: `en-US`.
4. The subsequent query to get the updated System Model still asks for a description in `it-IT`, the preferred locale for the browser, so the backend still returns no description. Although it is persisted, the previously inserted description seems to have disappeared from the user's point of view.

The change proposed here fixes the odd behavior in the backend by:
- Preloading both the default locale of the tenant and the locale specified by the user, if any.
- Extracting the best match of translated descriptions according to locale preferences when rendering System Models in the GraphQL layer.
This way, the user is always presented with a translated description for one of the locales, if it exists at least for the tenant's default locale.

Then another fix is added to support removing previously entered descriptions:
1. Allow the frontend to pass blank descriptions.
2. Make sure the backend discards and remove empty descriptions.

With these changes in place, users can now correctly see and manage descriptions at least in the tenant's locale until a full i18n support is implemented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Start using logfmt as logging format.
 
+### Fixed
+- Use the tenant's default locale when managing translated descriptions if the user's locale is not
+available.
+
 ## [0.5.1] - 2022-06-01
 ### Added
 - Add `connected` field to wifi scan result and highlight the latest connected network
--- a/backend/lib/edgehog/devices/system_model_description.ex
+++ b/backend/lib/edgehog/devices/system_model_description.ex
@@ -44,8 +44,8 @@ defmodule Edgehog.Devices.SystemModelDescription do
     |> unique_constraint([:locale, :system_model_id, :tenant_id])
   end
 
-  def localized(locale) do
+  def localized(locales) when is_list(locales) do
     from d in SystemModelDescription,
-      where: d.locale == ^locale
+      where: d.locale in ^locales
   end
 end
--- a/backend/lib/edgehog_web/context.ex
+++ b/backend/lib/edgehog_web/context.ex
@@ -21,6 +21,8 @@
 defmodule EdgehogWeb.Context do
   @behaviour Plug
 
+  alias Edgehog.Tenants.Tenant
+
   def init(opts), do: opts
 
   def call(conn, _opts) do
@@ -28,31 +30,31 @@ defmodule EdgehogWeb.Context do
     Absinthe.Plug.put_options(conn, context: context)
   end
 
-  defp build_context(conn) do
+  def build_context(conn) do
     current_tenant = get_current_tenant(conn)
+    preferred_locales = get_preferred_locales(conn, current_tenant)
 
-    %{current_tenant: current_tenant}
-    |> maybe_put_locale(conn)
+    %{current_tenant: current_tenant, preferred_locales: preferred_locales}
   end
 
   defp get_current_tenant(conn) do
     conn.assigns[:current_tenant]
   end
 
-  defp maybe_put_locale(context, conn) do
-    case Plug.Conn.get_req_header(conn, "accept-language") do
-      # If no header or *, we don't add an explicit locale
-      [] ->
-        context
+  defp get_preferred_locales(conn, %Tenant{} = tenant) do
+    default_locale = tenant.default_locale
 
-      ["*" | _] ->
-        context
+    # If there are some locales in accept-language, list them before the default locale
+    conn
+    |> Plug.Conn.get_req_header("accept-language")
+    |> Enum.flat_map(&get_locales/1)
+    |> Enum.concat([default_locale])
+    |> Enum.uniq()
+  end
 
-      # If there's one (or more) accept-language, we use the first one
-      [language | _] ->
-        [locale | _] = String.split(language, ",", parts: 2)
-
-        Map.put(context, :locale, locale)
-    end
+  defp get_locales(string) when is_binary(string) do
+    ~r/[a-z]{2,3}-[A-Z]{2}/
+    |> Regex.scan(string)
+    |> List.flatten()
   end
 end
--- a/backend/lib/edgehog_web/resolvers/astarte.ex
+++ b/backend/lib/edgehog_web/resolvers/astarte.ex
@@ -60,7 +60,7 @@ defmodule EdgehogWeb.Resolvers.Astarte do
 
   defp preload_system_model_for_device(target, %{locale: locale}) do
     # Explicit locale, use that one
-    descriptions_query = Devices.localized_system_model_description_query(locale)
+    descriptions_query = Devices.localized_system_model_description_query([locale])
     preload = [descriptions: descriptions_query, hardware_type: [], part_numbers: []]
 
     Astarte.preload_system_model_for_device(target, preload: preload)
@@ -69,7 +69,7 @@ defmodule EdgehogWeb.Resolvers.Astarte do
   defp preload_system_model_for_device(target, %{current_tenant: tenant}) do
     # Fallback
     %{default_locale: default_locale} = tenant
-    descriptions_query = Devices.localized_system_model_description_query(default_locale)
+    descriptions_query = Devices.localized_system_model_description_query([default_locale])
     preload = [descriptions: descriptions_query, hardware_type: [], part_numbers: []]
 
     Astarte.preload_system_model_for_device(target, preload: preload)
--- a/backend/lib/edgehog_web/resolvers/astarte.ex
+++ b/backend/lib/edgehog_web/resolvers/astarte.ex
@@ -58,18 +58,12 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     end
   end
 
-  defp preload_system_model_for_device(target, %{locale: locale}) do
-    # Explicit locale, use that one
-    descriptions_query = Devices.localized_system_model_description_query([locale])
-    preload = [descriptions: descriptions_query, hardware_type: [], part_numbers: []]
+  defp preload_system_model_for_device(target, context) do
+    descriptions_query =
+      context
+      |> Map.fetch!(:preferred_locales)
+      |> Devices.localized_system_model_description_query()
 
-    Astarte.preload_system_model_for_device(target, preload: preload)
-  end
-
-  defp preload_system_model_for_device(target, %{current_tenant: tenant}) do
-    # Fallback
-    %{default_locale: default_locale} = tenant
-    descriptions_query = Devices.localized_system_model_description_query([default_locale])
     preload = [descriptions: descriptions_query, hardware_type: [], part_numbers: []]
 
     Astarte.preload_system_model_for_device(target, preload: preload)
--- a/backend/lib/edgehog_web/resolvers/devices.ex
+++ b/backend/lib/edgehog_web/resolvers/devices.ex
@@ -77,16 +77,9 @@ defmodule EdgehogWeb.Resolvers.Devices do
     {:ok, system_models}
   end
 
-  defp localize_system_model_description(target, %{locale: locale}) do
-    # Explicit locale, use that one
-    Devices.preload_localized_descriptions_for_system_model(target, [locale])
-  end
-
-  defp localize_system_model_description(target, %{current_tenant: tenant}) do
-    # Fallback
-    %{default_locale: default_locale} = tenant
-
-    Devices.preload_localized_descriptions_for_system_model(target, [default_locale])
+  defp localize_system_model_description(target, context) do
+    %{preferred_locales: preferred_locales} = context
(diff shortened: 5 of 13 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.2] - Unreleased
+### Fixed
+- Correctly handle queries and mutations on translated descriptions.
```
Pavinati, line 9: `Correctly` doesn't explain how it changed.
I'd rather write something like: "Return the text in the default tenant language when the selected language is not available or no translation was inserted"

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.5.2] - Unreleased
```
Pavinati, line 7: Should this be under the 0.6.0 tag?
Regardless, we usually make a dedicated commit for the new release preparation (whether it is 0.5.2 or 0.6.0) and then add the changelog row directly in the commit that introduces it.

backend/lib/edgehog/devices/system_model_description.ex
```diff
@@ -44,8 +44,8 @@ defmodule Edgehog.Devices.SystemModelDescription do
     |> unique_constraint([:locale, :system_model_id, :tenant_id])
   end
 
-  def localized(locale) do
+  def localized(locales) when is_list(locales) do
```
Pavinati, line 47: changing this function forces you to edit all the existing code that uses a single locale.
I would add a new function with a different name or use the same name but differentiate using guards

backend/test/edgehog_web/schema/query/system_models_test.exs
```diff
@@ -137,7 +137,7 @@ defmodule EdgehogWeb.Schema.Query.SystemModelsTest do
       assert system_model["description"]["text"] == "Un modello di sistema"
     end
 
-    test "returns empty description for non existing locale", %{
```
Pavinati, line 140: I would still keep a test for nonexisting descriptions. In that instance we should still return null

backend/lib/edgehog_web/resolvers/devices.ex
```diff
@@ -127,12 +121,12 @@ defmodule EdgehogWeb.Resolvers.Devices do
...
          system_model =
            system_model
-           |> Devices.preload_localized_descriptions_for_system_model(default_locale),
+           |> localize_system_model_description(context),
          {:ok, %SystemModel{} = system_model} <-
            Devices.update_system_model(system_model, attrs) do
       system_model =
         system_model
-        |> Devices.preload_localized_descriptions_for_system_model(default_locale)
+        |> localize_system_model_description(context)
```
Pavinati, line 129: same as above

backend/lib/edgehog_web/context.ex
```diff
@@ -55,4 +55,10 @@ defmodule EdgehogWeb.Context do
         Map.put(context, :locale, locale)
     end
   end
+
+  def list_preferred_locales(%{current_tenant: tenant} = context) do
+    locale = Map.get(context, :locale)
+    default_locale = Map.get(tenant, :default_locale)
+    Enum.reject([locale, default_locale], &is_nil/1)
+  end
```
rbino, line 63: This is a `Plug`, and as such it usually doesn't expose additional functions. This should probably be moved in the resolver.
Moreover on the implementation: the `:default_locale` is non-null at the database level (and it also has a default), so you can be sure it's present. So instead of building the list and rejecting you could just
```elixir
%{default_locale: default_locale} = tenant
case Map.fetch(context, :locale) do
  {:ok, locale} ->
    [locale, default_locale]
  :error ->
    [default_locale]
end
```

backend/lib/edgehog_web/resolvers/devices.ex
```diff
@@ -153,19 +147,29 @@ defmodule EdgehogWeb.Resolvers.Devices do
...
   defp wrap_description(attrs), do: attrs
 
-  def extract_localized_description(%SystemModel{descriptions: descriptions}, %{}, _context) do
-    # We should always either 0 or 1 description here since the upper layer should take care
-    # of only preloading the localized description.
-    case descriptions do
-      [description] -> {:ok, description}
-      _ -> {:ok, nil}
-    end
+  def extract_localized_description(%SystemModel{descriptions: nil}, _args, _resolution) do
```
rbino, line 159: `descriptions` would probably be an empty list instead of `nil` since it's populated with `Repo.all`

rbino (changes requested):

backend/lib/edgehog/devices/system_model_description.ex
davidebriani: I'd avoid having the same function with different signatures. Also, it makes sense to use the same function throughout the code and in the tests, where previously you had to call it twice since it didn't support preloading two locales at the same time.

frontend/src/forms/UpdateSystemModel.tsx
```diff
@@ -137,6 +137,10 @@ const transformOutputData = (
       name: data.hardwareType,
     },
     partNumbers: data.partNumbers.map((pn) => pn.value),
+    description: {
+      locale,
+      text: data.description,
+    },
```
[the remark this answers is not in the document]
davidebriani: I updated this piece so that a `null` value is passed to the backend when the description is erased, i.e. in the case of `description` == `""`.
Note that only the fields that have changed are sent to the backend in the update mutation by the upper layer, the SystemModel page.

Pavinati (approved):

backend/lib/edgehog_web/resolvers/astarte.ex
```diff
@@ -58,21 +58,21 @@ defmodule EdgehogWeb.Resolvers.Astarte do
     end
   end
 
-  defp preload_system_model_for_device(target, %{locale: locale}) do
-    # Explicit locale, use that one
-    descriptions_query = Devices.localized_system_model_description_query(locale)
+  defp preload_system_model_for_device(target, context) do
+    preferred_locales = list_preferred_locales(context)
+    descriptions_query = Devices.localized_system_model_description_query(preferred_locales)
```
rbino, line 63: Use pipes since you're just passing the intermediate result to the next function
```elixir
descriptions_query =
  context
  |> preferred_locales()
  |> Devices.localized_system_model_description_query()
```

backend/lib/edgehog_web/resolvers/astarte.ex
```diff
@@ -58,21 +58,21 @@ defmodule EdgehogWeb.Resolvers.Astarte do
...
     preload = [descriptions: descriptions_query, hardware_type: [], part_numbers: []]
 
     Astarte.preload_system_model_for_device(target, preload: preload)
   end
 
-  defp preload_system_model_for_device(target, %{current_tenant: tenant}) do
-    # Fallback
+  defp list_preferred_locales(%{current_tenant: tenant} = context) do
```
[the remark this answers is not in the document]
rbino: Actually now that I think about it: you're actually copy-pasting this function in a couple of resolvers, and this made me realize it probably belongs to the `EdgehogWeb.Context` plug instead. You have both the tenant and the explicit locale there, just populate a `:preferred_locales` key in the context there and access it in the resolvers (check if `:locale` is still used at that point, if it's not you can remove it from the context).

Moreover this also unlocks another interesting feature: the `accept-language` header can actually contain a list of accepted languages, but we were [just using the first one](https://github.com/edgehog-device-manager/edgehog/blob/b1193c26f403132dead6964c1c052e5dcae533af/backend/lib/edgehog_web/context.ex#L51). If you populate `:preferred_locales` there you can put all the languages in the `accept-language` first and at the end add the default tenant locale.

backend/lib/edgehog_web/resolvers/devices.ex
```diff
@@ -153,19 +140,37 @@ defmodule EdgehogWeb.Resolvers.Devices do
...
-      [description] -> {:ok, description}
-      _ -> {:ok, nil}
+  def extract_localized_description(%SystemModel{descriptions: descriptions}, _args, %{
+        context: context
+      })
+      when is_list(descriptions) do
+    description =
+      context
+      |> list_preferred_locales()
+      |> Enum.flat_map(fn locale -> Enum.filter(descriptions, &(&1.locale == locale)) end)
```
rbino, line 162: There should be no need to filter for preferred locales here since the DB query should already have taken care of that, what you should do instead is find a way to sort descriptions with the same order as the locales preference and extract the first one

rbino (changes requested):

backend/lib/edgehog_web/resolvers/devices.ex
davidebriani: You're right. However here I'm already ordering by preferred locales since I'm looping on them, and I'm using `Enum.filter` just because it works well with `Enum.flat_map` and lets you discard void results of descriptions for a certain locale.
A different approach would be to directly take the descriptions and sort them, e.g.:
```elixir
description =
      descriptions
      |> Enum.sort_by(&Enum.find_index(preferred_locales, &1.locale))
      |> List.first()
```
I have no preference here, you tell me.

backend/lib/edgehog_web/context.ex
```diff
@@ -29,30 +29,31 @@ defmodule EdgehogWeb.Context do
...
-  defp maybe_put_locale(context, conn) do
-    case Plug.Conn.get_req_header(conn, "accept-language") do
-      # If no header or *, we don't add an explicit locale
-      [] ->
-        context
-
-      ["*" | _] ->
-        context
+  defp get_preferred_locales(conn) do
+    default_locale = get_current_tenant(conn).default_locale
```
rbino, line 43: You're already querying the tenant in `build_context`, query it once and pass it as an argument to `get_preferred_locales` too

backend/lib/edgehog_web/context.ex
```diff
@@ -29,30 +29,31 @@ defmodule EdgehogWeb.Context do
...
+    |> Enum.map(&get_locale/1)
+    |> Enum.reject(&is_nil/1)
+    |> Enum.concat([default_locale])
+  end
 
-        Map.put(context, :locale, locale)
+  defp get_locale(string) when is_binary(string) do
+    case Regex.run(~r/[a-z]{2,3}-[A-Z]{2}/, string) do
+      [locale] -> locale
+      _ -> nil
```
rbino, line 56: This gets run on a _single_ `accept-language` header, but an header can contain multiple languages. If you want to add them all you should probably do
```elixir
    |> Plug.Conn.get_req_header("accept-language")
    |> Enum.flat_map(&get_locale/1)
    |> Enum.concat([default_locale])
```

and

```elixir
defp get_locale(string) when is_binary(string) do
  case Regex.run(~r/[a-z]{2,3}-[A-Z]{2}/, string) do
    matches when is_list(matches) -> matches
    nil -> []
  end
end
```

Your code would currently not return any language (besides the default locale) if there's more than one language in `accept-language`

backend/lib/edgehog_web/resolvers/devices.ex
rbino: Imho the second approach is clearer in its intent. Both approaches are not ideal since they possibly scan the other list entirely for each element of the original list though, but we can optimize in a second pass, maybe leave a TODO at least.

rbino (changes requested):

backend/lib/edgehog_web/context.ex
```diff
@@ -21,38 +21,41 @@
...
+    default_locale = tenant.default_locale
 
-      # If there's one (or more) accept-language, we use the first one
-      [language | _] ->
-        [locale | _] = String.split(language, ",", parts: 2)
+    # If there are some locales in accept-language, list them before the default locale
+    conn
+    |> Plug.Conn.get_req_header("accept-language")
+    |> Enum.flat_map(&get_locales/1)
+    |> Enum.reject(&is_nil/1)
```
rbino, line 51: You replaced `Regex.run` with `Regex.scan`, which returns an empty list with no matches, so this is not needed anymore.

rbino (changes requested):

rbino (approved):

PR #187, opened 2022-03-24
doc: add documentation on OTA Updates, Tags, Attributes and Groups

rbino:
Document the (future) supported features of Edgehog's OTA mechanism, including
Groups, Tag and Attribute functionalities

Signed-off-by: Riccardo Binetti <riccardo.binetti@secomind.com>

PR #205, opened 2022-06-13
Add support for Device Custom Attributes

rbino:
This PR introduces the concept of Attributes already present in the [draft OTA Documentation](https://github.com/edgehog-device-manager/edgehog/pull/187).

The first kind of attributes implemented are Custom Attributes, which are completely user-defined.

This PR introduces the infrastructure to support attributes both at the database and at the GraphQL level.

A variant type is also introduced at both ends of the solution to support attributes of all types (currently only scalar types, support for arrays will be tracked by [this issue](https://github.com/edgehog-device-manager/edgehog/issues/206).

### DIFF
--- a/backend/lib/edgehog_web/schema/astarte_types.ex
+++ b/backend/lib/edgehog_web/schema/astarte_types.ex
@@ -92,6 +92,23 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
   end
 
   @desc """
+  An input object for a device attribute.
+  """
+  input_object :device_attribute_input do
+    @desc "The namespace of the device attribute."
+    field :namespace, non_null(:device_attribute_namespace)
+
+    @desc "The key of the device attribute."
+    field :key, non_null(:string)
+
+    @desc "The type of the device attribute."
+    field :type, non_null(:variant_type)
+
+    @desc "The value of the device attribute."
+    field :value, non_null(:variant_value)
+  end
+
+  @desc """
   Describes hardware-related info of a device.
 
   It exposes data read by a device's operating system about the underlying \
@@ -579,6 +596,9 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
 
         @desc "The tags of the device. These replace all the current tags."
         field :tags, list_of(non_null(:string))
+
+        @desc "The custom attributes of the device. These replace all the current custom attributes."
+        field :custom_attributes, list_of(non_null(:device_attribute_input))
       end
 
       output do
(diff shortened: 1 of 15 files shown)


backend/test/edgehog_web/schema/query/device_test.exs
```diff
@@ -171,5 +171,144 @@ defmodule EdgehogWeb.Schema.Query.DeviceTest do
...
+                 "key" => "boolean",
+                 "namespace" => "CUSTOM",
+                 "type" => "BOOLEAN",
+                 "value" => true
+               },
+               %{
+                 "key" => "longinteger",
+                 "namespace" => "CUSTOM",
+                 "type" => "LONGINTEGER",
+                 "value" => 420
```
szakhlypa, line 291: What about longinteger value, shouldn't it be passed as a string or something else? [spec issue](https://github.com/graphql/graphql-spec/issues/73)

rbino: I changed it to a string in tests to cover it, but both passing as a string and as a number is supported (since `cast` will accept both). As said in the issue above JSON has limited support for integers > 32 bits, so users can choose the representation that best works for them.

backend/priv/repo/migrations/20220607055759_create_device_attributes.exs
```diff
@@ -0,0 +1,45 @@
...
+
+      add :namespace, :string, null: false, primary_key: true
+      add :key, :string, null: false, primary_key: true
+      add :typed_value, :map, null: false
+
+      timestamps()
+    end
+
+    create index(:device_attributes, [:tenant_id])
+    create index(:device_attributes, [:device_id, :tenant_id])
```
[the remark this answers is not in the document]
rbino: If you look at the migration the 4 columns you mentioned are used as composite primary key, so the uniqueness of their combination is implicitly enforced

davidebriani (approved):

davidebriani (approved):

backend/lib/edgehog_web/schema/astarte_types.ex
```diff
@@ -553,6 +596,9 @@ defmodule EdgehogWeb.Schema.AstarteTypes do
 
         @desc "The tags of the device. These replace all the current tags."
         field :tags, list_of(non_null(:string))
+
+        @desc "The custom attributes of the device. These replace all the current custom attributes."
+        field :custom_attributes, list_of(non_null(:device_attribute_input))
```
Pavinati, line 601: is it ok to return `null` instead of an empty list when no custom attribute is available?

rbino: Is there a specific reason we should do this? Usually lists with no elements are just sent as empty lists (like tags are sent as an empty lists if there's none)

Pavinati: Oh right, didn't realize it as the context was hidden. My bad

Pavinati (approved):

szakhlypa (approved):

PR #208, opened 2022-06-16
Allow configuring the max upload size

rbino:
Add documentation regarding this in the Kubernetes deployment guide. Also
document the Ingress part of the upload limit.

Signed-off-by: Riccardo Binetti <riccardo.binetti@secomind.com>

### DIFF
--- a/backend/lib/edgehog_web/endpoint.ex
+++ b/backend/lib/edgehog_web/endpoint.ex
@@ -54,7 +54,7 @@ defmodule EdgehogWeb.Endpoint do
   plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
 
   plug Plug.Parsers,
-    parsers: [:urlencoded, :multipart, :json],
+    parsers: [:urlencoded, Edgehog.Multipart, :json],
     pass: ["*/*"],
     json_decoder: Phoenix.json_library()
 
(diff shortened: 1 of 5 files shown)


backend/lib/edgehog_web/endpoint.ex
```diff
@@ -54,7 +54,7 @@ defmodule EdgehogWeb.Endpoint do
   plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
 
   plug Plug.Parsers,
-    parsers: [:urlencoded, :multipart, :json],
+    parsers: [:urlencoded, {:multipart, length: 500_000_000}, :json],
```
bettio, line 57: This should be a configuration parameter, and we should start I think with 1GB/2GB or maybe 4GB.

bettio (changes requested):

backend/lib/edgehog_web/endpoint.ex
rbino: I'm not sure if this is actually configurable at runtime - Plug initialization happens (or actually can happen) at compile time

rbino: Ok the library provides the support for [dynamic configuration](https://hexdocs.pm/plug/Plug.Parsers.MULTIPART.html#module-dynamic-configuration), so that's feasible.

bettio (approved):

PR #211, opened 2022-06-23
Add device selectors

rbino:
Allow selecting a group of devices with a DSL


backend/lib/edgehog/devices/selector/parser.ex
```diff
@@ -0,0 +1,306 @@
...
+
+  attribute_filter =
+    optional(blankspace)
+    |> concat(attribute)
+    |> optional(blankspace)
+    |> concat(attribute_operator)
+    |> optional(blankspace)
+    |> concat(attribute_value)
+    |> post_traverse(:finalize_attribute_filter)
+    |> label("attribute filter")
```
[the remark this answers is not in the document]
rbino: I thought it was needed to write stuff like `(    "foo" in tags)` but it's actually useless because `selector` (for bare expressions) or `factor` (for parenthesized expressions) are already handling the optional leading blankspace

backend/.formatter.exs
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: CC0-1.0
 
 [
-  import_deps: [:ecto, :phoenix, :absinthe, :skogsra],
+  import_deps: [:ecto, :phoenix, :absinthe, :skogsra, :nimble_parsec],
```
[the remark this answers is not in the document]
rbino: Makes sense

doc/pages/user/core_concepts.md
```diff
@@ -107,22 +107,45 @@ A Selector allows selecting a subset of devices based on their tags and attribut
 can be evaluated for a Device and return `true` if the device matches the Selector and `false`
 otherwise.
 
-Each Selector can be made of one or more expressions, combined using `and` and `or`.
+Each Selector can be made of one or more filters, combined using `and` and `or` and (possibly)
+parenthesized. When no parenthesis are present, `and` has an higher priority than `or` in
```
fruchtkeks, line 111: [...] has **a** higher priority [...]

doc/pages/user/core_concepts.md
```diff
@@ -107,22 +107,45 @@ A Selector allows selecting a subset of devices based on their tags and attribut
...
+##### Attribute filter
+
+Created with the syntax `attributes["<namespace>:<key>"] <operator> <value>`, it returns `true` if
+the value of the chosen attribute satisfies the expression.
+
+The supported operators are:
+- `==` and `!=` for all value types
+- `>`, `>=`, `<`, `<=` for numeric or `datetime` values
+
+`<value>` can be a boolean (`true` or `false`), a string, a number of one of the values supported
```
fruchtkeks, line 131: [...] a string **or** a number of one [...]

doc/pages/user/core_concepts.md
```diff
@@ -107,22 +107,45 @@ A Selector allows selecting a subset of devices based on their tags and attribut
...
+expressions.
 
-These are the supported expressions:
+#### Supported filters
+
+##### Tag filter
+
+Created with the syntax `"<value>" in tags`, it returns `true` if `value` is included in the Device
+tags. It's also possible to use a negative filter with `"value" not in tags`, in this case the
+filter will match all Devices which _don't_ have the tag.
```
fruchtkeks, line 120: The spelling of "device" changes throughout the document between lower and upper case. Maybe, settling on one form would be good?

fruchtkeks (commented):
I think the breakdown of the new feature is quite good.
What would be great is an example that brings together the different areas (tags, attributes, etc.) in a real-life manner (maybe even with some visualizations).

rbino:
Thanks for the review @fruchtkeks, I fixed your comments and pushed again. I agree that an end-to-end example would be valuable, I'm not sure it's going to be introduced in this PR or in a future documentation enrichment phase, but it's definitely on the radar.

backend/lib/edgehog/devices/selector/ast/attribute_filter.ex
```diff
@@ -0,0 +1,225 @@
...
+      namespace: namespace,
+      key: key,
+      operator: operator,
+      type: type,
+      value: value
+    } = attribute_filter
+
+    base_query =
+      from a in Attribute,
+        where: a.namespace == ^namespace and a.key == ^key,
```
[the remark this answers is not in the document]
Pavinati: I agree, but we can also leave it as it is and just add an issue to the repo. In the future, we can measure the impact of this and judge if an additional index solves it or if it's just better to include the tenant id in the query at some level

Pavinati (approved):

Pavinati (approved):

backend/test/edgehog/devices/selector/ast/attribute_filter_test.exs
```diff
@@ -0,0 +1,724 @@
...
+          where: ^dynamic
+
+      result =
+        Repo.all(query)
+        # Preload the same associations preloaded in the Astarte context
+        |> Repo.preload([:tags, :custom_attributes])
+
+      assert is_list(result)
+      assert device_41 in result
+      assert device_42 in result
```
davidebriani, line 443: Really nitpicking here. In all the tests you're testing that some devices are present in a query's result, but not that those devices are *the only* ones present.
For instance, in this test you might add:
```elixir
refute device_43 in result
```

rbino: Totally legit observation, I changed to this style instead of explicit matching on the resulting list to avoid depending on the order of the results (which is not explicitly defined in our case) but as you're saying the tests as they are now would pass also for a superset of the expected list

rbino: Actually now that I think about, instead of doing `refute device_43 in result` I think I'll just assert the list length, which is a stronger guarantee (because otherwise a `device_44` out of nowhere could be in the list)

szakhlypa (approved):

davidebriani (approved):

bettio (approved):

PR #213, opened 2022-07-08
Contexts refactoring

rbino:
- Add `Devices.Device` schema

- Separate `%Astarte.Device{}` from `%Devices.Device{}`

These two have different responsibilities: the former is responsible of interacting with Astarte, creating the device in the database and updating it with triggers, the latter is responsible of everything Edgehog builds upon the Astarte device. The two map to the same database table so they share data through it.

- Make the Astarte context accept an AppEngine client when doing its queries instead of a device

This makes the Astarte context independent, with the Devices context calling into it without circular dependencies

- Extract Capabilities to a separate context

Capabilities are not an Astarte concept, they're an Edgehog concept built upon Astarte. The new structure reflects that, without polluting the Astarte context

- Move GraphQL modules around to better map to contexts

Create Geolocation and Capabilities types and resolvers

- Extract Labeling and Selector context

Instead of cramming everything in the Devices context, extract the Selector to its own top level context. Contextually, also add a labeling context that will handle tags and attributes.

- Fix all tests to make them work with the new structure


backend/test/edgehog/capabilities_test.exs
```diff
@@ -0,0 +1,137 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2021 SECO Mind Srl
```
davidebriani, line 4: ~~2021~~ **2022**

backend/lib/edgehog/selector.ex
```diff
@@ -18,13 +18,11 @@
...
 #
 
-defmodule Edgehog.Devices.Selector do
-  alias Edgehog.Astarte.Device
-  alias Edgehog.Devices.Selector.AST.AttributeFilter
-  alias Edgehog.Devices.Selector.AST.BinaryOp
-  alias Edgehog.Devices.Selector.AST.TagFilter
-  alias Edgehog.Devices.Selector.Parser
-  alias Edgehog.Devices.Selector.Parser.Error
+defmodule Edgehog.Selector do
```
davidebriani, line 21: I reckon that contexts are usually named with the plural form. Should we adopt `Edgehog.Selectors` here as well?

davidebriani (changes requested):

backend/lib/edgehog/selector.ex
rbino: It depends on the context (pun not intended), for examples it's `Geolocation` not `Geolocations`. In this case the `Selector` is a concept rather than a group of resources (like `Devices` or `Tenants`) so I left it singular. I don't have a strong opinion about changing it to plural tbh, but for me in this context it "sounds better" as singular.

davidebriani: Makes sense, let's keep it in the singular form.

szakhlypa (approved):

bettio (approved):

PR #223, opened 2022-10-20
Use new Dockerfile template, with images from hex.pm

bettio:
This template is based on the Dockerfile that nowadays is generated with mix phx.gen.release, and make use of tagged images with well known versions, rather a rolling tag.
Elixir and OTP version are explicitly set, and should be manually updated.

Signed-off-by: Davide Bettio <davide.bettio@secomind.com>


backend/Dockerfile
```diff
@@ -1,66 +1,82 @@
...
+# Only copy the final release from the build stage
+COPY --from=builder --chown=nobody:root /app/_build/prod/rel/edgehog ./
 
-COPY --from=builder /app/_build/$BUILD_ENV/rel/edgehog .
-COPY entrypoint.sh .
+USER nobody
 
-ENTRYPOINT ["/bin/bash", "entrypoint.sh"]
-CMD ["start"]
+CMD ["/app/bin/server"]
```
davidebriani, line 82: I guess this command is doomed to fail because the `server` executable is not present.
The Phoenix generator should also provide the additional files that should be included, such as `rel/overlays/bin/server`.

davidebriani (changes requested):

rbino (approved):

szakhlypa (approved):

PR #238, opened 2023-01-13
[frontend] update SystemModel pages for new GraphQL schema

szakhlypa:
Update frontend to respect type change of SystemModel description field

### DIFF
--- a/frontend/src/pages/SystemModel.tsx
+++ b/frontend/src/pages/SystemModel.tsx
@@ -1,7 +1,7 @@
 /*
   This file is part of Edgehog.
 
-  Copyright 2021,2022 SECO Mind Srl
+  Copyright 2021-2023 SECO Mind Srl
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


frontend/src/pages/SystemModel.tsx
```diff
@@ -209,12 +203,9 @@ const SystemModelContent = ({
...
               ...systemModel!,
-              ..._.pick(systemModelChanges, [
-                "name",
-                "handle",
-                "partNumbers",
-                "description",
-              ]),
+              ..._.pick(systemModelChanges, ["name", "handle", "partNumbers"]),
+              description:
+                systemModelChanges.description?.text || systemModel.description,
```
davidebriani, line 208: It looks like this won't let you see the updated description when it's an empty text `""`.
You could use the following instead:
```
description: systemModelChanges.description ? systemModelChanges.description.text : systemModel.description
```

davidebriani (changes requested):

frontend/src/pages/SystemModel.tsx
Pavinati: or we could start using the `??` operator
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

davidebriani (approved):

davidebriani (approved):

Pavinati (approved):

rbino (approved):

PR #240, opened 2023-01-18
Add Base Images

rbino:
Allow uploading a Base Image to a Base Image Collection and doing all CRUD operations on it

### DIFF
--- a/backend/lib/edgehog/base_images.ex
+++ b/backend/lib/edgehog/base_images.ex
@@ -27,6 +27,7 @@ defmodule Edgehog.BaseImages do
   alias Edgehog.Repo
 
   alias Edgehog.Devices
+  alias Edgehog.BaseImages.BaseImage
   alias Edgehog.BaseImages.BaseImageCollection
 
   @doc """
@@ -149,7 +150,16 @@ defmodule Edgehog.BaseImages do
     BaseImageCollection.changeset(base_image_collection, attrs)
   end
 
-  alias Edgehog.BaseImages.BaseImage
+  @doc """
+  Preloads the default associations for a Base Image (or a list of base images)
+  """
+  def preload_defaults_for_base_image(image_or_images) do
+    Repo.preload(image_or_images,
+      base_image_collection: [
+        system_model: [:hardware_type, :part_numbers]
+      ]
+    )
+  end
 
   @doc """
   Returns the list of base_images.
@@ -162,40 +172,50 @@ defmodule Edgehog.BaseImages do
   """
   def list_base_images do
     Repo.all(BaseImage)
+    |> preload_defaults_for_base_image()
   end
 
   @doc """
-  Gets a single base_image.
+  Fetches a single base_image.
 
-  Raises `Ecto.NoResultsError` if the Base image does not exist.
+  Returns `{:error, :not_found}` if the Base image does not exist.
 
   ## Examples
 
-      iex> get_base_image!(123)
-      %BaseImage{}
+      iex> fetch_base_image(123)
+      {:ok, %BaseImage{}}
 
-      iex> get_base_image!(456)
-      ** (Ecto.NoResultsError)
+      iex> fetch_base_image(456)
+      {:error, :not_found}
 
   """
-  def get_base_image!(id), do: Repo.get!(BaseImage, id)
+  def fetch_base_image(id) do
+    case Repo.get(BaseImage, id) do
+      nil -> {:error, :not_found}
+      %BaseImage{} = base_image -> {:ok, preload_defaults_for_base_image(base_image)}
+    end
+  end
 
   @doc """
   Creates a base_image.
 
   ## Examples
 
-      iex> create_base_image(%{field: value})
+      iex> create_base_image(%BaseImageCollection{}, %{field: value})
       {:ok, %BaseImage{}}
 
-      iex> create_base_image(%{field: bad_value})
+      iex> create_base_image(%BaseImageCollection{}, %{field: bad_value})
       {:error, %Ecto.Changeset{}}
 
   """
-  def create_base_image(attrs \\ %{}) do
-    %BaseImage{}
-    |> BaseImage.changeset(attrs)
-    |> Repo.insert()
+  def create_base_image(%BaseImageCollection{} = base_image_collection, attrs \\ %{}) do
+    changeset =
+      %BaseImage{base_image_collection_id: base_image_collection.id}
+      |> BaseImage.create_changeset(attrs)
+
+    with {:ok, base_image} <- Repo.insert(changeset) do
+      {:ok, preload_defaults_for_base_image(base_image)}
+    end
   end
 
   @doc """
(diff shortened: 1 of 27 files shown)


backend/test/support/mocks/base_images/storage.ex
```diff
@@ -0,0 +1,49 @@
...
+    %BaseImage{
+      tenant_id: tenant_id,
+      base_image_collection_id: base_image_collection_id,
+      version: version
+    } = scope
+
+    ext = Path.extname(upload.filename)
+
+    file_name =
+      "uploads/tenants/#{tenant_id}/base_image_collections/#{base_image_collection_id}/base_images/#{version}.#{ext}"
```
davidebriani, line 39: Just nitpicking: the actual uploader does not keep the version in the filename, so you could skip that here too and still have uniformity.

backend/lib/edgehog/base_images.ex
```diff
@@ -43,6 +43,7 @@ defmodule Edgehog.BaseImages do
   """
   def preload_defaults_for_base_image_collection(collection_or_collections) do
     Repo.preload(collection_or_collections,
+      base_images: [],
```
davidebriani, line 46: Not sure if loading the entire base image list by default is ideal. It seems to me that you don't use these preloads, since you can always call `BaseImages.list_base_images_for_collection/1` when you need the list.
Is the preload  required when defining the `base_images` assoc in the schema as `translatable_has_many`?

davidebriani (changes requested):

backend/lib/edgehog/base_images.ex
```diff
@@ -148,4 +158,168 @@ defmodule Edgehog.BaseImages do
...
+  end
+
+  @doc """
+  Fetches a single base_image.
+
+  Returns `{:error, :not_found}` if the Base image does not exist.
+
+  ## Examples
+
+      iex> fetch_base_image!(123)
```
szakhlypa, line 209: Remove exclamation marks in examples (`fetch_base_image!` -> `fetch_base_image`)

szakhlypa (changes requested):

backend/lib/edgehog/base_images.ex
rbino: The problem with calling `list_base_images_for_collection` is that it generates an N+1 query problem if, e.g., you're asking for the _list of base image collections_ and then for the _list of base images_ for each of them (which is something you could do using the GraphQL API, since I'm going to add a `base_images` field to the `base_image_collection` object).
Right now I'm always leaning towards the "Preload all the possible associations" approach, which is the fastest way to proceed. Once we have a good amount of code, I want to try to tackle this with a more structured and fine-grained approach (e.g. Dataloader), but I don't want to slow the development in this phase to experiment with this

backend/test/edgehog/base_images_test.exs
```diff
@@ -120,81 +119,159 @@ defmodule Edgehog.BaseImagesTest do
...
+      assert {:ok, %BaseImage{}} =
+               create_base_image(base_image_collection: collection_2, version: "1.3.0")
     end
 
-    test "create_base_image/1 with invalid data returns error changeset" do
-      assert {:error, %Ecto.Changeset{}} = BaseImages.create_base_image(@invalid_attrs)
+    test "create_base_image/1 with invalid version returns error changeset" do
+      assert {:error, %Ecto.Changeset{errors: errors}} = create_base_image(version: "foobaz")
+
+      assert errors[:version] != nil
```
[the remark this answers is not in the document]
rbino: TIL

backend/test/edgehog/base_images_test.exs
```diff
@@ -120,81 +119,159 @@ defmodule Edgehog.BaseImagesTest do
...
+      assert {:error, %Ecto.Changeset{errors: errors}} =
+               update_base_image(starting_version_requirement: "invalid")
 
-      assert base_image == BaseImages.get_base_image!(base_image.id)
+      assert errors[:starting_version_requirement] != nil
     end
 
     test "delete_base_image/1 deletes the base_image" do
-      base_image = base_image_fixture()
+      base_image = create_base_image!()
```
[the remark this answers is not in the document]
rbino: While I linked the relevant resources, I agree with you that `create_base_image!` and `base_image_fixture` are doing basically the same thing, so I'll remove `create_base_image!` and use `base_image_fixture` (adapting it accordingly) when I'm not _testing_ the creation of the image (but rather I just want an image to use it in a test).
I'll do the same for `create_base_image_collection!`.

backend/lib/edgehog/base_images.ex
davidebriani: Totally legit. Thanks for the explanation!

backend/test/support/mocks/base_images/storage.ex
rbino: It does though: `storage_dir` creates the prefix, and then `filename` returns the version (the extension gets automatically appended from the original filename)

backend/test/edgehog/base_images_test.exs
```diff
@@ -120,81 +119,159 @@ defmodule Edgehog.BaseImagesTest do
...
+                 release_display_name: %{"en-US" => "Happy Hyena"},
+                 starting_version_requirement: "~> 1.0",
+                 version: "1.4.0"
+               )
+
+      assert base_image.description == %{"en-US" => "A feature packed release"}
+      assert base_image.release_display_name == %{"en-US" => "Happy Hyena"}
+      assert base_image.starting_version_requirement == "~> 1.0"
+      assert base_image.version == "1.4.0"
+    end
```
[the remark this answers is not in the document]
rbino: As discussed out of band, this is an attempt to make tests clearer to concentrate only on the details that matter and reduce unrelated noise

backend/lib/edgehog_web/schema/base_images_types.ex
```diff
@@ -39,6 +39,50 @@ defmodule EdgehogWeb.Schema.BaseImagesTypes do
...
+    The language of the description can be controlled passing an \
+    Accept-Language header in the request. If no such header is present, the \
+    default tenant language is returned.
+    """
+    field :release_display_name, :string do
+      resolve &Resolvers.BaseImages.extract_localized_release_display_name/3
+    end
+
+    @desc "The Base Image Collection the Base Image belongs to"
+    field :base_image_collection, :base_image_collection
```
szakhlypa, line 85: `, non_null(:base_image_collection)`

szakhlypa (changes requested):

backend/lib/edgehog_web/schema/base_images_types.ex
```diff
@@ -254,5 +254,22 @@ defmodule EdgehogWeb.Schema.BaseImagesTypes do
 
       resolve &Resolvers.BaseImages.update_base_image/2
     end
+
+    @desc "Deletes a base image."
+    payload field :delete_base_image do
+      input do
+        @desc "The ID of the base image to be updated."
```
davidebriani, line 261: The ID of the base image to be ~~updated~~ **deleted**.

davidebriani (changes requested):

szakhlypa (approved):

PR #245, opened 2023-02-27
[frontend] Add Update Channels

szakhlypa:
Adds forms and logic to create, update and delete `Update Channels`.

#### Create `Update Channel`:

https://user-images.githubusercontent.com/26611935/221592922-7ebf79b3-1e7c-4f23-aef7-df244ad3cbad.mp4

#### Create `Update Channel` (disabled target groups are in use for other `Update Channels`):

https://user-images.githubusercontent.com/26611935/221593266-768187c4-93e2-4169-ad8d-cd7d36c365da.mp4

#### Update and delete `Update Channel`:

https://user-images.githubusercontent.com/26611935/221593299-779eb3d6-b72c-406d-8cbd-4118b93fabf0.mp4

### DIFF
--- a/frontend/src/components/MultiSelect.tsx
+++ b/frontend/src/components/MultiSelect.tsx
@@ -1,7 +1,7 @@
 /*
   This file is part of Edgehog.
 
-  Copyright 2022 SECO Mind Srl
+  Copyright 2022-2023 SECO Mind Srl
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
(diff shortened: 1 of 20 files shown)


frontend/src/components/MultiSelect.tsx
```diff
@@ -81,38 +67,66 @@ const customComponents = {
...
+        {...restProps}
+        isMulti
+        isDisabled={disabled}
+        isLoading={loading}
+        components={customComponents}
+        className={`multi-select ${invalid ? "is-invalid" : ""}`}
+        classNamePrefix="multi-select"
+      />
+    );
+  }
```
davidebriani, line 120: You're returning different components but passing the same props to both. How about:
```tsx
const SelectComponent = creatable ? CreatableSelect : Select;

return (
      <SelectComponent
        {...restProps}
        isMulti
        isDisabled={disabled}
        isLoading={loading}
        components={customComponents}
        className={`multi-select ${invalid ? "is-invalid" : ""}`}
        classNamePrefix="multi-select"
      />
    );
```

frontend/src/forms/UpdateUpdateChannel.tsx
```diff
@@ -0,0 +1,293 @@
...
+    if (
+      "message" in errors &&
+      typeof (errors as Record<"message", unknown>).message === "string"
+    ) {
+      const message = (errors as Record<"message", string>).message;
+      return <FormattedMessage id={message} />;
+    }
+  }
+  return null;
+};
```
davidebriani, line 66: Not sure if this component with all this code is really necessary. I wonder if following the pattern already used for the other error messages would keep the code leaner and clearer:
```
errors.targetGroups?.message && <FormattedMessage id={errors.targetGroups.message} />
```

frontend/src/forms/UpdateUpdateChannel.tsx
```diff
@@ -0,0 +1,293 @@
...
+    readonly name: string;
+  } | null;
+};
+
+const targetGroupWithUpdateChannelLabel = defineMessage({
+  id: "forms.UpdateUpdateChannel.targetGroupWithChannelLabel",
+  defaultMessage: "{targetGroupName} (used for {updateChannelName})",
+  description:
+    "Target group label of select option with optional update channel name it used for.",
+});
```
davidebriani, line 82: You're using `targetGroupWithUpdateChannelLabel` just once. You could evaluate whether to directly define the message inline where you're using it, so that you're colocating the values with the message definition and it's clear how they are rendered in the message. It might also make more evident the relation of this case "group used for channel" with the other cases "group not used" and "group used for this channel".

davidebriani (changes requested):

frontend/src/pages/UpdateChannelCreate.tsx
```diff
@@ -0,0 +1,204 @@
...
+      id
+      name
+      updateChannel {
+        name
+      }
+    }
+  }
+`;
+
+const CREATE_BASE_IMAGE_COLLECTION_MUTATION = graphql`
```
davidebriani, line 54: I guess `CREATE_BASE_IMAGE_COLLECTION_MUTATION` is a left-over from a copy&paste. Or is the name intentional?

davidebriani (changes requested):

Pavinati (approved):

davidebriani (changes requested):

frontend/src/forms/UpdateUpdateChannel.tsx
szakhlypa: RHF does not return valid type for `targetGroups` validation errors so we need to type guard it.
I've added comment with more detailed explanation and `TODO` note.

frontend/src/components/Sidebar.tsx
```diff
@@ -170,6 +170,20 @@ const Sidebar = () => (
...
+            defaultMessage="Update Channels"
+          />
+        }
+        route={Route.updateChannels}
+        activeRoutes={[
+          Route.updateChannels,
+          Route.updateChannelsEdit,
+          Route.updateChannelsNew,
+        ]}
+      />
```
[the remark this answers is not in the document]
szakhlypa: We will review Sidebar menu groups after first implementation of fleet operations.

PR #272, opened 2023-04-05
Update Campaigns support (part 1)

rbino:
This PR will enable the creation of Update Campaigns via GraphQL, but it will not yet execute them.

The Update Campaign execution will be implemented in a subsequent PR, to allow frontend development to proceed in parallel.


backend/lib/edgehog/update_campaigns/push_rollout.ex
```diff
@@ -0,0 +1,57 @@
...
+      :max_in_progress_updates,
+      :ota_request_retries,
+      :ota_request_timeout_seconds
+    ])
+    |> validate_number(:max_errors_percentage,
+      greater_than_or_equal_to: 0,
+      less_than_or_equal_to: 100
+    )
+    |> validate_number(:max_in_progress_updates, greater_than: 0)
+    |> validate_number(:ota_request_retries, greater_than: 0)
```
szakhlypa, line 54: `greater_than_or_equal_to`

backend/lib/edgehog_web/schema/update_campaigns_types.ex
```diff
@@ -41,6 +42,167 @@ defmodule EdgehogWeb.Schema.UpdateCampaignsTypes do
...
+  end
+
+  @desc """
+  The outcome of an Update Campaign
+  """
+  enum :update_campaign_outcome do
+    @desc "The Update Campaign has finished succesfully"
+    value :success
+    @desc "The Update Campaign has finished unsuccesfully"
+    value :error
```
szakhlypa, line 177: `failure`

backend/lib/edgehog_web/schema/update_campaigns_types.ex
```diff
@@ -139,5 +301,35 @@ defmodule EdgehogWeb.Schema.UpdateCampaignsTypes do
       middleware Absinthe.Relay.Node.ParseIDs, update_channel_id: :update_channel
       resolve &Resolvers.UpdateCampaigns.delete_update_channel/2
     end
+
+    @desc "Creates a new update campaign."
+    payload field :create_update_campaign do
+      input do
+        @desc """
```
szakhlypa, line 308: Should we add `name` for update campaign?

backend/lib/edgehog/update_campaigns.ex
```diff
@@ -385,8 +386,59 @@ defmodule Edgehog.UpdateCampaigns do
...
+    |> Repo.transaction()
+    |> case do
+      {:ok, %{finished_update_campaign: update_campaign}} ->
+        {:ok, preload_defaults_for_update_campaign(update_campaign)}
+
+      {:ok, %{update_campaign: update_campaign}} ->
+        {:ok, preload_defaults_for_update_campaign(update_campaign)}
+
+      {:error, _failed_operation, failed_value, _changes_so_far} ->
+        {:error, failed_value}
```
davidebriani, line 419: It's not immediately clear that calling `insert_campaign_targets_multi` can also result in a `:finished_update_campaign` step without having a `:insert_campaign_targets` step.
You could consider an approach such as:

```elixir
Multi.new()
|> Multi.insert(:update_campaign, changeset)
|> Multi.merge(fn %{update_campaign: update_campaign} ->
  updatable_devices = list_updatable_devices(update_channel, base_image)
  if Enum.empty?(updatable_devices) do
    finish_update_campaign_multi(update_campaign)
  else
    insert_campaign_targets_multi(update_campaign, updatable_devices)
  end
end)
|> Repo.transaction()
|> case do
  {:ok, %{finish_update_campaign: update_campaign}} ->
    {:ok, preload_defaults_for_update_campaign(update_campaign)}

  {:ok, %{update_campaign: update_campaign}} ->
    {:ok, preload_defaults_for_update_campaign(update_campaign)}

  {:error, _failed_operation, failed_value, _changes_so_far} ->
    {:error, failed_value}
end
```

Where the `*_multi` functions follow the convention of returning steps named after the function but without the `_multi` suffix. For instance:
- `finish_update_campaign_multi` has a `:finish_update_campaign`  step
- `insert_campaign_targets_multi` has a `:insert_campaign_targets` step (instead of the `:targets` step name it currently has)

davidebriani (changes requested):

backend/test/edgehog/update_campaigns_test.exs
```diff
@@ -343,6 +348,70 @@ defmodule Edgehog.UpdateCampaignsTest do
...
+      bar_device =
+        device_fixture_compatible_with(base_image)
+        |> add_tags(["bar"])
+
+      updatable_devices = UpdateCampaigns.list_updatable_devices(update_channel, base_image)
+      assert length(updatable_devices) == 2
+      assert foo_device in updatable_devices
+      assert bar_device in updatable_devices
+    end
+  end
```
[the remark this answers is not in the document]
rbino: Good idea, done

backend/lib/edgehog_web/schema/update_campaigns_types.ex
rbino: Done

backend/lib/edgehog/update_campaigns/update_campaign.ex
```diff
@@ -0,0 +1,56 @@
...
+    has_many :update_targets, Target
+
+    timestamps()
+  end
+
+  @doc false
+  def changeset(update_campaign, attrs) do
+    update_campaign
+    |> cast(attrs, [:name])
+    |> unique_constraint(:name)
```
szakhlypa, line 53: `:name` is not nullable, let's require it with
`|> validate_required([:name])`

szakhlypa (changes requested):

backend/lib/edgehog/update_campaigns.ex
rbino: I refactored the whole thing, branching early on `updatable_devices`. This way `Multi` is used only if there's the need and the code is more readable imho.

backend/lib/edgehog/os_management.ex
```diff
@@ -140,6 +141,44 @@ defmodule Edgehog.OSManagement do
...
+    })
+    |> Multi.run(:send_ota_request, fn _repo, _changes ->
+      device_id = device.device_id
+
+      with {:ok, client} <- Devices.appengine_client_from_device(device),
+           :ok <- Astarte.send_ota_request(client, device_id, ota_operation_id, base_image.url) do
+        {:ok, nil}
+      end
+    end)
+    |> Repo.transaction()
```
Pavinati, line 172: In light of our recent discussion over `Ecto.Multi`, what about dropping it entirely if we're not using or combining the intermediate steps?
```elixir
    Repo.transaction(fn ->
      device_id = device.device_id
      ota_operation = Repo.insert!(%OTAOperation{
        id: ota_operation_id,
        base_image_url: base_image.url,
        device_id: device_id
      })

      with {:ok, client} <- Devices.appengine_client_from_device(device),
           :ok <- Astarte.send_ota_request(client, device_id, ota_operation_id, base_image.url) do
        {:ok, nil}
      end

      ota_operation
    end)
```
By using a dedicated function for the basic `Repo.insert` part, this could also be handled in a single `with` statement

rbino: The code above would not work since it ignores failures in sending the OTA Request. The point of the `Multi` here is reverting the insertion of the OTAOperation if the push to the device fails.
We could also switch to `Repo.transact` instead, but given all the rest of the code is still using `Multi` I'd rather keep that for a separate refactoring

Pavinati: I do agree, but i'm afraid we'll ad more and more code and never have the opportunity to refactor it.
I'd suggest starting now and leave the remaining code for a different PR

backend/priv/repo/migrations/20230308114354_create_update_campaigns.exs
```diff
@@ -22,21 +22,40 @@ defmodule Edgehog.Repo.Migrations.CreateUpdateCampaigns do
...
+          ),
+          null: false
+
+      add :base_image_id,
+          references(:base_images,
+            with: [tenant_id: :tenant_id],
+            match: :full,
+            on_delete: :nilify_all
+          ),
+          null: false
```
szakhlypa, line 45: Same as above, I expect that `null: false` will not work with `nilify_all`

backend/lib/edgehog/update_campaigns.ex
```diff
@@ -280,4 +315,191 @@ defmodule Edgehog.UpdateCampaigns do
...
+      [%UpdateCampaign{}, ...]
+
+  """
+  def list_update_campaigns do
+    Repo.all(UpdateCampaign)
+    |> preload_defaults_for_update_campaign()
+  end
+
+  @doc """
+  Fetches a single update_campaign.
```
szakhlypa, line 357:
```suggestion
  Fetches a single update campaign.
```

backend/test/edgehog/update_campaigns_test.exs
```diff
@@ -281,6 +287,226 @@ defmodule Edgehog.UpdateCampaignsTest do
...
+      assert UpdateCampaigns.fetch_update_campaign(update_campaign.id) == {:ok, update_campaign}
+    end
+
+    test "returns {:error, :not_found} for non-existing id" do
+      assert UpdateCampaigns.fetch_update_campaign(1_234_567) == {:error, :not_found}
+    end
+  end
+
+  describe "create_update_campaign/3" do
+    test "with valid data creates a update_campaign" do
```
szakhlypa, line 307:
```suggestion
    test "with valid data creates an update_campaign" do
```

backend/lib/edgehog/update_campaigns.ex
```diff
@@ -280,4 +315,191 @@ defmodule Edgehog.UpdateCampaigns do
...
+      {:ok, %{update_campaign: update_campaign}} ->
+        {:ok, preload_defaults_for_update_campaign(update_campaign)}
+
+      {:error, _failed_operation, failed_value, _changes_so_far} ->
+        {:error, failed_value}
+    end
+  end
+
+  @doc """
+  Preloads the default associations for an Target or a list of Targets
```
szakhlypa, line 468:
```suggestion
  Preloads the default associations for a Target or a list of Targets
```

szakhlypa:
In commit `42928dee90` there are tests with `managed_ota_operation_fixture` usage, that fixture was added in commit `fbc424e5e7`, so commit `42928dee90` should be after the `fbc424e5e7` one.

backend/lib/edgehog/os_management.ex
szakhlypa, line 1: We may also add `preload_defaults_for_ota_operation` (as in another contexts).
```elixir
  @doc """
  Preloads the default associations for an OTAOperation or a list of OTAOperations
  """
  def preload_defaults_for_ota_operation(operation_or_operations) do
    Repo.preload(operation_or_operations, [:device])
  end
```

szakhlypa (changes requested):

backend/lib/edgehog/os_management.ex
rbino: The OTAOperation currently doesn't have many associations and currently I don't see the future need of adding more of them (nor preloading them from the outside), so for now I'd avoid extracting this in a separate function.

szakhlypa (approved):
You may also want to update year in Copyright header for following modified files:
- `backend/lib/edgehog/error.ex`
- `backend/lib/edgehog/os_management.ex`
- `backend/lib/edgehog/os_management/ota_operation.ex`
- `backend/lib/edgehog_web/schema.ex`
- `backend/test/edgehog/os_management_test.exs`
- `backend/test/edgehog/selector/ast/tag_filter_test.exs`
- `backend/test/edgehog/selector_test.exs`

szakhlypa (changes requested):

szakhlypa (approved):

PR #275, opened 2023-05-16
[frontend] Update campaigns

szakhlypa:
Adds forms and logic to create and view Update Campaigns.

Create Update Campaign:

https://github.com/edgehog-device-manager/edgehog/assets/26611935/368b8764-2079-4a24-ae6a-911a2764f4f7

Auto-refresh Update Campaign:

https://github.com/edgehog-device-manager/edgehog/assets/26611935/64133f31-3217-479a-9bb5-4b5b0ba06d64

Create Update Campaign validation:
Screenshot from 2023-05-24 11-56-18

No Base Image Collection:
update-campaign_no-base-image-collection

No Update Channel:
update-campaign_no-update-channel

### DIFF
--- a/frontend/src/forms/UpdateCampaignForm.tsx
+++ b/frontend/src/forms/UpdateCampaignForm.tsx
@@ -1,0 +1,248 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2023 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  SPDX-License-Identifier: Apache-2.0
+*/
+
+import type { ReactNode } from "react";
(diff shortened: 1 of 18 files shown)


frontend/src/forms/UpdateCampaignForm.tsx
```diff
@@ -0,0 +1,248 @@
...
+          <Link
+            route={Route.updateChannelsEdit}
+            params={{ updateChannelId: updateChannel.id }}
+          >
+            {updateChannel.name}
+          </Link>
+        </FormRow>
+      </Col>
+      <Col lg>
+        {maxInProgressUpdates !== undefined && (
```
Pavinati, line 173:
```suggestion
        {maxInProgressUpdates ?? (
```

szakhlypa: `maxInProgressUpdates` can be `undefined` or `number`. We want to render it's section only when `maxInProgressUpdates !== undefined`. Having nullish coalescing operator will render:
- only number (without FormRow/label/ecc when we have maxInProgressUpdates in response)
- FormRow with empty child when we does not have maxInProgressUpdates in response.

We may use conditional rendering with `&&` but with numbers it also will hide section for zeros. We expect that `maxInProgressUpdates` should be more than zero, but `maxErrorsPercentage`, `otaRequestRetries` may be 0 and I prefered to use same syntax for them. Please let me know if it seems ok for you.

Pavinati: I actually meant to write `&&` but you're right, this will consider 0 as invalid

Pavinati (approved):

PR #280, opened 2023-05-22
Add validation for `base_api_url` in Astarte.Cluster changeset

### LINKED ISSUE #250: Validate `base_api_url` in Astarte.Cluster changeset
Currently no validation is performed, and since this is (at the moment) inserted manually by the user, this could lead to invalid URLs (e.g. missing `https://`) which would be discovered only later (e.g. when receiving the first trigger).

noaccOS:
Validate that the given `base_api_url` is a valid URL by cheking that
- The scheme is either `"http"` or `"https"`
- The host is **not** `nil` or `""`, or contains spaces

Closes #250

### DIFF
--- a/backend/lib/edgehog/astarte/cluster.ex
+++ b/backend/lib/edgehog/astarte/cluster.ex
@@ -1,7 +1,7 @@
 #
 # This file is part of Edgehog.
 #
-# Copyright 2021 SECO Mind Srl
+# Copyright 2021-2023 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -38,5 +38,23 @@ defmodule Edgehog.Astarte.Cluster do
     |> cast(attrs, [:name, :base_api_url])
     |> validate_required([:name, :base_api_url])
     |> unique_constraint(:name)
+    |> validate_change(:base_api_url, &validate_url/2)
+  end
+
+  defp validate_url(field, url) do
+    %URI{scheme: scheme, host: maybe_host} = URI.parse(url)
+
+    host = to_string(maybe_host)
+    empty_host? = host == ""
+    space_in_host? = host =~ " "
+
+    valid_host? = not empty_host? and not space_in_host?
+    valid_scheme? = scheme in ["http", "https"]
+
+    if valid_host? and valid_scheme? do
+      []
+    else
+      [{field, "is not a valid URL"}]
+    end
   end
 end
--- a/backend/test/edgehog/astarte_test.exs
+++ b/backend/test/edgehog/astarte_test.exs
@@ -34,6 +34,9 @@ defmodule Edgehog.AstarteTest do
   describe "clusters" do
     alias Edgehog.Astarte.Cluster
 
+    import Edgehog.AstarteFixtures
+
+    @valid_attrs %{base_api_url: "http://some-base-api.url", name: "some name"}
     @invalid_attrs %{base_api_url: nil, name: nil}
 
     test "list_clusters/0 returns all clusters" do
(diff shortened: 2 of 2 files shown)


backend/lib/edgehog/astarte/cluster.ex
```diff
@@ -38,5 +38,16 @@ defmodule Edgehog.Astarte.Cluster do
     |> cast(attrs, [:name, :base_api_url])
     |> validate_required([:name, :base_api_url])
     |> unique_constraint(:name)
+    |> validate_change(:base_api_url, &validate_url/2)
+  end
+
+  defp validate_url(field, url) do
+    with parsed_uri = URI.parse(url),
```
rbino, line 45: `with` chains should never start with a `=` match (or generally contain `=` matches which don't depend on the data inside the `with`), so that can be factored out (this was also caught [by the CI](https://github.com/edgehog-device-manager/edgehog/actions/runs/5046799181/jobs/9054549979?pr=280)).
Moreover, both conditions below are simple boolean, so replacing the `with` with an `if` actually makes sense and removes some noise (using pattern matching reduces it further)
```elixir
%URI{scheme: scheme, host: host} = URI.parse(url)
if scheme in ["http", "https"] and to_string(host) != "" do
  []
else
  ...
end
```

backend/test/edgehog/astarte_test.exs
```diff
@@ -42,23 +42,33 @@ defmodule Edgehog.AstarteTest do
...
+        @invalid_attrs,
+        %{base_api_url: "some url", name: "some name"},
+        %{base_api_url: "http://", name: "some name"},
+        %{base_api_url: "some-url.com", name: "some name"},
+        %{base_api_url: "ftp://some-url.com", name: "some name"}
+      ]
+
+      invalid_attrs_list
+      |> Enum.map(&Astarte.create_cluster/1)
+      |> Enum.each(fn cluster -> assert {:error, %Ecto.Changeset{}} = cluster end)
```
rbino, line 63: While this leads to a single compact test, it has the defect that the first failing test will stop the testcase so you only discover that more than one case is failing when you fix the first one, opposed to discovering all the failing testcases in a single test run.
I'd create two more separate tests:
```
test "create_cluster/1 with invalid URL schema returns error changeset"
test "create_cluster/1 with invalid URL host returns error changeset"
```

rbino (changes requested):
This kind of change is localized enough that is should be contained in a single commit, please squash them together and provide a summary of the changes in the commit message.

backend/test/edgehog/astarte_test.exs
noaccOS: > I'd create two more separate tests:

You mean I should have a total of 3 tests? Isn't it better to just split it in two in that case?

rbino: Legit, you can leave the old test as is and introduce a new test that uses the technique above (attrs_list + `each`) to test some different corner cases of the URL validation.
Just a note: to further underline the fact that `name` is not actually relevant for the failur, build the attr_list something like:
```elixir
base_attrs = %{name: unique_cluster_name()}

invalid_url_attrs_list = 
  [
    %{base_api_url: "..."},
    %{base_api_url: "..."},
  ]
  |> Enum.map(&Enum.into(&1, base_attrs))
```
This makes it more evident to the reader that the name is not involved in the validity of the attributes

noaccOS: I ended up implementing 3 test blocks for create_cluster, as you originally suggested

I didn't consider there'd be cases of invalid data _outside_ the invalid url stuff, my bad

backend/lib/edgehog/astarte/cluster.ex
noaccOS: Since I had to validate for "no spaces in host" anyway I preferred declaring `valid_scheme?` and `valid_host?` to make what I was doing more explicit (e4d78b5b518f0fb5248b3348815998539146bd93)

backend/lib/edgehog/astarte/cluster.ex
```diff
@@ -38,5 +38,21 @@ defmodule Edgehog.Astarte.Cluster do
...
+    host = to_string(maybe_host)
+    empty_host? = host == ""
+    space_in_host? = host =~ " "
+
+    valid_host? = not empty_host? and not space_in_host?
+    valid_scheme? = scheme in ["http", "https"]
+
+    if valid_host? and valid_scheme?,
+      do: [],
+      else: [{field, "is not a valid URL"}]
```
rbino, line 56: Nit: as coding style, we usually use the keyword `do:` only for short function heads which return a value. For everything else (especially if there's an `else`, no pun intended) we use
```elixir
if ... do
  ...
else
  ...
end
```

rbino (changes requested):

noaccOS:
Commits squashed as requested

rbino (approved):

PR #288, opened 2023-06-01
[frontend] Relay v14.1

szakhlypa:
Update Relay to 14.1 version.
Use [Relay Resolvers](https://relay.dev/docs/guides/relay-resolvers/) to define BaseImage client-only field `name`


frontend/src/api/index.ts
```diff
@@ -18,6 +18,7 @@
   SPDX-License-Identifier: Apache-2.0
 */
 
+import "api/relay";
```
davidebriani, line 21: I'm not sure why this import is here. Is it required?

szakhlypa: This way we ensure that `relay-runtime` feature flags are configured before any use of Relay.
We use it to enable [ENABLE_RELAY_RESOLVERS](https://relay.dev/docs/guides/relay-resolvers/#current-limitations).

PR #305, opened 2023-06-22
feat(UpdateCampaign): add API for updates

### LINKED ISSUE #278: Add API to increase `max_in_progress_updates`
Right now we don't have any timeout for all the OTAOperation states besides `Pending`. This is because a device can, potentially, update its state only a long time after it actually changes it (e.g.: a device reboots and in that moment an internet fiber cable gets cut, resulting in it being able to update the state to `Success` only after some time).
This however means that all the slots for `max_in_progress_updates` could be occupied by devices waiting for status updates. To provide an escape hatch if the user wants the campaign to still make progress, we should allow to raise (not lower) the number of `max_in_progress_updates`, so new slots can become available.

noaccOS:
Allows editing of an existing update campaign.
Closes #278

### DIFF
--- a/backend/lib/edgehog/update_campaigns.ex
+++ b/backend/lib/edgehog/update_campaigns.ex
@@ -401,7 +401,7 @@ defmodule Edgehog.UpdateCampaigns do
         update_channel_id: update_channel_id,
         base_image_id: base_image_id
       }
-      |> UpdateCampaign.changeset(attrs)
+      |> UpdateCampaign.create_changeset(attrs)
 
     if updatable_devices == [] do
       create_empty_update_campaign(changeset)
--- a/backend/lib/edgehog/update_campaigns.ex
+++ b/backend/lib/edgehog/update_campaigns.ex
@@ -466,6 +466,26 @@ defmodule Edgehog.UpdateCampaigns do
   end
 
   @doc """
+  Updates an update campaign.
+
+  ## Examples
+
+      iex> update_update_campaign(update_campaign, %{field: new_value})
+      {:ok, %UpdateCamapaign{}}
+
+      iex> update_update_campaign(update_campaign, %{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def update_update_campaign(%UpdateCampaign{} = update_campaign, attrs) do
+    changeset = UpdateCampaign.changeset(update_campaign, attrs)
+
+    with {:ok, update_campaign} <- Repo.update(changeset) do
+      {:ok, preload_defaults_for_update_campaign(update_campaign)}
+    end
+  end
+
+  @doc """
   Preloads the default associations for a Target or a list of Targets
   """
   def preload_defaults_for_target(target_or_targets, opts \\ []) do
--- a/backend/lib/edgehog/update_campaigns/push_rollout.ex
+++ b/backend/lib/edgehog/update_campaigns/push_rollout.ex
@@ -23,6 +23,13 @@ defmodule Edgehog.UpdateCampaigns.PushRollout do
   import Ecto.Changeset
   alias Edgehog.UpdateCampaigns.PushRollout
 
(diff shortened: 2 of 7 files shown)


backend/lib/edgehog/update_campaigns.ex
```diff
@@ -462,6 +462,23 @@ defmodule Edgehog.UpdateCampaigns do
...
+      iex> update_update_campaign(update_campaign, %{field: new_value})
+      {:ok, %UpdateCamapaign{}}
+
+      iex> update_update_campaign(update_campaign, %{field: bad_value})
+      {:error, %Ecto.Changeset{}}
+
+  """
+  def update_update_campaign(%UpdateCampaign{} = update_campaign, attrs) do
+    UpdateCampaign.changeset(update_campaign, attrs)
+    |> Repo.update()
```
szakhlypa, line 479:
```suggestion
    changeset = UpdateCampaign.changeset(update_campaign, attrs)
    
    with {:ok, update_campaign} <- Repo.update(changeset) do
      {:ok, preload_defaults_for_update_campaign(update_campaign)}
    end
```

backend/lib/edgehog/update_campaigns/push_rollout.ex
szakhlypa, line 1: Maybe better to have `def create_changeset` / `def update_changeset` and `defp changeset`? It seems more explicit.

backend/lib/edgehog_web/schema/update_campaigns_types.ex
```diff
@@ -140,6 +140,47 @@ defmodule EdgehogWeb.Schema.UpdateCampaignsTypes do
...
+    The timeout (in seconds) Edgehog has to wait before considering an OTA \
+    Request lost (and possibly retry).
+    """
+    field :ota_request_timeout_seconds, :integer
+
+    @desc """
+    This boolean flag determines if the Base Image will be pushed to the \
+    Device even if it already has a greater version of the Base Image.
+    """
+    field :force_downgrade, :boolean
```
szakhlypa, line 181: I think we should not allow to update `force_downgrade` (at least for now).
Otherwise, let's discuss how fleet execution should behave on update if campaign is in progress and there are some skipped devices that have greater version of the Base Image.

backend/lib/edgehog_web/schema/update_campaigns_types.ex
```diff
@@ -373,5 +418,31 @@ defmodule EdgehogWeb.Schema.UpdateCampaignsTypes do
...
+      end
+
+      output do
+        @desc "The updated Update Campaign."
+        field :update_campaign, non_null(:update_campaign)
+      end
+
+      middleware Absinthe.Relay.Node.ParseIDs,
+        # base_image_id: :base_image,
+        # update_channel_id: :update_channel,
```
szakhlypa, line 442:
```suggestion
```

backend/lib/edgehog/update_campaigns.ex
szakhlypa: Or maybe it should be `UpdateCampaign.update_changeset`

backend/lib/edgehog_web/schema/update_campaigns_types.ex
noaccOS: I've removed it for the time being

backend/lib/edgehog/update_campaigns/push_rollout.ex
noaccOS: I don't love this idea as
- it is inconsistent with what we already have, although a refactor would be quite easy
- having a `changeset/2` function is the default behavior and having it private looks counter intuitive to me
 
If we decide to go with `create_changeset` and `update_changeset`, I'd still leave a `changeset/2` public

rbino:
I guess we can close this, since this will need to be reimplemented using Ash

PR #310, opened 2023-06-23
fix(seeds): improper `Code.eval_file/1` usage

noaccOS:
from the [docs](https://hexdocs.pm/elixir/Code.html#eval_file/2), `eval_file` _"returns the evaluation result and its binding"_, but it was previously matched against just the return value, which failed to execute the whole seed.

### DIFF
--- a/backend/lib/release.ex
+++ b/backend/lib/release.ex
@@ -1,7 +1,7 @@
 #
 # This file is part of Edgehog.
 #
-# Copyright 2021 SECO Mind Srl
+# Copyright 2021-2023 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


backend/lib/release.ex
```diff
@@ -41,7 +41,9 @@ defmodule Edgehog.Release do
     priv_dir = "#{:code.priv_dir(@app)}"
     seeds_file = Path.join([priv_dir, "repo", "seeds.exs"])
 
-    {:ok, _} = Code.eval_file(seeds_file)
+    {:ok, _} =
+      Code.eval_file(seeds_file)
+      |> elem(0)
```
bettio, line 46: it is a common opinion to think that `elem(n)` (except some specific cases) is less readable than pattern matching.
I suggest doing `{{:ok, _result}, _bindings}` or something like this.

bettio (changes requested):

noaccOS:
closed in favor of #311

PR #312, opened 2023-06-30
Update Erlang and Elixir

bettio:


Pavinati (approved):

.github/workflows/backend-test.yaml
```diff
@@ -78,9 +76,9 @@ jobs:
       with:
         path: backend/priv/plts
         key: |
-          ${{ runner.os }}-dialyzer-plt-${{ env.otp_version }}-${{ env.elixir_version }}
+          ${{ runner.os }}-dialyzer-plt-${{ steps.beam.outputs.otp-version }}-${{ steps.beam.outputs.elixir-version }}
```
szakhlypa, line 79: You should run step `erlef/setup-beam@v1` before `actions/cache@v3` to get data from it's output

szakhlypa:
I think we should also add `.tool-versions` into paths (`on.push.paths` and `on.pull_request.paths`) of updated workflows yaml files. 
Pros: run workflow on changes in `.tool-versions`
Cons: changing nodejs version will also trigger workflow.

.github/workflows/docs-workflow.yaml
```diff
@@ -60,10 +60,10 @@ jobs:
       run: npm install -g @astarte-platform/astarte-docs-cli@0.0.7
     - name: Generate Interfaces Docs
       run: astarte-docs interfaces gen-markdown -d ./edgehog-astarte-interfaces -o ./edgehog/doc/pages/integrating/astarte_interfaces.md
-    - uses: erlef/setup-beam@v1.15
+    - uses: erlef/setup-beam@v1
       with:
-        otp-version: "26.0.1"
-        elixir-version: "1.15.0-rc.2"
+        version-file: .tool-versions
```
szakhlypa, line 65: For with workflow we do repo checkout into `edgehog` path, seems that we need to use `edgehog/.tool-versions` as 
 `version-file` (haven't test it yet)

Pavinati:
> Cons: changing nodejs version will also trigger workflow.

@szakhlypa we could have 2 `.tool-versions` files, one for the frontend one for the backend. It doesn't have to be in the root folder

szakhlypa (approved):

PR #322, opened 2023-07-11
refactor: add fetch functions to Repo module

noaccOS:
they work like get functions but always return a tuple {:ok, item} or {:error, :not_found}


backend/lib/edgehog/repo.ex
```diff
@@ -35,6 +35,24 @@ defmodule Edgehog.Repo do
     Process.get(@tenant_key)
   end
 
+  def fetch(queryable, id, opts \\ []) do
+    empty = Keyword.get(opts, :empty_return, :not_found)
```
rbino, line 39: I'd call this `:error` instead of `:empty_return`

rbino (changes requested):

backend/lib/edgehog/repo.ex
```diff
@@ -35,6 +35,24 @@ defmodule Edgehog.Repo do
     Process.get(@tenant_key)
   end
 
+  def fetch(queryable, id, opts \\ []) do
+    empty = Keyword.get(opts, :error, :not_found)
```
szakhlypa, line 39: Should we also remove `:error` tuple from `opts` before passing them to `get`/`get_by`?
```suggestion
    {empty, opts} = Keyword.pop_first(opts, :error, :not_found)
```

noaccOS: done

rbino (approved):

PR #325, opened 2023-07-14
feat(update_campaign_executor): handle update campaign updates

noaccOS:
depends on #305 obviously


rbino:
Closing since this will need to be reimplemented using Ash

PR #330, opened 2023-07-19
Resume Update Campaigns at startup

rbino:
Pick them up again after an application restart

### DIFF
--- a/backend/test/edgehog/update_campaigns/resumer/core_test.exs
+++ b/backend/test/edgehog/update_campaigns/resumer/core_test.exs
@@ -1,0 +1,127 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2023 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


szakhlypa:
Correct me if I'm wrong but I did not found where after `:resume_campaign`  executor subscribes on ota_operation updates.
Also, as `Core.start_target_update` may create `%OTAOperation` and fail to update `%UpdateCampaigns.Target` should we count `in_progress_count` using `pending_ota_operation`?

rbino:
@szakhlypa both comments are legit.
The first part of the comment is addressed here: https://github.com/edgehog-device-manager/edgehog/pull/349

rbino:
The second part must remain like this temporarily (note that it's unlikely that the Target update fails). This would have to be addressed when we support canceling campaigns as indicated in the comment [here](https://github.com/edgehog-device-manager/edgehog/blob/f82fa642539e06779a0c906a4c86dd1bd9703425/backend/lib/edgehog/update_campaigns/push_rollout/core.ex#L95)

backend/test/edgehog/update_campaigns/resumer/core_test.exs
```diff
@@ -0,0 +1,127 @@
...
+
+    list
+  end
+
+  defp count_fun_called(acc \\ 0) do
+    receive do
+      :fun_called ->
+        count_fun_called(acc + 1)
+    after
+      100 ->
```
szakhlypa, line 123:
```suggestion
      1000 ->
```
We already changed receive timeouts from 100 to 1000 for other tests in #338 , I think we should update this one too.

szakhlypa (approved):

PR #335, opened 2023-07-20
Fix timeouts during Base Image uploads

rbino:
Fix both ephemeral and persistent cases


szakhlypa (changes requested):
Please sign off both commits

szakhlypa (approved):

PR #344, opened 2023-07-26
executor: don't terminate immediately in case of failure

### LINKED ISSUE #342: UpdateCampaign Executor misses target updates after a failure
If a campaign reaches its failure threshold, the Executor is terminated. However this means that all the status updates for the in progress targets get lost (since the PubSub lives in the executor progress), leading to a status that is not coheren with their respective OTA Operations (e.g. an OTA Operation for a target might show a success status while the target itself is shown as In Progress in the Update Campaign since it didn't receive the OTA Operation updates).

We should probably keep the Executor process alive even after a failure to wait for pending targets. The problem with this approach is that we might wait indefinitely (if, e.g., a device disconnects in the middle of the update and never connects back), so we should probably define a timeout to terminate the Executor anyway and document the fact that the Update Campaign status (in case of a failure) might not reflect 100% correctly the status of targets that were in progress when the failure was declared.

rbino:
We want to wait for all in progress target updates to provide a consistent state of the Update Campaign. If we receive no updates, we terminate anyway after a grace period.
Change existing tests and add a new one to test the new behaviour.
Close #342

### DIFF
--- a/backend/lib/edgehog/update_campaigns/push_rollout/executor.ex
+++ b/backend/lib/edgehog/update_campaigns/push_rollout/executor.ex
@@ -332,7 +332,23 @@ defmodule Edgehog.UpdateCampaigns.PushRollout.Executor do
       Core.get_update_campaign!(data.update_campaign_id)
       |> Core.mark_update_campaign_as_failed!()
 
-    {:stop, :normal}
+    if targets_in_progress?(data) do
+      # Here we don't terminate immediately, otherwise we would lose all the updates for the targets
+      # that are currently in progress. If all the remaining targets reach a final state, we will
+      # terminate while handling the relative :ota_operation_completion internal event. Otherwise,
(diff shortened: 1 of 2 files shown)


szakhlypa:
If some ota_operation fails after campaign failure, `def handle_event(:enter, _old_state, :campaign_failure, data)` will be executed twice. This will duplicate failure notice in logs, update `completion_timestamp`, ecc. Is this desired behavior?

rbino:
@szakhlypa the OTP documentation seems to indicate that `state_enter` callbacks will not be called again if you're already in the target state, unless you use `:repeat_state`, so we shouldn't pass twice there if we're already in the `:campaign_failure` state.
See [here](https://www.erlang.org/doc/design_principles/statem.html#state-enter-calls) and [here](https://www.erlang.org/doc/man/gen_statem.html)

szakhlypa (approved):

PR #345, opened 2023-07-26
Return newer update campaign first when listing them

### LINKED ISSUE #341: Reverse UpdateCampaigns query order
We should return the UpdateCampaigns list with the newest campaigns first, because it makes more sense for the frontend to show them at the top

rbino:
Document the new behaviour and add test to verify it. Close #341.

### DIFF
--- a/backend/lib/edgehog/update_campaigns.ex
+++ b/backend/lib/edgehog/update_campaigns.ex
@@ -337,7 +337,8 @@ defmodule Edgehog.UpdateCampaigns do
   end
 
   @doc """
-  Returns the list of update campaigns.
+  Returns the list of update campaigns, in descending creation order (i.e. update campaigns created
+  more recently are returned first)
 
   ## Examples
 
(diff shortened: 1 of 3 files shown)


backend/lib/edgehog/update_campaigns.ex
```diff
@@ -346,7 +347,13 @@ defmodule Edgehog.UpdateCampaigns do
 
   """
   def list_update_campaigns do
-    Repo.all(UpdateCampaign)
+    # We also order by id since inserted_at has a resolution of about 1 second, which could not be
+    # enough to define the proper order if two update campaigns are created within the same second
+    query =
+      from UpdateCampaign,
+        order_by: [desc: :inserted_at, desc: :id]
```
szakhlypa, line 354: We can order by id only, otherwise I think we need to add index on `:inserted_at, :id`.

szakhlypa (approved):

PR #346, opened 2023-07-27
push_rollout: rename max_errors_percentage to max_failure_percentage

### LINKED ISSUE #287: Push rollout: `max_errors_percentage` should be named `max_failure_percentage`
We recently changed the naming around the OTA Updates using "Error" for temporary (possibly recoverable) errors and "Failure" to indicate the final unsuccessful state. The aforementioned property is relative to the second case, so it should be renamed accordingly.

rbino:
This fixes two different problems: one, the plural "errors" was grammatically incorrect, and the correct concept is "failure", since that's the term we use for exceptions which result in a final state, while error is a possibly temporary condition.
Add a migration to rename the field in the database and replace all occurrences both in frontend and backend. The only remaining occurrence is in the docs, but they have to be completely revamped anyway so it doesn't make sense to touch them here.

Close #287

### DIFF
--- a/backend/lib/edgehog/update_campaigns/push_rollout.ex
+++ b/backend/lib/edgehog/update_campaigns/push_rollout.ex
@@ -26,7 +26,7 @@ defmodule Edgehog.UpdateCampaigns.PushRollout do
   @primary_key false
   embedded_schema do
     field :force_downgrade, :boolean, default: false
-    field :max_errors_percentage, :float
+    field :max_failure_percentage, :float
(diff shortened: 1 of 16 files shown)


szakhlypa:
Will `Roll-out mechanism` section of `doc/pages/user/ota_updates.md` be updated in separate PR?

szakhlypa (approved):

rbino:
@szakhlypa yes, as indicated in the commit message

PR #355, opened 2023-08-08
[frontend] Show OTA operations for Update Campaign Targets

szakhlypa:
## Round 3

Divide Update Targets in tabs by their statuses. Configure visible columns based on Update Target status. (Same as in round 2). 
Add donut chart with Update Campaign stats.

Screenshot HD: Update Campaign page with chart (legend position `bottom`)
hd Screen Shot 2023-08-30 at 12 03 39-fullpage

Screenshot FHD: Update Campaign page with chart (legend position `right`)
fhd Screen Shot 2023-08-30 at 12 03 26-fullpage

## Round 2 (outdated)

Divide Update Targets in tabs by their statuses. Configure visible columns based on Update Target status.
Group Update Targets by status and show in progress bar.

Update Campaign page with tabs (`Successful` is active by default)
Screenshot from 2023-08-11 10-34-12

`Successful` tab with `Device` and `Completed at` columns
Screenshot from 2023-08-11 10-36-20

`Failed` tab with `Device`, `Operation Code` and `Completed at` columns
Screenshot from 2023-08-11 10-35-16

`In progress` tab with `Device`, `Operation`, `Operation progress` and `Latest attempt at` columns
Screenshot from 2023-08-11 10-35-28

`Idle` tab with only `Device` column
Screenshot from 2023-08-11 10-35-43

## Round 1 (outdated)

Add SubComponent feature for Table component.
Add new columns for Update Targets table on Update Campaign page:
- updateTarget info: Latest attempt, Retry #, Completed at.
- otaOperation info: Operation, Operation progress.
Add option to display otaOperation's statusCode and message when available.

Update Targets with new GraphQL data
Screenshot from 2023-08-08 10-12-10

Expanded row with error code and message
Screenshot from 2023-08-08 10-12-37


frontend/src/components/Table.tsx
```diff
@@ -147,13 +181,22 @@ const Table = <T extends RowData>({
...
+            <Fragment key={row.id}>
+              <tr {...(getRowProps ? getRowProps(row) : {})}>
+                {row.getVisibleCells().map((cell) => (
+                  <td key={cell.id}>
+                    {flexRender(cell.column.columnDef.cell, cell.getContext())}
+                  </td>
+                ))}
+              </tr>
+              {SubComponent && row.getIsExpanded() && (
+                <tr id={`row-subcomponent-${row.id}`}>
```
Pavinati, line 202: Here we're adding an optional whole new row which can mess up other table features like alternating row background colors.
Here it looks like we're abusing tables as a layout tool. If that's the case i would rather create a new component that uses React table, but actually renders items in a flex/grid layout.

Pavinati: By looking on how that `SubComponent` is used by later commits, i would rather expand the single `Operation` cell rather than the whole row.
Instead of mapping GraphQL fields 1-to-1 with table columns, we could have a single colum called `status` which represents the status/progress and optionally some errors

davidebriani (approved):

PR #357, opened 2023-08-17
Add stats to UpdateCampaign

rbino:
For now, they expose the target count for each of the different statuses, and the total target count


backend/lib/edgehog/update_campaigns.ex
```diff
@@ -498,4 +499,38 @@ defmodule Edgehog.UpdateCampaigns do
...
+      from t in Target,
+        where: t.update_campaign_id == ^update_campaign_id,
+        group_by: t.status,
+        select: {t.status, count(t.id)}
+
+    query
+    |> Repo.all()
+    |> Enum.into(%{})
+    |> case do
+      empty_map when map_size(empty_map) == 0 ->
```
szakhlypa, line 519:
```suggestion
      empty_map when empty_map == %{} ->
```

szakhlypa (changes requested):

szakhlypa (approved):

PR #360, opened 2023-09-01
[frontend] Migrate to Vite

szakhlypa:
This PR migrates the frontend's development environment from [Create React App](https://create-react-app.dev/) to [Vite](https://vitejs.dev/).
Notable changes:

- Webpack and custom scripts are removed, and configuring the development environment is now done with Vite and its plugins.
- Proxy definitions for the Webpack's dev server are removed. The frontend now is just provided directly with the backend's URL.
- The [Jest](https://jestjs.io/) test runner has been removed in favor of [Vitest](https://vitest.dev/), an almost drop-in replacement.
- Command `npm run jest:test` has been renamed to `npm run test`, `npm run jest:coverage` to `npm run coverage`
- Other commands (`npm run start` and `npm run build`) and the development UX stay the same.


frontend/src/react-app-env.d.ts
```diff
@@ -19,6 +19,3 @@
 */
 
 /// <reference types="react-scripts" />
-declare module "babel-plugin-relay/macro" {
-  export { graphql as default } from "react-relay";
-}
```
davidebriani, line 24: I think we can remove the entire file, not just these lines, since we're not using `react-scripts` anymore.

Instead, you might want to have a `vite.env.d.ts` file with this content to support TypeScript intellisense:
```
/// <reference types="vite/client" />
/// <reference types="vitest" />
```

frontend/vite.config.ts
```diff
@@ -1,20 +1,30 @@
...
+      open: env.mode !== 'test',
       port: 3000,
     },
     build: {
       outDir: "build",
     },
     plugins: [react(), eslint(), viteTsconfigPaths(), relay()],
+    test: {
+      environment: "jsdom",
+      globals: true,
```
davidebriani, line 22: By enabling globals, you should be able to avoid explicitly importing `vi` in each test file.
Just remember to add `"types": ["vitest/globals"]` to the `compilerOptions` in the tsconfig to support Typescript types, as explained in [Vitest docs](https://vitest.dev/config/#globals).

davidebriani (changes requested):

frontend/vite.config.ts
szakhlypa: `vitest` was added as dev dependency, so we do not install it while building docker image `frontend/Dockerfile` (`RUN npm ci --omit=dev`).
In build command `tsc -P tsconfig.build.json && vite build` first part (`tsc -P tsconfig.build.json`) is used for type-checking and second one (`vite build`) runs build using Rollup bundler. Having `"types": ["vitest/globals"]` in `compilerOptions` will fail first part (as no dev dependency will be present).

We can:
- add `vitest` as NOT dev dependency and update `compilerOptions` with `vite/globals`
- stick with named imports from `vitest` (also right now there are lack of `it` and `expect` imports it test files, WIP) and remove `globals: true` from `test` section.

Please let me know what would you prefer, thanks.

frontend/src/react-app-env.d.ts
szakhlypa: This file will be removed in next "clean up" commit.
Triple slash directives for `vite/client` and `vitest` for intellisense were added into `frontend/vite.config.ts`.

frontend/vite.config.ts
davidebriani: I really have no preference here, see what's suits you best :)

frontend/src/components/MultiSelect.tsx
```diff
@@ -98,13 +92,13 @@ type MultiSelectCreatableProps =
 type MultiSelectProps<Option> = MultiSelectBaseProps<Option> &
   MultiSelectCreatableProps;
 
-const MultiSelect = <Option extends any>({
+const MultiSelect = ({
```
davidebriani, line 95: Not sure why you removed the generic parameter here.
I think it was there to help infer types for all component's props; for instance, when you supply an array of `TargetGroup` as `options`, then the callbacks `getOptionLabel` and `getOptionValue` would be automatically typed as `(option: TargetGroup) => string`.

davidebriani (approved):

frontend/src/components/MultiSelect.tsx
szakhlypa: You are totally right.
There were `@typescript-eslint/no-unnecessary-type-constraint` errors on `Option extends any` as `any` does not add restrictions on which types can be used in `Option` but I missed inferring types for other component's props part.
Updated, thanks.

frontend/src/react-app-env.d.ts
szakhlypa: Moved triple slash directives into `frontend/src/vite-env.d.ts`.

davidebriani (approved):

Pavinati (approved):

PR #384, opened 2023-10-06
CI: use a single workflow to build all docs

rbino:
Avoid the two workflows stomping on each other's feet


.github/workflows/docs-workflow.yaml
```diff
@@ -77,7 +101,10 @@ jobs:
         # Restore Device SDK docs. Don't fail if they're not there.
         cd docs && git restore $DOCS_DIRNAME/device-sdks || true && cd ..
         cd docs && git restore $DOCS_DIRNAME/graphql-api-docs || true && cd ..
```
szakhlypa, line 103: Why do we need to restore `graphql-api-docs` if than we make copy of new generated `./edgehog/backend/graphql-api-docs`?

rbino: Legit, leftover, removing it

szakhlypa (approved):

PR #391, opened 2023-10-13
[CI] Add jobs to build Docker images

### LINKED ISSUE #389: Add workflow to test docker builds
Add workflow to test that Docker can successfully build `frontend` and `backend` images.

szakhlypa:
- add jobs to build Docker images for backend and frontend
- trigger workflows on pushes to `release-*` branch

Fixes #389

### DIFF
--- a/.github/workflows/backend-test.yaml
+++ b/.github/workflows/backend-test.yaml
@@ -29,6 +29,7 @@ on:
     - '.github/workflows/backend-test.yaml'
     branches:
     - 'main'
+    - 'release-*'
   # Run on pull requests matching apps
   pull_request:
     paths:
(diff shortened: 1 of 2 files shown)


.github/workflows/backend-test.yaml
```diff
@@ -168,3 +169,13 @@ jobs:
 
     - name: Test
       run: mix coveralls.github -o coverage_results
+
+  build-docker-image:
+    name: Build Docker image
+    runs-on: ubuntu-latest
```
Pavinati, line 175: Depending on the OS version, Erlang might complain about incompatibilities with the SSL library used.
I would suggest using a fixed tag, Same goes for the image built on the `Dockerfile`

Pavinati (changes requested):

.github/workflows/backend-test.yaml
szakhlypa: The purpose of new jobs is to ensure that docker image can be built successfully. Building process should be independent of host machine (the only requirement is docker itself).
Any custom requirement for backend/frontend build should be resolved inside correspective `Dockerfile` and this jobs will help to test this behavior also.

Issue #390, opened 2023-10-11
Add more unit tests


szakhlypa:
Let's start with components (`frontend/src/components/`) without relay fragments:

- [x] Center
- [x] Footer
- [x] Map
- [x] Page
- [x] Tabs

PR #392, opened 2023-10-16
Add test for the Center component

harisbikovic:
Related: #390 
Added tests for the Center component.

### DIFF
--- a/frontend/src/components/Center.test.tsx
+++ b/frontend/src/components/Center.test.tsx
@@ -1,0 +1,32 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2023 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


frontend/src/components/Center.test.tsx
```diff
@@ -0,0 +1,33 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2021-2023 SECO Mind Srl
```
szakhlypa, line 4:
```suggestion
  Copyright 2023 SECO Mind Srl
```

frontend/src/components/Center.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+import Center from "./Center";
+
+it("renders with the passed className", () => {
+  const { container } = render(
+    <Center children="innerText" className="test-class" />,
+  );
+  expect(container.firstChild).toHaveClass(
+    "h-100 d-flex flex-column justify-content-center align-items-center test-class",
+  );
+});
```
szakhlypa, line 33: We do not need to test actual implementation of `Center` component, I think it's better test that passing `foo-class` in `className` property this class is present in rendered container, not to check for every class. 
For example:

```suggestion
it("renders with the passed className", () => {
  const { container } = render(
    <Center children="innerText" className="foo-class" />,
  );
  expect(container.querySelector(".foo-class")).toBeInTheDocument();
});
```

szakhlypa (changes requested):

szakhlypa (approved):

szakhlypa (approved):

PR #397, opened 2023-10-18
add tests for Footer component

harisbikovic:
Related #390
This PR containes tests for the Footer component.

### DIFF
--- a/frontend/src/components/Footer.test.tsx
+++ b/frontend/src/components/Footer.test.tsx
@@ -1,0 +1,90 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2023 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  SPDX-License-Identifier: Apache-2.0
+*/
+
+import { it, expect } from "vitest";
+import { screen, within } from "@testing-library/react";
+import { renderWithProviders } from "setupTests";
+
+import Footer from "./Footer";
+
+it("displays correct app name and app version", () => {
+  renderWithProviders(
+    <Footer
+      appName="Edgehog Device Manager"
+      appVersion="0.1.0-alpha.1"
+      homepageUrl="https://edgehog.io/"
+      repoUrl="https://github.com/edgehog-device-manager/edgehog"
+      issueTrackerUrl="https://github.com/edgehog-device-manager/edgehog/issues"
+    />,
+  );
+
+  const appNameElement = screen.getByText("Edgehog Device Manager");
+  expect(appNameElement).toBeVisible();
+
+  const appVersionElement = screen.getByText(/0\.1\.0-alpha\.1/);
+  expect(appVersionElement).toBeVisible();
+});
+
+it("app logo points to the homepage URL", () => {
+  renderWithProviders(
+    <Footer
+      appName="Edgehog Device Manager"
+      appVersion="0.1.0-alpha.1"
+      homepageUrl="https://edgehog.io/"
+      repoUrl="https://github.com/edgehog-device-manager/edgehog"
+      issueTrackerUrl="https://github.com/edgehog-device-manager/edgehog/issues"
+    />,
+  );
+
+  const homepageLink = screen.getByRole("link", {
+    name: "Edgehog Homepage Link",
+  });
+  expect(homepageLink).toBeVisible();
+  expect(homepageLink).toHaveAttribute("href", "https://edgehog.io/");
+
+  const appLogo = within(homepageLink).getByRole("img");
+  expect(appLogo).toBeVisible();
+});
+
+it("anchors point to correct URL's", () => {
+  renderWithProviders(
+    <Footer
+      appName="Edgehog Device Manager"
+      appVersion="0.1.0-alpha.1"
+      homepageUrl="https://edgehog.io/"
+      repoUrl="https://github.com/edgehog-device-manager/edgehog"
+      issueTrackerUrl="https://github.com/edgehog-device-manager/edgehog/issues"
+    />,
+  );
+
+  const repositoryLink = screen.getByRole("link", { name: "GitHub" });
+  expect(repositoryLink).toBeVisible();
+  expect(repositoryLink).toHaveAttribute(
+    "href",
+    "https://github.com/edgehog-device-manager/edgehog",
+  );
+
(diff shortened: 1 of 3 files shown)


szakhlypa:
I prefer to not depend on link's order.
What if we add some more accessibility into component with `aria-label` and use it in `getByRole`?
For example:
repository link in `Footer.tsx`:
```typescript
  <a
    href={repoUrl}
    _other_props_here_
    aria-label={
      intl.formatMessage({
        id: "components.Footer.repositoryLinkLabel",
        defaultMessage: "GitHub",
      })
    }
  >
    <Icon icon="github" />
  </a>
```
And its test with `getByRole`
```typescript
  const repositoryLink = screen.getByRole("link", { name: "GitHub" });
  expect(repositoryLink).toHaveAttribute("href", "https://github.com/edgehog-device-manager/edgehog");
```

Also, can we use different values for `homepageUrl` and `repoUrl` props in test to distinguish in case of error.

frontend/src/components/Footer.test.tsx
szakhlypa, line 1: Please fix Copyright header

frontend/src/components/Footer.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+it("anchors redirect to correct links", () => {
+  renderWithProviders(
+    <Footer
+      appName="appName"
+      appVersion="appVersion"
+      homepageUrl="https://edgehog.io/"
+      repoUrl="https://github.com/edgehog-device-manager/edgehog"
+      issueTrackerUrl="https://github.com/edgehog-device-manager/edgehog/issues"
+    />,
+  );
```
szakhlypa, line 36: We pass 5 properties but test only 2 of them. Can you also add tests for others?

frontend/src/i18n/langs/en.json
szakhlypa, line 1: Please, whenever you auto generate "stuff" (like with `npm run i18n:extract`), always put it in its own commit.

szakhlypa (changes requested):

frontend/src/components/Footer.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const appNameElement = screen.getByText("appName");
+  const appVersionElement = screen.getByText("appName");
+
+  expect(appNameElement).toBeInTheDocument();
```
szakhlypa, line 41: We can use [`toBeVisible`](https://github.com/testing-library/jest-dom#tobevisible) to check if an element is currently visible to the user.
```suggestion
  expect(appNameElement).toBeVisible();
```

frontend/src/components/Footer.test.tsx
```diff
@@ -0,0 +1,84 @@
...
+  expect(appVersionElement).toBeInTheDocument();
```
szakhlypa, line 42: Do you mind grouping node query with expectations? This way we can test node-by-node, e.g.:
```typescript
const foo = screen.getByText("foo");
expect(foo).toBeVisible();

const bar = screen.queryByText("bar");
expect(bar).not.toBeInTheDocument();
```

frontend/src/components/Footer.tsx
```diff
@@ -22,6 +22,7 @@ import assets from "assets";
 import Center from "components/Center";
 import Icon from "components/Icon";
 import Stack from "components/Stack";
+import { useIntl } from "react-intl";
```
szakhlypa: You add `useIntl` import in first commit, than move it upper in second one.
I like the idea to have ordered imports (we can add [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import) in future), but can you move this import on top in first commit, so we do not need to change it's position in second one?

frontend/src/i18n/langs/en.json
szakhlypa: Can we structure this PR to have 3 commits?
1. Changes to `components/Footer.tsx`.
2. i18n changes (`i18n/langs/en.json`). Whenever you automatically generate "stuff" (e.g. with `npm run i18n:extract`), always put it in a separate commit.
3. New test file `components/Footer.test.tsx`.

szakhlypa (changes requested):

frontend/src/components/Footer.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+  );
+
+  const appLogo = screen.getByRole("img");
+  expect(appLogo).toBeVisible();
+
+  const homepageLink = screen.getByRole("link", {
+    name: "Edgehog Homepage Link",
+  });
+  expect(homepageLink).toBeVisible();
+  expect(homepageLink).toHaveAttribute("href", "https://edgehog.io/");
```
szakhlypa, line 63: If the goal is to check whether link to the homepage has an image - we can use [`within`](https://testing-library.com/docs/dom-testing-library/api-within/) and query `getByRole("img")` only within the link. Ofc we need to import it with `import { screen, within } from "@testing-library/react";` 

```suggestion
  const homepageLink = screen.getByRole("link", {
    name: "Edgehog Homepage Link",
  });
  expect(homepageLink).toBeVisible();
  expect(homepageLink).toHaveAttribute("href", "https://edgehog.io/");

  const appLogo = within(homepageLink).getByRole("img");
  expect(appLogo).toBeVisible();
```

szakhlypa (changes requested):

frontend/src/components/Footer.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      homepageUrl="https://edgehog.io/"
+      repoUrl="https://github.com/edgehog-device-manager/edgehog"
+      issueTrackerUrl="https://github.com/edgehog-device-manager/edgehog/issues"
+    />,
+  );
+
+  const appNameElement = screen.getByText("Edgehog Device Manager");
+  expect(appNameElement).toBeVisible();
+
+  const appVersionElement = screen.getByText(/0.1.0-alpha.1/i);
```
szakhlypa, line 41: Do we want to ignore case?
```suggestion
  const appVersionElement = screen.getByText(/0.1.0-alpha.1/);
```

szakhlypa (changes requested):

frontend/src/components/Footer.test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      homepageUrl="https://edgehog.io/"
+      repoUrl="https://github.com/edgehog-device-manager/edgehog"
+      issueTrackerUrl="https://github.com/edgehog-device-manager/edgehog/issues"
+    />,
+  );
+
+  const appNameElement = screen.getByText("Edgehog Device Manager");
+  expect(appNameElement).toBeVisible();
+
+  const appVersionElement = screen.getByText(/0.1.0-alpha.1/);
```
szakhlypa, line 41: Dot in RegExp matches any single character except line terminators, so we need to escape it with backslash. Otherwise test will pass even with `appVersion="0x1x0-alphax1"` property.
```suggestion
  const appVersionElement = screen.getByText(/0\.1\.0-alpha\.1/);
```

frontend/src/components/Footer.tsx
szakhlypa, line 1: Can you update year in copyright header please?

szakhlypa (changes requested):

szakhlypa (approved):

Issue #290, opened 2023-06-05
Add 'housekeeping' API to create and delete tenants


davidebriani:
A set of APIs should be made available to perform the following operations:
- Create an Edgehog tenant.
  - It should be possible to specify the private key for the tenant.
  - It should be possible to specify a reference to an Astarte realm and its private key.
- Delete an Edgehog tenant.

It's safe to assume that the associated Astarte realm will be created and deleted out-of-band.

The APIs can have an interface separated from the existing GraphQL schema, e.g. REST APIs.

rbino:
We should also define how this API will be authenticated

PR #400, opened 2023-10-25
[backend] Add Admin REST API

szakhlypa:
Add Admin REST API on `/admin-api/`.
Require JWT authentication with validation via public key.
Expose tenant provisioning via POST to `/admin-api/v1/tenants`.

Part of #290.

### DIFF
--- a/backend/lib/edgehog/config.ex
+++ b/backend/lib/edgehog/config.ex
@@ -23,8 +23,21 @@ defmodule Edgehog.Config do
   This module handles the configuration of Edgehog
   """
   use Skogsra
-  alias Edgehog.Config.{GeocodingProviders, GeolocationProviders}
+  alias Edgehog.Config.{GeocodingProviders, GeolocationProviders, JWTPublicKeyPEMType}
   alias Edgehog.Geolocation
+
+  @envdoc """
+  Disables admin authentication. CHANGING IT TO TRUE IS GENERALLY A REALLY BAD IDEA IN A PRODUCTION ENVIRONMENT, IF YOU DON'T KNOW WHAT YOU ARE DOING.
+  """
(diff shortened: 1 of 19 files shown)


backend/lib/edgehog/config.ex
```diff
@@ -71,6 +71,11 @@ defmodule Edgehog.Config do
     type: GeocodingProviders,
     default: [Geolocation.Providers.GoogleGeocoding]
 
+  @envdoc "The Admin API public key."
+  app_env :admin_api_public_key, :edgehog, :admin_api_public_key,
```
rbino, line 75: Not sure which way to go here, but in Astarte Housekeeping API we provide [the path](https://github.com/astarte-platform/astarte/blob/01c851032da01dfa7f476408e75d1ae2a687df0b/apps/astarte_housekeeping_api/lib/astarte_housekeeping_api/config.ex#L39) to a file that contains the public key (using a custom Skogsra type). Let's discuss this to see which approach is better.

backend/lib/edgehog_web/router.ex
```diff
@@ -33,6 +33,8 @@ defmodule EdgehogWeb.Router do
     plug EdgehogWeb.PopulateTenant
   end
 
+  forward "/admin-api", EdgehogWeb.AdminAPI.Router
```
rbino, line 36: Why do we need a separate router for this? We can just add a scope + `pipeline :admin_api` here I guess

rbino (changes requested):

backend/lib/edgehog_web/router.ex
szakhlypa: We may want to divide routes by domain in future, for now - added Admin API scope and pipeline into main router.

PR #401, opened 2023-10-30
add tests for the Tabs component

harisbikovic:
Related: #390 
This PR adds tests for the Tabs component.

### DIFF
--- a/frontend/src/components/Tabs.test.tsx
+++ b/frontend/src/components/Tabs.test.tsx
@@ -1,0 +1,207 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2023-2024 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  SPDX-License-Identifier: Apache-2.0
+*/
+
+import { it, expect, describe } from "vitest";
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+
+import Tabs, { Tab } from "./Tabs";
+
+describe("Tabs", () => {
+  it("assigns className to the Tabs component", () => {
+    render(<Tabs className="custom-tabs-class">Tabs Content</Tabs>);
+
+    const tabsElement = screen.getByText("Tabs Content");
+    expect(tabsElement).toHaveClass("custom-tabs-class");
+  });
+
+  describe("defaultActiveKey", () => {
+    it("when not specified, makes the first tab active", () => {
+      render(
+        <Tabs>
+          <Tab eventKey="tab1" title="Tab 1">
+            <div data-testid="tab1-content" />
+          </Tab>
+          <Tab eventKey="tab2" title="Tab 2">
+            <div data-testid="tab2-content" />
+          </Tab>
+        </Tabs>,
+      );
+
+      expect(
+        screen.getByRole("tab", { name: "Tab 1", selected: true }),
+      ).toBeVisible();
+      expect(screen.getByTestId("tab1-content")).toBeVisible();
+    });
+
+    it("when specified, renders the content of the active tab", () => {
+      render(
+        <Tabs defaultActiveKey="tab2">
+          <Tab eventKey="tab1" title="Tab 1">
+            <div data-testid="tab1-content" />
+          </Tab>
+          <Tab eventKey="tab2" title="Tab 2">
+            <div data-testid="tab2-content" />
+          </Tab>
+        </Tabs>,
+      );
+
+      expect(
+        screen.getByRole("tab", { name: "Tab 2", selected: true }),
+      ).toBeVisible();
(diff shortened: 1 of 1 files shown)


szakhlypa (changes requested):

frontend/src/components/Tabs.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+          </Tab>
+          <Tab eventKey="tab2" title="Tab 2">
+            <div data-testid="tab2-content"></div>
+          </Tab>
+        </Tabs>,
+      );
+      expect(screen.getByTestId("tab1-content")).toBeVisible();
+      //Check if active tab has valid aria-selected attribute
+      expect(
+        screen.getByRole("tab", { name: "Tab 1", selected: true }),
```
szakhlypa, line 87: Can we also add a check that `Tab 2` is NOT selected?

frontend/src/components/Tabs.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+      ).toBeVisible();
+      expect(screen.queryByTestId("tab2-content")).not.toBeInTheDocument();
+
+      await userEvent.click(screen.getByRole("tab", { name: "Tab 2" }));
+
```
szakhlypa, line 92: Let's also check that selected `tab` is changed from `Tab 1` to `Tab 2`

frontend/src/components/Tabs.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+            <div data-testid="tab1-content"></div>
+          </Tab>
+          <Tab eventKey="tab2" title="Tab 2">
+            <div data-testid="tab2-content"></div>
+          </Tab>
+        </Tabs>,
+      );
+
+      expect(screen.getByTestId("tab1-content")).toBeVisible();
+      expect(screen.queryByTestId("tab2-content")).not.toBeInTheDocument();
```
szakhlypa, line 109: Does this like repeat `renders the content of the active tab only` test?

frontend/src/components/Tabs.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+          <Tab eventKey="tab1" title="Tab 1">
+            <div data-testid="tab1-content"></div>
+          </Tab>
+          <Tab eventKey="tab2" title="Tab 2">
+            <div data-testid="tab2-content"></div>
+          </Tab>
+        </Tabs>,
+      );
+
+      expect(screen.getByTestId("tab1-content")).toBeVisible();
```
szakhlypa, line 108: Can we also check the selected tab?

frontend/src/components/Tabs.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+      expect(screen.queryByTestId("tab2-content")).not.toBeInTheDocument();
+    });
+    it("when the key specified is not present in any of the Tabs won't render any Tab's content", () => {
```
szakhlypa, line 111: Is this the intended behaviour?

frontend/src/components/Tabs.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+          <Tab eventKey="tabTwo" title="Tab 2" />
+          <Tab eventKey="tabThree" title="Tab 3" />
+        </Tabs>,
+      );
+      const tabs = screen.getAllByRole("tab");
+      expect(tabs[0]).toHaveTextContent("Tab 2");
+      expect(tabs[1]).toHaveTextContent("Tab 3");
+      expect(tabs[2]).toHaveTextContent("Tab 1");
+    });
+    it("when defaultActiveKey given and tabsOrder differs from the order of children, renders tabs as specified in tabsOrder", () => {
```
szakhlypa, line 183: How does the `tabs` rendering order depend on `defaultActiveKey`?
Do you want to check that if `defaultActiveKey` is present, regardless of the `tabsOrder`, it determines which tab's content is rendered?

frontend/src/components/Tabs.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+        <Tab eventKey="tab1" title="Tab1" className="custom-tab-class">
+          Tab 1
+        </Tab>
+      </Tabs>,
+    );
+
+    const tab1 = screen.getByText("Tab 1");
+    expect(tab1).toHaveClass("custom-tab-class");
+  });
+  it("renders the Tab component with the specified title", () => {
```
szakhlypa, line 226: Does this test repeat the `it("renders tab elements correctly"` above?
Do you want to set `className` here, and if so - what for?

frontend/src/components/Tabs.test.tsx
harisbikovic: Actually line 89 checks that:
`expect(screen.queryByTestId("tab2-content")).not.toBeInTheDocument();`
Is this valid or you have something different on your mind?

frontend/src/components/Tabs.test.tsx
harisbikovic: In this case we test what happens when there is no **defaultActiveKey** specified whereas in the above mentioned `renders the content of the active tab only` test we specify the **defaultActiveKey**. So we're testing the same behaviour but in different scenarious. Is this ok wtih you?

frontend/src/components/Tabs.test.tsx
harisbikovic: Hmm, I'm not sure what can be done about it?

frontend/src/components/Tabs.test.tsx
harisbikovic: You're right, I mistakenly used `defaultActiveKey` in this section. Therefore, there's no need for this test since there's one  test above called `when specified renders tabs in the requested order` that tests the same thing.

szakhlypa (approved):

PR #406, opened 2023-11-15
fix(seeds): restore functionality outside docker

### LINKED ISSUE #393: Restore working seeds
Assuming an empty Postgresql running on the same machine (e.g. `docker run --net=host --name some-postgres --rm -e POSTGRES_PASSWORD=postgresql postgres:latest`), it should be possible to bring up a working backend with just

```bash
git clone git@github.com:edgehog-device-manager/edgehog.git
cd edgehog/backend
mix deps.get
mix ecto.reset
```

Right now, this fails with
```
** (MatchError) no match of right hand side value: {:error, #Ecto.Changeset<action: :insert, changes: %{name: "Test Cluster"}, errors: [base_api_url: {"can't be blank", [validation: :required]}], data: #Edgehog.Astarte.Cluster<>, valid?: false>}
    priv/repo/seeds.exs:45: (file)
    (elixir 1.15.5) lib/code.ex:1435: Code.require_file/2
    (mix 1.15.5) lib/mix/tasks/run.ex:146: Mix.Tasks.Run.run/5
    (mix 1.15.5) lib/mix/tasks/run.ex:85: Mix.Tasks.Run.run/1
    (mix 1.15.5) lib/mix/task.ex:447: anonymous fn/3 in Mix.Task.run_task/5
    (mix 1.15.5) lib/mix/task.ex:506: Mix.Task.run_alias/6
    (mix 1.15.5) lib/mix/cli.ex:92: Mix.CLI.run_task/2
    /home/rbino/.asdf/installs/elixir/1.15.5-otp-26/bin/mix:2: (file)
```

It should be possible to just seed default data without any additional step

noaccOS:
adds a dependency to load the .env file when it exists.

the method `read` is used to get back a map of the loaded values. this is because the `load` method prioritizes variables in the .env file to system variables, a behavior shared by all the other libraries taken in consideration
- dotenv_elixir
- dotenvy
- dotenv-parser

but we want a behavior consistent with docker, where exported variables are prioritized.

closes #393


rbino (approved):

rbino:
@noaccOS I changed the PR to point to 0.7, please rebase this accordingly

noaccOS:
> @noaccOS I changed the PR to point to 0.7, please rebase this accordingly

Done 
also added a line to the changelog

rbino (approved):

rbino:
@noaccOS actually I tried this locally and this probably didn't take into account the fact that now the Astarte private key is actually validated to be a valid PEM, so it still fails.

Please add a default private key instead of `notaprivatekey` and change the error message if no explicit private key is set to "You're using the default private key, this instance will not be able..."

rbino (changes requested):
Need to fix the default private key (which is now validated)

PR #409, opened 2023-11-29
Add lazy creation for resource fixtures

szakhlypa:
Update resource fixtures (listed below) to make nested resources optional and to lazy create them in case of absence.

- system_model_fixture.
- base_image_collection_fixture.
- base_image_fixture.


rbino (approved):
Very nice, cleans up the code and highlights what's actually important (if anything) when creating a fixture in a test.

PR #418, opened 2023-12-07
Add relay fragment to System Model

harisbikovic:
This PR creates GraphQL fragment for data into SystemModelsTable component and use it on System Models page where this data is rendered inside a table. 
This also gets rid of relay/unused-fields lint warnings.

Updates the ESLint plugin for Testing Library as a  dependency using _**npm install eslint-plugin-testing-library --save-dev**_

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -106,7 +106,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.3",
     "eslint-plugin-relay": "^1.8.3",
-    "eslint-plugin-testing-library": "^6.1.0",
+    "eslint-plugin-testing-library": "^6.2.0",
     "jsdom": "^22.1.0",
     "prettier": "^3.0.3",
     "react-select-event": "^5.5.1",
--- a/frontend/src/components/SystemModelsTable.tsx
+++ b/frontend/src/components/SystemModelsTable.tsx
@@ -20,21 +20,33 @@
 
 import React from "react";
 import { FormattedMessage } from "react-intl";
+import { graphql, useFragment } from "react-relay/hooks";
 
 import Table, { createColumnHelper } from "components/Table";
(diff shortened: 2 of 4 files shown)


frontend/src/components/SystemModelsTable.tsx
```diff
@@ -20,21 +20,32 @@
...
 import { FormattedMessage } from "react-intl";
+import { graphql, useFragment } from "react-relay/hooks";
 
 import Table, { createColumnHelper } from "components/Table";
 import { Link, Route } from "Navigation";
+import type {
+  SystemModelsTable_SystemModelsFragment$key,
+  SystemModelsTable_SystemModelsFragment$data,
+} from "../api/__generated__/SystemModelsTable_SystemModelsFragment.graphql";
+// We use graphql fields below in columns configuration
```
szakhlypa, line 31: Let's divide block of imports with new line.
```suggestion
} from "../api/__generated__/SystemModelsTable_SystemModelsFragment.graphql";

// We use graphql fields below in columns configuration
```

szakhlypa (changes requested):

frontend/package.json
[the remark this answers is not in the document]
szakhlypa: Please be more specific: `Update dependencies` does not explain what has been updated. Are these updates for the frontend or the backend? Also, it's better to include the command used to update in the commit description.
One more question, why does the dependency update commit come after the refactoring one? Shouldn't they be in the reverse order?

szakhlypa, line 1: Please be more specific: `Update dependencies` does not explain what has been updated. Are these updates for the frontend or the backend? Also, it's better to include the command used to update in the commit description.
One more question, why does the dependency update commit come after the refactoring one? Shouldn't they be in the reverse order?

szakhlypa (changes requested):

szakhlypa (approved):

PR #420, opened 2023-12-13
Introduce Ash

rbino:
Add Ash dependencies and prepare the ground for refactoring


backend/config/config.exs
```diff
@@ -88,6 +88,9 @@ config :edgehog, Edgehog.PromEx,
   grafana: :disabled,
   metrics_server: :disabled
 
+config :ash, :default_belongs_to_type, :integer
+config :ash_graphql, :default_managed_relationship_type_name_template, :action_name
```
davidebriani, line 92: We can update the year for this file's copyright header to 2023

davidebriani (changes requested):

backend/config/config.exs
rbino: me thinking about the impending new year and consequent Copyright bump: :sob:

.github/workflows/backend-test.yaml
```diff
@@ -167,8 +168,9 @@ jobs:
         mix deps.get --only test
         mix deps.compile
 
-    - name: Test
-      run: mix coveralls.github -o coverage_results
+    # TODO: enable this back after Ash porting is complete
+    # - name: Test
+    #   run: mix coveralls.github -o coverage_results
```
szakhlypa, line 173: I think there is no need for the `test-coverage` job without its last step, so we can disable it at job level for now.

szakhlypa (changes requested):

.github/workflows/backend-test.yaml
rbino: Yes, but I enable it back in the next PR (with `mix test` instead of `mix coveralls`) so we might as well just leave this here. I would've done that here but `mix test` returns non-zero if the `--only` filter doesn't match _any_ tests

szakhlypa (approved):

PR #421, opened 2023-12-13
Port Tenants context to Ash

rbino:
Let's get the ball rolling


backend/lib/edgehog/validations/validations.ex
```diff
@@ -0,0 +1,29 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Validations do
```
davidebriani, line 21: Is it intended that the `Edgehog.Validations` module lives in the file `edgehog/validations/validations.ex` instead of `edgehog/validations.ex`?
I am curious if there are some practices we want to follow with module names and folder structure, or if it doesn't really matter.

backend/lib/edgehog_web/populate_tenant.ex
```diff
@@ -30,12 +30,12 @@ defmodule EdgehogWeb.PopulateTenant do
   def call(conn, _opts) do
     tenant_slug = conn.path_params["tenant_slug"]
 
-    case Tenants.fetch_tenant_by_slug(tenant_slug) do
+    case Tenants.Tenant.by_slug(tenant_slug) do
```
davidebriani, line 33: Let's update the copyright's year

davidebriani (changes requested):

backend/lib/edgehog/tenants/tenant.ex
```diff
@@ -19,34 +19,55 @@
...
+    attribute :public_key, :string, allow_nil?: false
+
+    create_timestamp :inserted_at
+    update_timestamp :updated_at
+  end
+
+  validations do
+    validate match(:slug, Validations.slug_regex()) do
+      message "should only contain lower case ASCII letters (from a to z), digits and -"
+    end
```
rbino, line 60: @szakhlypa it's possible to setup a Spark (the Ash DSL) [formatter plugin](https://hexdocs.pm/ash/development-utilities.html#configuration) that enforces the order of sections inside a resource. There is no predefined order, so we can indulge in some bikeshedding here int the first PR so we come to an agreement and enforce that in all the subsequent ones.

My proposal is: resource, graphql, code_interface, actions, attributes, relationships, calculations, aggregates, validations, preparations, changes, pub_sub, postgres.

The rough idea is going (top to bottom) from user-facing things to internal details.

backend/lib/edgehog/validations/validations.ex
szakhlypa: It's one of the Ash [recommendations](https://www.ash-hq.org/docs/guides/ash/latest/how_to/structure-your-project) on how to structure project:
`If the resource has a folder, we suggest putting the resource.ex in the folder with the same name.`
This way every context file should be inside context directory.

rbino: We decided to adopt [Ash suggested project structure](https://hexdocs.pm/ash/structure-your-project.html). Of course this is not a requirement (modules work regardless of where you put them), but I've come to appreciate this mainly due to this: if you execute `ls` in Edgehog's backend `lib` directory today, you see this

2023-12-14-155308_1801x348_scrot

If instead you move "root" context files _inside_ their directory, there is much less noise

2023-12-14-160009_1717x173_scrot

 You either see single-file contexts (`.ex` files) _or_ you see a context directory. This goes on recursively inside (if a `User` resource just has its resource file it will go in the top context level, if it also has some additional module-based stuff it goes inside its `user` directory).

rbino: Oops took too much to reply and you already figured all this out :smile:

szakhlypa (approved):

PR #422, opened 2023-12-14
Clean up S3 after tenant deletion

szakhlypa:
Clean up S3 storage files asynchronously after successfully deleting a tenant.

### DIFF
--- a/backend/lib/edgehog/application.ex
+++ b/backend/lib/edgehog/application.ex
@@ -53,11 +53,13 @@ defmodule Edgehog.Application do
       {Finch, name: EdgehogFinch},
       # Start the UpdateCampaigns supervisor
       Edgehog.UpdateCampaigns.Supervisor,
-      # Start the Endpoint (http/https)
-      EdgehogWeb.Endpoint,
       # Start the Tenant Reconciler Supervisor
       {Edgehog.Tenants.Reconciler.Supervisor,
-       tenant_to_trigger_url_fun: tenant_to_trigger_url_fun}
+       tenant_to_trigger_url_fun: tenant_to_trigger_url_fun},
+      # Start Supervisor for Provisioning Cleanup Tasks
+      {Task.Supervisor, name: Edgehog.Provisioning.CleanupSupervisor},
+      # Start the Endpoint (http/https)
+      EdgehogWeb.Endpoint
     ]
 
     # See https://hexdocs.pm/elixir/Supervisor.html
--- a/backend/test/edgehog/provisioning_test.exs
+++ b/backend/test/edgehog/provisioning_test.exs
@@ -25,6 +25,8 @@ defmodule Edgehog.ProvisioningTest do
 
   import Edgehog.AstarteFixtures
   import Edgehog.BaseImagesFixtures
+  import Edgehog.DevicesFixtures
+  import Edgehog.OSManagementFixtures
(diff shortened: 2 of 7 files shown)


backend/lib/edgehog/application.ex
```diff
@@ -57,7 +57,9 @@ defmodule Edgehog.Application do
       EdgehogWeb.Endpoint,
       # Start the Tenant Reconciler Supervisor
       {Edgehog.Tenants.Reconciler.Supervisor,
-       tenant_to_trigger_url_fun: tenant_to_trigger_url_fun}
+       tenant_to_trigger_url_fun: tenant_to_trigger_url_fun},
+      # Start Supervisor for Provisioning Tasks
+      {Task.Supervisor, name: Edgehog.ProvisioningTaskSupervisor}
```
rbino, line 62: Nit: this should be started _before_ Endpoint because as it is right now theoretically you could receive a request to the Provisioning controller and not have the cleanup supervisor ready

rbino: Also, use a namespaced (and more specific) name: `Edgehog.Provisioning.CleanupSupervisor`. This way you can alias it without conflicts.

backend/test/edgehog/provisioning_test.exs
```diff
@@ -185,6 +189,203 @@ defmodule Edgehog.ProvisioningTest do
...
+        send(test_pid, {:base_image_deleted, ref})
+
+        :ok
+      end)
+
+      assert {:ok, _tenant} = Provisioning.delete_tenant_by_slug(tenant.slug)
+
+      1..other_base_images_count
+      |> Enum.each(fn _ -> assert_receive {:base_image_deleted, ^ref} end)
+    end
```
rbino, line 276: You're delaying the tests by 4 second to basically test that `Task.async` works as expected, I'd drop this test

backend/test/edgehog/provisioning_test.exs
```diff
@@ -185,6 +189,203 @@ defmodule Edgehog.ProvisioningTest do
...
+        # Clean up task has a default 5_000 ms timeout.
+        if base_image.version == base_image_slow_response.version, do: Process.sleep(6_000)
+
+        :ok
+      end)
+
+      assert {:ok, _tenant} = Provisioning.delete_tenant_by_slug(tenant.slug)
+
+      wait_for_provisioning_tasks()
+    end
```
rbino, line 252: Imho this is still testing the inner working of `Task.async`, we can drop this too

rbino (changes requested):
A couple of nits and a couple of tests that can be dropped then we're good to go

backend/lib/edgehog/application.ex
szakhlypa: Renamed, moved both supervisors (`Reconciler` and `Provisioning`) to start before Endpoint.

backend/test/edgehog/provisioning_test.exs
szakhlypa: Removed unnecessary test

backend/test/edgehog/provisioning_test.exs
szakhlypa: Removed unnecessary test

rbino (approved):

PR #423, opened 2023-12-14
Port Astarte.Cluster resource to Ash

rbino:


backend/test/edgehog/astarte_test.exs
```diff
@@ -91,106 +83,26 @@ defmodule Edgehog.AstarteTest do
...
-      cluster = cluster_fixture()
-      url = cluster.base_api_url
-
-      assert {:ok, ^cluster} = Astarte.fetch_or_create_cluster(url)
+      |> Enum.map(&Cluster.create/1)
+      |> Enum.each(fn result -> assert {:error, %Ash.Error.Invalid{}} = result end)
     end
 
-    test "update_cluster/2 with valid data updates the cluster" do
+    test "create/1 with existing base_api_url succeeds and updates the name" do
```
szakhlypa, line 102: Do we want to test `and updates the name` part in the same test?
```elixir
    test "create/1 with existing base_api_url succeeds and updates the name" do
      %Cluster{base_api_url: base_api_url} = cluster_fixture()

      assert {:ok, %Cluster{} = cluster} =
               Cluster.create(%{base_api_url: cluster.base_api_url, name: "some updated name"})

      assert cluster.base_api_url == base_api_url
      assert cluster.name == "some updated name"
    end
```
or to have two tests like
```elixir
    test "create/1 with existing base_api_url returns existing cluster" do
      cluster = cluster_fixture()

      assert {:ok, ^cluster} = Cluster.create(%{base_api_url: cluster.base_api_url})
    end

    test "create/1 with existing base_api_url updates the name" do
      %Cluster{id: id, base_api_url: base_api_url} = cluster_fixture()

      assert {:ok, %Cluster{} = cluster} =
               Cluster.create(%{base_api_url: base_api_url, name: "some updated name"})

      assert cluster.id == id
      assert cluster.base_api_url == base_api_url
      assert cluster.name == "some updated name"
    end
```

szakhlypa (changes requested):

backend/test/edgehog/astarte_test.exs
rbino: TBH yesterday I was re-evaluating the behaviour, I think I will not include the name in the upserted fields. The idea is that if a cluster already exists, then its name is not going to be updated when you try to re-create it.

szakhlypa (changes requested):

backend/test/edgehog/astarte/cluster_test.exs
```diff
@@ -0,0 +1,109 @@
...
+      assert cluster.name == name
+    end
+
+    test "strips trailing slash from base_api_url" do
+      attrs = %{base_api_url: "https://api.test.astarte.example/foo/", name: "test-trailing"}
+
+      assert {:ok, %Cluster{} = cluster} = Cluster.create(attrs)
+      assert cluster.base_api_url == "https://api.test.astarte.example/foo"
+    end
+
```
szakhlypa, line 48: Do we want to test that `name` is optional?
```elixir
    test "creates cluster without name" do
      %{base_api_url: url} = @valid_attrs

      assert {:ok, %Cluster{} = cluster} = Cluster.create(%{base_api_url: url})
      assert cluster.base_api_url == url
      assert cluster.name == nil
    end
```

szakhlypa (approved):

szakhlypa (approved):

PR #424, opened 2023-12-18
Port Astarte.Realm resource to Ash

rbino:
Since it's the first multitenant resource, also initialize multitenancy


backend/test/edgehog/astarte_test.exs
```diff
@@ -91,192 +83,122 @@ defmodule Edgehog.AstarteTest do
...
-    test "delete_cluster/1 deletes the cluster" do
-      cluster = cluster_fixture()
-      assert {:ok, %Cluster{}} = Astarte.delete_cluster(cluster)
-      assert_raise Ecto.NoResultsError, fn -> Astarte.get_cluster!(cluster.id) end
+      assert {:error, %Ash.Error.Invalid{errors: [error]}} = Realm.create(attrs, tenant: tenant)
+      assert %{field: :name, message: "has already been taken"} = error
     end
 
-    test "change_cluster/1 returns a cluster changeset" do
+    test "create with a duplicate name in another tenant returns error" do
```
szakhlypa, line 154: We can create realm with a duplicate name in another tenant:
```elixir
    test "create with a duplicate name in another tenant creates a realm" do
      other_tenant = tenant_fixture()
      other_cluster = cluster_fixture()
      realm = realm_fixture(tenant: other_tenant, cluster_id: other_cluster.id)

      tenant = tenant_fixture()
      cluster = cluster_fixture()
      attrs = %{cluster_id: cluster.id, name: realm.name, private_key: @valid_private_key}

      assert {:ok, %Realm{} = _realm} = Realm.create(attrs, tenant: tenant)
    end
```

But we can not create realm with a duplicate name in the same cluster:
```elixir
    test "create with a duplicate name in the same cluster returns error" do
      other_tenant = tenant_fixture()
      cluster = cluster_fixture()
      realm = realm_fixture(tenant: other_tenant, cluster_id: cluster.id)

      tenant = tenant_fixture()
      attrs = %{cluster_id: cluster.id, name: realm.name, private_key: @valid_private_key}

      assert {:error, %Ash.Error.Invalid{errors: [error]}} = Realm.create(attrs, tenant: tenant)
      assert %{field: :name, message: "has already been taken"} = error
    end
```

szakhlypa (changes requested):

backend/lib/edgehog/astarte/realm/realm.ex
```diff
@@ -0,0 +1,77 @@
...
+#
+
+defmodule Edgehog.Astarte.Realm do
+  use Edgehog.MultitenantResource
+
+  alias Edgehog.Validations
+
+  code_interface do
+    define_for Edgehog.Astarte
+    define :fetch_by_name, action: :by_name, args: [:name]
```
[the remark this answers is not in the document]
rbino: Yep, I noticed this yesterday working on the Provisioning context. I'd go with the `fetch` prefix which better expresses the tuple return value (and Ash also conveniently generates `fetch_by_name!` which raises as you'd expect).

backend/test/edgehog/astarte_test.exs
rbino: Is this the intended behaviour?

szakhlypa: To be able to create two realms with the same name for different tenants and clusters is the intended behaviour. Example: realm name is `test` (default one in Astarte in 5 minutes), tenants: `foo` (cluster `https://api.astarte.foo.com`) and `bar` (cluster `https://api.astarte.bar.com`).

I expect that to NOT be able to create two realms with the same name pointing on SAME cluster is also the intended behaviour.
First of all, if we allow to use same Astarte Realm for few Edgehog Tenants there may be some inconsistency problems.
Example: Edgehog installs Astarte trigger to handle device connection event (trigger name `edgehog-connection`), so having two Edgehog Tenants using same Astarte Realm we will have provisioning of conflicting triggers (same name, different action). We can resolve this adding tenant's slug into trigger name, but: when Edgehog receives data from Astarte for unknown device it fetches device data from Astarte and creates it in local. Having two triggers on same Astarte Realm will populate two Edgehog Tenants with same devices. I don't think that this is the intended behaviour.

davidebriani (approved):

szakhlypa (approved):

PR #425, opened 2023-12-18
Restore Tenant reconciliation

rbino:
Expose `:realm_management_client` as calculation from the Realm resource and restructure Interface and Trigger modules to implement an access pattern similar to Ash resources.


backend/lib/edgehog/astarte/realm/calculations/realm_management_client.ex
```diff
@@ -18,24 +18,29 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-defmodule Edgehog.Mocks.Astarte.Realm.Triggers do
-  @behaviour Edgehog.Astarte.Realm.Triggers.Behaviour
-
-  import Edgehog.AstarteFixtures
-  alias Astarte.Client.RealmManagement
+defmodule Edgehog.Astarte.Realm.Calculations.RealmManagementClient do
```
rbino, line 21: This is clearly not a renamed file, `git` go home you're drunk

backend/lib/edgehog/astarte/realm/realm.ex
```diff
@@ -64,6 +65,15 @@ defmodule Edgehog.Astarte.Realm do
     belongs_to :cluster, Edgehog.Astarte.Cluster
   end
 
+  calculations do
+    calculate :realm_management_client, :struct, Realm.Calculations.RealmManagementClient do
+      allow_nil? false
```
szakhlypa, line 70: We declare here that the calculation cannot return `nil`, however, in the implementation it can. Is this intended?

backend/lib/edgehog/tenants/reconciler.ex
```diff
@@ -66,8 +66,8 @@ defmodule Edgehog.Tenants.Reconciler do
       schedule_reconciliation(@reconcile_interval)
     end
 
-    Tenants.list_tenants()
-    |> Tenants.preload_astarte_resources_for_tenant()
+    Tenant
+    |> Tenants.read(load: tenant_reconciliation_loads())
```
szakhlypa, line 70: Shoudn't it be `read!` to get list of tenants?
```suggestion
    |> Tenants.read!(load: tenant_reconciliation_loads())
```

szakhlypa (changes requested):

davidebriani (approved):

PR #427, opened 2024-01-10
add tests for HardwareTypesTable component

harisbikovic:
This PR adds tests for the HardwareTypesTable component.

### DIFF
--- a/frontend/src/components/HardwareTypesTable.test.tsx
+++ b/frontend/src/components/HardwareTypesTable.test.tsx
@@ -1,0 +1,131 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2024 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+
+  SPDX-License-Identifier: Apache-2.0
+*/
+
+import { it, expect } from "vitest";
+import { screen } from "@testing-library/react";
+import { renderWithProviders } from "setupTests";
+import { createMockEnvironment } from "relay-test-utils";
+
+import { graphql, useLazyLoadQuery } from "react-relay/hooks";
+import type { HardwareTypesTable_getHardwareTypes_Query } from "api/__generated__/HardwareTypesTable_getHardwareTypes_Query.graphql";
+
+import HardwareTypesTable from "./HardwareTypesTable";
+
+const GET_HARDWARE_TYPES_QUERY = graphql`
+  query HardwareTypesTable_getHardwareTypes_Query @relay_test_operation {
+    hardwareTypes {
+      ...HardwareTypesTable_HardwareTypesFragment
+    }
+  }
+`;
+
+const ComponentWithQuery = () => {
+  const { hardwareTypes } =
+    useLazyLoadQuery<HardwareTypesTable_getHardwareTypes_Query>(
+      GET_HARDWARE_TYPES_QUERY,
+      {},
+    );
+  return <HardwareTypesTable hardwareTypesRef={hardwareTypes} />;
+};
+
+type HardwareType = {
+  id: string;
+  handle: string;
+  name: string;
+  partNumbers: string[];
+};
+
+const renderComponent = (hardwareTypes: HardwareType[] = []) => {
+  const relayEnvironment = createMockEnvironment();
+  relayEnvironment.mock.queueOperationResolver((_operation) => ({
+    data: { hardwareTypes },
+  }));
+  renderWithProviders(<ComponentWithQuery />, { relayEnvironment });
+};
+
+it("renders column headers", () => {
+  renderComponent();
+
+  expect(
+    screen.getByRole("columnheader", { name: "Hardware Type Name" }),
+  ).toBeVisible();
+  expect(screen.getByRole("columnheader", { name: "Handle" })).toBeVisible();
+  expect(
+    screen.getByRole("columnheader", { name: "Part Numbers" }),
+  ).toBeVisible();
+});
+
+it("renders Hardware Type data", () => {
+  renderComponent([
+    {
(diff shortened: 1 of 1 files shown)


frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+  `;
+
+  const ComponentWithQuery = () => {
+    const { hardwareTypes } =
+      useLazyLoadQuery<HardwareTypesTable_getHardwareTypes_Query>(
+        GET_HARDWARE_TYPES_QUERY,
+        {},
+      );
+    return (
+      hardwareTypes && <HardwareTypesTable hardwareTypesRef={hardwareTypes} />
```
szakhlypa, line 50: Do we need to check on `hardwareTypes`?

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+  };
+  const SuspenseWithComponent = ({
+    fallback = "Loading...",
+  }: SuspenseWithComponentProps) => (
+    <Suspense fallback={fallback}>
+      <ComponentWithQuery />
+    </Suspense>
+  );
+
+  it("does suspension correctly", () => {
```
szakhlypa, line 65:
```suggestion
  it("suspends correctly", () => {
```

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+    renderComponent([hwType1, hwType2, hwType3]);
+
+    const table = await screen.findByRole("table");
+    expect(table).toBeInTheDocument();
+    const tableElements = within(table).getAllByRole("rowgroup");
+    expect(tableElements).toHaveLength(2);
+
+    const [head, body] = tableElements;
+
+    const columns = within(head).getAllByRole("columnheader");
```
szakhlypa, line 153: Why do we repeat the `columnheader` check in every rendering test? Could we create separate unit-test for it?

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+    expect(cells).toHaveLength(3);
+    expect(
+      within(cells[0]).getByRole("link", { name: hwType.name }),
+    ).toBeInTheDocument();
+    expect(cells[0]).toHaveTextContent(hwType.name);
+    expect(cells[1]).toHaveTextContent(hwType.handle);
+    expect(cells[2]).toHaveTextContent(hwType.partNumbers.join(", "));
+  });
+
+  it("renders table with multiple rows when data fetched contains multiple elements", async () => {
```
szakhlypa, line 123: What the reason to add this test? Rendering multiple rows should be tested on `Table.tsx` level.

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+    const row = rows[0];
+    const cells = within(row).getAllByRole("cell");
+    expect(cells).toHaveLength(3);
+    expect(
+      within(cells[0]).getByRole("link", { name: hwType.name }),
+    ).toBeInTheDocument();
+    expect(cells[0]).toHaveTextContent(hwType.name);
+    expect(cells[1]).toHaveTextContent(hwType.handle);
+    expect(cells[2]).toHaveTextContent(hwType.partNumbers.join(", "));
```
szakhlypa, line 120: We add a test to make sure that `HardwareType` with few `partNumbers`  has them rendered with comma separators, but there is no test for what happens when `HardwareType` has only one partNumber.

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+    const table = await screen.findByRole("table");
+    expect(table).toBeInTheDocument();
+    const tableElements = within(table).getAllByRole("rowgroup");
+    expect(tableElements).toHaveLength(2);
+
+    const [head, body] = tableElements;
+
+    const columns = within(head).getAllByRole("columnheader");
+    expect(columns).toHaveLength(3);
+    expect(columns[0]).toHaveTextContent("Hardware Type Name");
```
szakhlypa, line 105: I think it's better not to depend on the order of the columns and test them with 

```typescript
expect(screen.getByRole("columnheader", { name: "Hardware Type Name" })).toBeInTheDocument();
```
Same for the cells below.

Please let me know if you agree. Maybe we can have one test with the order check and all the others just checking for presence?

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+      handle: "HW-HANDLE3",
+      name: "HW name3",
+      partNumbers: ["HW-PN31", "HW-PN32"],
+    };
+    renderComponent([hwType1, hwType2, hwType3]);
+
+    const table = await screen.findByRole("table");
+    expect(table).toBeInTheDocument();
+    const tableElements = within(table).getAllByRole("rowgroup");
+    expect(tableElements).toHaveLength(2);
```
szakhlypa, line 149: Why do we repeat this code in every rendering test? Can we have one unit test for it?

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+    const rows = within(body).getAllByRole("row");
+    expect(rows).toHaveLength(1);
+
+    const row = rows[0];
+    const cells = within(row).getAllByRole("cell");
+    expect(cells).toHaveLength(3);
+    expect(
+      within(cells[0]).getByRole("link", { name: hwType.name }),
+    ).toBeInTheDocument();
```
szakhlypa, line 117: Do we want to test that link has correct href attribute?

frontend/src/components/HardwareTypesTable.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+    expect(cells[0]).toHaveTextContent(hwType.name);
```
szakhlypa, line 118: Expect seems redundant, you already tested on link's name just above.

szakhlypa (changes requested):

frontend/src/components/HardwareTypesTable.test.tsx
harisbikovic: Yes, I think that's a good idea.

szakhlypa (changes requested):

szakhlypa (approved):

PR #429, opened 2024-01-15
Restore Tenant Provisioning

rbino:
Move it to the `Tenants` API, removing the `Provisioning` context.
Use AshJsonApi to expose a new JsonApi compatible API.


backend/lib/edgehog/tenants/tenant/changes/provision_astarte_resources.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  alias Edgehog.Astarte
+  alias Edgehog.Provisioning.AstarteConfig
+  alias Edgehog.Provisioning.Tenant
+  alias Edgehog.Tenants
+
+  @impl true
+  def change(changeset, _opts, _ctx) do
+    Ash.Changeset.after_action(changeset, &provision_astarte_resources/2)
+  end
```
davidebriani, line 32: Does this run within a transaction with respect to the action?
I wonder whether it could be that tenant creation succeeds while provisioning the Astarte resources fails.

backend/lib/edgehog/tenants/tenant/manual_actions/reconciler_action.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  def run(input, opts, _context) do
+    tenant = input.arguments.tenant
+
+    result =
+      case input.action.name do
+        :reconcile -> @reconciler_module.reconcile_tenant(tenant)
+      end
+
+    {:ok, result}
```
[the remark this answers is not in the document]
rbino: Yes, `run` must return an `:ok/:error` tuple as can be seen by the callback
https://hexdocs.pm/ash/Ash.Resource.Actions.Implementation.html#callbacks
(note to self: add the behaviour because it's missing)

backend/test/edgehog_web/admin_api/tenants/tenant_test.exs
```diff
@@ -0,0 +1,139 @@
...
+      conn =
+        post(conn, path, params)
+
+      assert %{"errors" => errors} = json_response(conn, 400)
+
+      required_data = ["base_api_url", "realm_name", "realm_private_key"]
+
+      for required <- required_data do
+        assert Enum.find(errors, &(&1["detail"] =~ "#{required} is required"))
+      end
```
davidebriani, line 136: Besides `detail`, is a `path` field also available to check that the error is related to the `astarte_config`?

davidebriani (changes requested):

backend/lib/edgehog/tenants/tenant/changes/provision_astarte_resources.ex
rbino: Yes `after/before_action` are executed inside the transaction, this is shown [here](https://hexdocs.pm/ash/actions.html#action-lifecycle). If the `after_action` fails, the transaction is rolled back.

backend/lib/edgehog/tenants/tenant/tenant.ex
```diff
@@ -22,12 +22,14 @@ defmodule Edgehog.Tenants.Tenant do
...
     data_layer: AshPostgres.DataLayer
 
+  alias Edgehog.Tenants.AstarteConfig
   alias Edgehog.Tenants.Tenant
   alias Edgehog.Validations
 
   code_interface do
     define_for Edgehog.Tenants
     define :create
+    define :provision
```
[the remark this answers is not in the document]
rbino: I've left it out so the call is `Tenant.provision(%{name: ..., astarte_config: ...})`. If you specify `arg` here, the mentioned argument gets promoted to a positional argument, so `Tenant.provision(astarte_config, %{name: ...,})`. Given `astarte_config` is no more special than the other params, I've decided to just leave it inside the params map.

backend/test/edgehog_web/admin_api/tenants/tenant_test.exs
rbino: This actually gets printed for attributes
2024-01-19-122746_948x328_scrot
But not for arguments inside `astarte_config`
2024-01-19-122820_884x284_scrot
I'll investigate further but I don't think it's a blocker currently

davidebriani (approved):

szakhlypa (changes requested):

backend/lib/edgehog/tenants/tenant/changes/trigger_reconciliation.ex
```diff
@@ -19,13 +18,17 @@
...
-defmodule EdgehogWeb.AdminAPI.FallbackController do
-  use EdgehogWeb, :controller
+defmodule Edgehog.Tenants.Tenant.Changes.TriggerReconciliation do
+  use Ash.Resource.Change
+
+  alias Edgehog.Tenants
+
+  @impl true
+  def change(changeset, _opts, _ctx) do
+    Ash.Changeset.after_action(changeset, fn _changeset, tenant ->
```
szakhlypa, line 28: Do we need to specify that this action should not abort the transaction?
```elixir
    Ash.Changeset.after_transaction(changeset, fn 
      _changeset, {:ok, tenant} -> 
        Tenants.Tenant.reconcile(tenant)
        
        {:ok, tenant}

      _changeset, {:error, error} ->
        
        {:error, error}
    end)
```

backend/test/edgehog_web/admin_api/tenants/tenant_test.exs
```diff
@@ -33,14 +35,20 @@ defmodule EdgehogWeb.AdminAPI.TenantsControllerTest do
...
+                         |> X509.PrivateKey.to_pem()
+                         |> String.trim()
 
   describe "POST /admin-api/v1/tenants" do
     setup %{conn: conn} do
-      path = Routes.tenants_path(conn, :create)
+      conn =
+        conn
+        |> put_req_header("content-type", "application/vnd.api+json")
+        |> put_req_header("accept", "application/vnd.api+json")
```
szakhlypa, line 49: Should we add headers in `EdgehogWeb.AdminAPI.ConnCase`?

backend/lib/edgehog/tenants/astarte_config.ex
```diff
@@ -18,16 +18,19 @@
...
 #
 
-defmodule EdgehogWeb.AdminAPI.TenantsController do
-  use EdgehogWeb, :controller
+defmodule Edgehog.Tenants.AstarteConfig do
+  use Ash.Resource,
+    data_layer: :embedded
 
-  alias Edgehog.Provisioning
+  alias Edgehog.Validations
```
szakhlypa, line 25:
```suggestion
```

szakhlypa:
`Edgehog.Tenants.Reconciler.Behaviour` [expects](https://github.com/edgehog-device-manager/edgehog/blob/from-ash-we-rise/backend/lib/edgehog/tenants/reconciler/behaviour.ex#L24) `Edgehog.Tenants.Tenant` to have a typespec `@type t()`.
Ash.Resource has [`@type record :: struct()`](https://github.com/ash-project/ash/blob/main/lib/ash/resource.ex#L9C3-L9C27).

I think we can start with `@type record :: Ash.Resource.record()` for `Tenant` and `Tenant.record()` in `Reconciler.Behaviour`

backend/lib/edgehog/tenants/tenant/changes/provision_astarte_resources.ex
```diff
@@ -0,0 +1,89 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Tenants.Tenant.Changes.ProvisionAstarteResources do
+  use Ash.Resource.Change
+
+  alias Edgehog.Astarte
+  alias Edgehog.Provisioning.AstarteConfig
+  alias Edgehog.Provisioning.Tenant
+  alias Edgehog.Tenants
```
szakhlypa, line 27: Unused aliases that you may have wanted to use for pattern matching.

backend/test/edgehog/tenants_test.exs
szakhlypa, line 1: If module `Edgehog.Tenants` has `lib/edgehog/tenants/tenants.ex` path, should also `Edgehog.TenantsTest` be   in `test/edgehog/tenants/tenants_test.exs` ?

backend/lib/edgehog/tenants/tenant/changes/trigger_reconciliation.ex
rbino: `after_action` doesn't get executed if the action is not successful

rbino: Ok I see you changed to `after_transaction`, but the original action can't fail, so I guess using `after_action` is the same (and it's more compact because you don't have to handle the error case).

Moreover, in the future we could start leveraging the Notification mechanism of Ash since I think this fits nicely that pattern (I'll add a TODO comment)

backend/test/edgehog/tenants_test.exs
rbino: Imho that's a different convention because `tenants` != `tenants_test` (and the `_test` suffix is required by ExUnit). I'd leave the tests as they are for now.

backend/test/edgehog_web/admin_api/auth_test.exs
```diff
@@ -22,32 +22,45 @@ defmodule EdgehogWeb.AdminAPI.AuthTest do
...
   }
 
   alias Edgehog.Config
 
   setup %{conn: conn} do
-    path = Routes.tenants_path(conn, :create)
+    conn =
+      conn
+      |> put_req_header("content-type", "application/vnd.api+json")
+      |> put_req_header("accept", "application/vnd.api+json")
```
szakhlypa, line 61: No need to add headers (anymore), they are now added in the `EdgehogWeb.AdminAPI.ConnCase`.

backend/lib/edgehog/tenants/tenant/manual_actions/reconciler_action.ex
```diff
@@ -0,0 +1,36 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Tenants.Tenant.ManualActions.ReconcilerAction do
+  alias Edgehog.Tenants.Tenant.Reconciler
```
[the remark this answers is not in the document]
szakhlypa: There is still an alias for `Edgehog.Tenants.Tenant.Reconciler` (module does not exists), we need to use `alias Edgehog.Tenants.Reconciler`.

szakhlypa (changes requested):

szakhlypa (approved):

PR #433, opened 2024-01-23
Port Hardware Type to Ash

rbino:
Add all the relevant tests in the Graphql part and remove unused old code


davidebriani (approved):

backend/lib/edgehog/devices/hardware_type.ex
```diff
@@ -19,39 +19,135 @@
...
+    hide_fields [:tenant, :part_number_strings]
+
+    queries do
+      get :hardware_type, :get
+      list :hardware_types, :list
+    end
+
+    mutations do
+      create :create_hardware_type, :create
+      update :update_hardware_type, :update
```
szakhlypa, line 41:
```graphql
  updateHardwareType(
    id: ID
    input: UpdateHardwareTypeInput
  ): UpdateHardwareTypeResult
```
Same here, `ID` and `input` should be non-nullable

backend/lib/edgehog/devices/hardware_type.ex
```diff
@@ -19,39 +19,135 @@
...
+      destroy :delete_hardware_type, :destroy
```
szakhlypa, line 42:
```graphql
deleteHardwareType(id: ID): DeleteHardwareTypeResult
```
id should be non-nullable

backend/test/edgehog_web/schema/mutation/delete_hardware_type_test.exs
```diff
@@ -19,75 +19,89 @@
...
+      assert %{"fields" => ["id"], "message" => "could not be found"} = extract_error!(result)
+    end
+  end
+
+  defp delete_hardware_type_mutation(opts) do
+    default_document = """
+    mutation DeleteHardwareType($id: ID!) {
+      deleteHardwareType(id: $id) {
+        result {
           id
```
szakhlypa, line 54: Querying `partNumbers` of deleted HardwareType (add next fields after `id`)
```graphql
          partNumbers {
            partNumber
          }
```
returns `null` in `result` with 
```elixir
errors: [
    %{
      message: "Cannot return null for non-nullable field",
      path: ["deleteHardwareType", "result", "partNumbers", 0, "partNumber"],
      locations: [%{line: 6, column: 9}]
    }
  ]
```

backend/test/edgehog_web/schema/mutation/update_hardware_type_test.exs
```diff
@@ -19,173 +19,235 @@
...
-                 }
-               }
-             } = assert(json_response(conn, 200))
+      assert %{"fields" => ["handle"], "message" => "should only contain" <> _} =
+               extract_error!(result)
+    end
 
-      assert {:ok, %HardwareType{name: ^initial_name, handle: ^initial_handle}} =
-               Devices.fetch_hardware_type(hardware_type.id)
+    test "returns error for empty part_numbers", %{tenant: tenant, hardware_type: hardware_type} do
```
szakhlypa, line 117: We can also check for invalid `part_number` like
```elixir
    test "returns error for invalid part_number", %{tenant: tenant, hardware_type: hardware_type} do
      result =
        update_hardware_type_mutation(
          tenant: tenant,
          id: hardware_type.id,
          part_numbers: [""]
        )

      assert %{"fields" => ["part_number"], "message" => "is required"} =
               extract_error!(result)
    end
```
But we have no `partNumber` input field, not sure how to return this error better.

backend/lib/edgehog/devices/hardware_type_part_number.ex
```diff
@@ -19,24 +19,47 @@
...
-  use Ecto.Schema
-  import Ecto.Changeset
+  use Edgehog.MultitenantResource,
+    extensions: [
+      AshGraphql.Resource
+    ]
 
-  alias Edgehog.Devices.HardwareType
+  graphql do
+    type :hardware_type_part_number
```
szakhlypa, line 28:
```graphql
type HardwareTypePartNumber {
  id: ID!

  "The part number identifier."
  partNumber: String!

  hardwareType: HardwareType
}
```
Should `hardwareType` field be non-nullable?

backend/test/edgehog_web/schema/query/hardware_types_test.exs
```diff
@@ -28,44 +30,94 @@ defmodule EdgehogWeb.Schema.Query.HardwareTypesTest do
...
+
+  defp hardware_types_query(opts) do
+    default_document =
+      """
+      query HardwareTypes($filter: HardwareTypeFilterInput, $sort: [HardwareTypeSortInput]) {
+        hardwareTypes(filter: $filter, sort: $sort) {
+          name
+          handle
+          partNumbers {
+            partNumber
```
szakhlypa, line 103: Having option to query `hardwareType` field on `HardwareTypePartNumber` type, do we want to test nested queries? like
```graphql
          partNumbers {
            partNumber
            hardwareType {
              name
              handle
              partNumbers {
                partNumber
                hardwareType {
                  name
                  handle
                }
              }
            }
          }
```

szakhlypa (changes requested):

backend/lib/edgehog/devices/hardware_type.ex
rbino: See above

backend/test/edgehog_web/schema/mutation/delete_hardware_type_test.exs
rbino: Minor, move to `ash-porting` issue

backend/test/edgehog_web/schema/mutation/update_hardware_type_test.exs
rbino: Minor, move to `ash-porting` issue

backend/lib/edgehog/devices/hardware_type_part_number.ex
rbino: See above

backend/test/edgehog_web/schema/query/hardware_types_test.exs
rbino: We'll re-evaluate filters and sorts at the end of the porting process, please open a generic issue about this with the `ash-porting` label so that we have an item tracking this

backend/lib/edgehog/devices/hardware_type_part_number.ex
rbino: I've checked and currently `hardware_type_id` is actually nullable at the DB level (see [here](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/priv/repo/migrations/20211108144240_create_hardware_type_part_numbers.exs#L31)), so this is coherent with the status-quo.
The goal of the first step of the migration is having the resources reflect the current state of the database, then we can start adjusting the constraint with subsequent migrations.

szakhlypa (approved):

Issue #441, opened 2024-02-08
Re-enable localized descriptions of System Models


rbino:
Since we're breaking the API during the porting to Ash, this could be a possibility to revisit how they're handled since we practically never supported locales other than the default tenant locale

davidebriani:
Closing as implemented by #524

PR #442, opened 2024-02-08
Port System Model to Ash

rbino:
Note that this is still missing descriptions (and their localization), but since they're not critical they will be handled in a second round (see #441).


backend/lib/edgehog/devices/system_model.ex
```diff
@@ -19,39 +19,159 @@
...
+    end
+
+    create :create do
+      description "Creates a system model."
+      primary? true
+
+      argument :hardware_type_id, Edgehog.Types.RelayId do
+        description "The ID of the hardware type that can be used by devices of this model"
+        allow_nil? false
+      end
```
davidebriani: I have a couple of doubts here:
1. What is the purpose of `Edgehog.Types.RelayId`?
2. Why using something related to Relay in a generic `action`?

davidebriani (changes requested):

backend/lib/edgehog/devices/system_model.ex
rbino: When I started working on this https://github.com/ash-project/ash_graphql/pull/109 was not merged so the `RelayId` type was also handling the translation between Relay and database IDs as a temporary workaround. 

Meanwhile that got merged, so in reality it makes no sense to call that `RelayId` so I'll change it to `Id`. We still need that because Ash kind of assumes that you use UUIDs as IDs, so the `:uuid` type gets exposed as `:id` at the GraphQL level, while the `:integer` type (which is what we use as ID) does not, the custom type basically just handles that.

davidebriani (approved):

backend/test/support/fixtures/devices_fixtures.ex
```diff
@@ -117,18 +117,25 @@ defmodule Edgehog.DevicesFixtures do
   Generate a system_model.
   """
   def system_model_fixture(opts \\ []) do
```
szakhlypa, line 119:
```suggestion
  def system_model_fixture(opts) do
```
We shouldn't define `[]` as the default value for `opts`, `opts` must be a non-empty keyword list with (at least) the `:tenant` key.

backend/lib/edgehog/devices/system_model/changes/handle_picture_upload.ex
```diff
@@ -0,0 +1,69 @@
...
+        Ash.Changeset.add_error(changeset, field: :picture_file, message: "failed to upload")
+    end
+  end
+
+  # If we've uploaded the picture and the transaction resulted in an error, we do our
+  # best to clean up
+  defp maybe_cleanup(changeset, {:error, _} = result, picture_file) do
+    {:ok, system_model} = Ash.Changeset.apply_attributes(changeset)
+
+    _ = Assets.delete_system_model_picture(system_model, picture_file)
```
szakhlypa, line 63: Shouldn't we use `picture_url` to delete the uploaded picture?

szakhlypa (changes requested):

backend/test/support/fixtures/devices_fixtures.ex
rbino: The problem (and reason I'm providing `opts` with a default value) is that if you do that you get:
```
error: undefined function system_model_fixture/0 (expected Edgehog.BaseImagesTest to define such a function or for it to be imported, but none are available)
  test/edgehog/base_images_test.exs:35: Edgehog.BaseImagesTest.__ex_unit_setup_0_0/1
[...]
```
and other errors, because unfortunately even if tests are excluded with `--only`, they still all get compiled and an undefined function is a compilation error, and I don't want to fix all calling code in tests that are probably going to be removed anyway.

I plan on removing all default opts at the end of the porting.

backend/lib/edgehog/devices/system_model/changes/handle_picture_upload.ex
rbino: Fixed, and since this was not caught by tests I've added two test cases covering this (i.e. image cleanup for failed create/update)

backend/lib/edgehog/devices/system_model/changes/handle_picture_upload.ex
```diff
@@ -0,0 +1,69 @@
...
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Devices.SystemModel.Changes.HandlePictureUpload do
+  use Ash.Resource.Change
+
+  alias Edgehog.Assets
+  alias Edgehog.Devices.SystemModel
```
szakhlypa, line 25: unused alias
```suggestion
```

backend/test/edgehog_web/schema/mutation/update_system_model_test.exs
```diff
@@ -19,215 +19,378 @@
...
+        update_system_model_mutation(
+          tenant: tenant,
+          id: id,
+          handle: duplicate.handle,
+          picture_file: %Plug.Upload{path: "/tmp/image.jpg", filename: "image.jpg"}
+        )
+
+      assert extract_error!(result)
+    end
+
```
szakhlypa, line 203: Maybe I'm missing something, I've tried to test the error response in case of `picture_file` uploading error with code below:
(also added expectation that `:delete` should not be invoked in this case)
```elixir
    test "returns error when fails to update picture_file", %{tenant: tenant, id: id} do
      Edgehog.Assets.SystemModelPictureMock
      |> expect(:upload, fn _, _ -> {:error, :no_space_left} end)
      |> expect(:delete, 0, fn _, _ -> :ok end)

      result =
        update_system_model_mutation(
          tenant: tenant,
          id: id,
          picture_file: %Plug.Upload{path: "/tmp/image.jpg", filename: "image.jpg"}
        )

      assert %{
        fields: [:picture_file],
        message: "failed to upload"
      } = extract_error!(result)
    end
```
But got in response `something_went_wrong`
```
%{
  fields: [],
  message: "Something went wrong. Unique error id: `0ad07194-8ee3-4d26-b5e1-8784e600e234`",
  code: "something_went_wrong",
  locations: [%{line: 2, column: 3}],
  path: ["updateSystemModel"],
  short_message: "Something went wrong.",
  vars: %{}
}
```

szakhlypa (changes requested):

backend/test/edgehog_web/schema/mutation/update_system_model_test.exs
rbino: This was actually a bug, the cleanup was tried even if the upload failure was the case of the failure and crashed on `apply_attributes` in `maybe_cleanup`.

I've refactored the code to only cleanup in case of successful uploads and added your test in the testsuite.

szakhlypa (approved):

PR #443, opened 2024-02-09
Refactor update device group's form info

harisbikovic:
This PR creates GraphQL fragment for data into UpdateDeviceGroup form and uses it in the DeviceGroup page.
This PR also gets rid of relay/unused-fields ESlint warning for the DeviceGroup page.

### DIFF
--- a/frontend/src/forms/UpdateDeviceGroup.tsx
+++ b/frontend/src/forms/UpdateDeviceGroup.tsx
@@ -1,7 +1,7 @@
 /*
   This file is part of Edgehog.
 
-  Copyright 2022 SECO Mind Srl
+  Copyright 2022-2024 SECO Mind Srl
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -30,6 +30,16 @@ import Row from "components/Row";
 import Spinner from "components/Spinner";
 import Stack from "components/Stack";
 import { deviceGroupHandleSchema, yup } from "forms";
+import { graphql, useFragment } from "react-relay/hooks";
+import type { UpdateDeviceGroup_DeviceGroupFragment$key } from "api/__generated__/UpdateDeviceGroup_DeviceGroupFragment.graphql";
+
+const UPDATE_DEVICE_GROUP_FRAGMENT = graphql`
+  fragment UpdateDeviceGroup_DeviceGroupFragment on DeviceGroup {
+    name
+    handle
+    selector
+  }
+`;
 
 const FormRow = ({
   id,
@@ -63,25 +73,30 @@ const deviceGroupSchema = yup
   .required();
 
 type Props = {
-  initialData: DeviceGroupData;
+  deviceGroupRef: UpdateDeviceGroup_DeviceGroupFragment$key;
   isLoading?: boolean;
   onSubmit: (data: DeviceGroupData) => void;
   onDelete: () => void;
 };
 
 const UpdateDeviceGroupForm = ({
-  initialData,
+  deviceGroupRef,
   isLoading = false,
   onSubmit,
   onDelete,
 }: Props) => {
+  const deviceGroupData = useFragment(
+    UPDATE_DEVICE_GROUP_FRAGMENT,
+    deviceGroupRef,
+  );
+
   const {
     register,
     handleSubmit,
     formState: { errors },
   } = useForm<DeviceGroupData>({
     mode: "onTouched",
-    defaultValues: initialData,
+    defaultValues: deviceGroupData,
     resolver: yupResolver(deviceGroupSchema),
   });
 
--- a/frontend/src/pages/DeviceGroup.tsx
+++ b/frontend/src/pages/DeviceGroup.tsx
@@ -1,7 +1,7 @@
 /*
   This file is part of Edgehog.
 
-  Copyright 2022-2023 SECO Mind Srl
+  Copyright 2022-2024 SECO Mind Srl
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -18,7 +18,7 @@
   SPDX-License-Identifier: Apache-2.0
 */
 
-import { Suspense, useCallback, useEffect, useMemo, useState } from "react";
+import { Suspense, useCallback, useEffect, useState } from "react";
 import { useParams } from "react-router-dom";
 import { ErrorBoundary } from "react-error-boundary";
 import {
@@ -29,7 +29,6 @@ import {
 } from "react-relay/hooks";
 import type { PreloadedQuery } from "react-relay/hooks";
 import { FormattedMessage } from "react-intl";
-import _ from "lodash";
 
 import type { DeviceGroup_getDeviceGroup_Query } from "api/__generated__/DeviceGroup_getDeviceGroup_Query.graphql";
 import type { DeviceGroup_updateDeviceGroup_Mutation } from "api/__generated__/DeviceGroup_updateDeviceGroup_Mutation.graphql";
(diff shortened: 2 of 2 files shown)


frontend/src/pages/DeviceGroup.tsx
```diff
@@ -45,16 +47,24 @@ import Spinner from "components/Spinner";
 import UpdateDeviceGroupForm from "forms/UpdateDeviceGroup";
 import type { DeviceGroupData } from "forms/UpdateDeviceGroup";
 
+// We use graphql fields below in columns configuration
+/* eslint-disable relay/unused-fields */
+const DEVICE_GROUP_TABLE_FRAGMENT = graphql`
+  fragment DeviceGroup_DeviceGroupFragment on DeviceGroup {
```
szakhlypa, line 53: How does this fragment resolve the `relay/unused-fields` warning?

[Quote](https://relay.dev/docs/principles-and-architecture/thinking-in-relay) from the Relay documentation:
 `components use one or more GraphQL fragments to describe their data requirements. These fragments are then nested within other fragments, and ultimately within queries`.

My assumption is that the form component should have a relay fragment and that we should spread it into a query on the current page.

szakhlypa (changes requested):

frontend/src/pages/DeviceGroup.tsx
```diff
@@ -291,7 +288,7 @@ const DeviceGroupContent = ({
 
   return (
     <Page>
-      <Page.Header title={deviceGroup.name} />
+      <Page.Header title={deviceGroupData.name} />
```
szakhlypa, line 291: If the `name` field  of `DeviceGroup` GraphQL type has been removed from `GET_DEVICE_GROUP_QUERY`, where do we get the Device Group `name` to render in the header?
Unless you're planning to create a separate component for the Device Group header with its own fragment, I think we should fetch for `name` in the query (as we use query result in `DeviceGroupContent`).

szakhlypa (changes requested):

frontend/src/pages/DeviceGroup.tsx
```diff
@@ -303,16 +300,20 @@ const DeviceGroupContent = ({
         </Alert>
         <div className="mb-3">
           <UpdateDeviceGroupForm
-            initialData={_.pick(deviceGroup, ["name", "handle", "selector"])}
+            deviceGroupRef={_.pick(deviceGroupData, [
```
szakhlypa, line 303: Why do we need to pick data here? The `UpdateDeviceGroupForm` expects [fragment reference](https://relay.dev/docs/api-reference/use-fragment/#arguments) in the `deviceGroupRef` property

szakhlypa (changes requested):

frontend/src/pages/DeviceGroup.tsx
harisbikovic: Is it the same for the `DeleteModal` where we need to pass `handle`, therefore we need to read it in the query too?

frontend/src/pages/DeviceGroup.tsx
```diff
@@ -269,7 +266,7 @@ const DeviceGroupContent = ({
     [updateDeviceGroup, deviceGroupId],
   );
 
-  if (!deviceGroup) {
+  if (!deviceGroupData) {
```
szakhlypa, line 269: `deviceGroupData` is always `true`, do you want to check on `deviceGroupData.deviceGroup`?

szakhlypa (changes requested):

frontend/src/pages/DeviceGroup.tsx
szakhlypa: Yes, any Device Group data in `DeviceGroupContent` that you want to use/to pass by value should be fetched in the query.

Also with the current query and `const deviceGroupData = usePreloadedQuery(...`, the Device Group fields can be found in `deviceGroupData.deviceGroup?.<fieldName>` (e.g.  `name` -> `deviceGroupData.deviceGroup?.name`). We can change it to `const { deviceGroup } = usePreloadedQuery(...` since we are not using `deviceGroupData` directly.

frontend/src/pages/DeviceGroup.tsx
```diff
@@ -185,10 +183,10 @@ const DeviceGroupContent = ({
...
 
   // TODO: handle readonly type without mapping to mutable type
-  const deviceGroup = useMemo(
-    () => deviceGroupData.deviceGroup && { ...deviceGroupData.deviceGroup },
-    [deviceGroupData.deviceGroup],
-  );
+  //const deviceGroup = useMemo(
+  //  () => deviceGroupData.deviceGroup && { ...deviceGroupData.deviceGroup },
+  //  [deviceGroupData.deviceGroup],
+  //);
```
szakhlypa, line 189: Don't you want to delete these lines?

frontend/src/pages/DeviceGroup.tsx
```diff
@@ -303,16 +301,16 @@ const DeviceGroupContent = ({
...
           <UpdateDeviceGroupForm
-            initialData={_.pick(deviceGroup, ["name", "handle", "selector"])}
+            deviceGroupRef={deviceGroup}
             onSubmit={handleUpdateDeviceGroup}
             onDelete={handleShowDeleteModal}
             isLoading={isUpdatingDeviceGroup}
           />
         </div>
-        <DevicesTable devicesRef={deviceGroup.devices} hideSearch />
+        <DevicesTable devicesRef={deviceGroup.devices || []} hideSearch />
```
szakhlypa, line 310: `devices` is not nullable field, why do we need an empty array here?

frontend/src/pages/DeviceGroup.tsx
```diff
@@ -303,16 +301,16 @@ const DeviceGroupContent = ({
...
         {showDeleteModal && (
           <DeleteModal
-            confirmText={deviceGroup.handle}
+            confirmText={deviceGroup.handle || ""}
```
szakhlypa, line 313: `handle` is not nullable field, why do we need an empty string here?

frontend/src/forms/UpdateDeviceGroup.tsx
```diff
@@ -48,11 +61,7 @@ const FormRow = ({
   </Form.Group>
 );
 
-type DeviceGroupData = {
-  name: string;
-  handle: string;
-  selector: string;
-};
+type DeviceGroupData = UpdateDeviceGroup_DeviceGroupFragment$data;
```
szakhlypa, line 64: `UpdateDeviceGroup_DeviceGroupFragment$data` has ` $fragmentType` property, shouldn't it be omitted in `DeviceGroupData` type? I don't think we need fragment property in `useForm` and `onSubmit`.

frontend/src/pages/DeviceGroup.tsx
szakhlypa, line 1: I think we should also update the `UPDATE_DEVICE_GROUP_MUTATION` to use `UpdateDeviceGroup_DeviceGroupFragment` in the selection set.

szakhlypa (changes requested):

frontend/src/pages/DeviceGroup.tsx
```diff
@@ -65,10 +63,7 @@ const UPDATE_DEVICE_GROUP_MUTATION = graphql`
   ) {
     updateDeviceGroup(input: $input) {
       deviceGroup {
-        id
-        name
-        handle
-        selector
```
szakhlypa, line 71: Why do we remove `name` and `handle` fields from the selection when we use them on the current page?

frontend/src/forms/UpdateDeviceGroup.tsx
```diff
@@ -30,6 +30,17 @@ import Row from "components/Row";
 import Spinner from "components/Spinner";
 import Stack from "components/Stack";
 import { deviceGroupHandleSchema, yup } from "forms";
+import { graphql, useFragment } from "react-relay/hooks";
+import type { UpdateDeviceGroup_DeviceGroupFragment$key } from "api/__generated__/UpdateDeviceGroup_DeviceGroupFragment.graphql";
+
+const UPDATE_DEVICE_GROUP_FRAGMENT = graphql`
+  fragment UpdateDeviceGroup_DeviceGroupFragment on DeviceGroup {
+    id
```
szakhlypa, line 38: How do you use `id` field of this fragment?

szakhlypa (changes requested):

szakhlypa (approved):

PR #450, opened 2024-02-20
Add option to run Edgehog without local instance of Astarte

szakhlypa:
The default docker-compose.yml assumes that the user is running Edgehog along with a local instance of Astarte.
For development purposes, it's also possible to run an instance of Edgehog without the Astarte instance with:
```
docker compose \
  -f docker-compose.yml \
  -f docker-compose.without-astarte.yml \
  up -d
```


rbino (changes requested):
Let's document the newly added command somewhere (README?)

rbino (approved):

Issue #446, opened 2024-02-13
Supports a GraphQL query to retrieve the status of a Forwarder session for a device


davidebriani:
As a complement to issue #445, API clients may need to retrieve the status of a Forwarder session they requested before attempting to access it.

The Edgehog backend can support a `forwarderSessionStatus` GraphQL query.
A client would use the query to inspect the status of a device → Forwarder session, and know whether the session is ready to be accessed and it is closed.

```graphql
query {
  forwarderSessionStatus(deviceId: ID!, sessionToken: String!) {
    sessionStatus: "CONNECTED" | "CONNECTING" | "DISCONNECTED"
  }
}
```

To support the query, the backend queries the Astarte interface [`ForwarderSessionsState`](https://github.com/edgehog-device-manager/edgehog-astarte-interfaces/blob/main/io.edgehog.devicemanager.ForwarderSessionsState.json) for the specified device and returns the status of the session identified by the sessionToken.

PR #455, opened 2024-02-22
Add GraphQL APIs to request forwarder sessions

### LINKED ISSUE #445: Supports a GraphQL mutation to request a Forwarder session for a device
API clients should be able to request a device → Forwarder session to communicate with a specific device, for example for accessing a remote terminal session.

The Edgehog backend can thus support a `requestForwarderSession` GraphQL mutation.

A client would use the mutation to request a device → Forwarder session for a specific `deviceId`, and the backend returns details about the Forwarder instance, the session token required to access the session, and the session status, indicating whether the session is ready to be accessed.

```graphql
mutation {
  requestForwarderSession(deviceId: ID!) {
    forwarderSession {
      forwarderHost: String!
      forwarderPort: Int!
      sessionToken: String!
      sessionStatus: "CONNECTED" | "CONNECTING" | "DISCONNECTED"
    }
  }
}
```

To support the mutation, the backend should do the following:

- It checks whether the device is online. If not, a `device is offline` error is returned.
- It queries the Astarte interface `ForwarderSessionsState` for the specified device, listing the existing sessions.
    - If there are any session with `status`: `Connected` it returns the ID (UUIDv4) of the session as the `sessionToken`, `CONNECTED` as the `sessionStatus`, and the appropriate details about the Forwarder instance.
    - If there is no `Connected` session available, it returns a newly generated session ID (UUIDv4) as the `sessionToken`, `DISCONNECTED` as the `sessionStatus`, and the appropriate details about the Forwarder instance.
    It also sends an Astarte message to the device on the Astarte interface `ForwarderConnectionRequest`, indicating the host and port of the cloud Forwarder and the session token.

davidebriani:
The PR implements:
- A `forwarderSession` GraphQL query that can be used by clients to inspect the status of a device → Forwarder session, and know whether the session is ready to be accessed and it is closed.
- A `requestForwarderSession` mutation to allow clients to request a device → Forwarder session for a specific deviceId.

To support the request of a forwarder session, the backend does the following:
- It checks whether the device is online. If not, an error is returned.
- It queries the existing forwarder sessions for the device.
  - If there is a session that is either connected or re-connecting then the backend returns the session token of that session. An already connected session takes precedence.
  - If there is no available session, it generates a new session token and sends a request to the device asking to establish that session. It then returns the session token of that session.

With this approach, any existing forwarder sessions is reused whenever possible, and devices are not forced to establish a high number of simultaneous connections.

The backend assumes that there is a single instance of the [cloud Forwarder](https://github.com/edgehog-device-manager/edgehog_device_forwarder), and it expects to be configured with the host, port and protocol relative to forwarder's instance. The configuration is supplied through environment variables.

Closes #445, #446

### DIFF
--- a/backend/config/config.exs
+++ b/backend/config/config.exs
@@ -1,7 +1,7 @@
 #
 # This file is part of Edgehog.
 #
-# Copyright 2021 SECO Mind Srl
+# Copyright 2021-2024 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -87,6 +87,13 @@ config :edgehog, Edgehog.PromEx,
   drop_metrics_groups: [],
   grafana: :disabled,
   metrics_server: :disabled
+
+config :edgehog, :edgehog_forwarder, %{
+  hostname: "localhost",
+  port: 4001,
+  secure_sessions?: false,
+  enabled?: true
+}
 
 # Import environment specific config. This must remain at the bottom
 # of this file so it overrides the configuration defined above.
--- a/backend/config/runtime.exs
+++ b/backend/config/runtime.exs
@@ -146,6 +146,36 @@ if config_env() == :prod do
     disabled: !use_google_cloud_storage,
     json: s3.gcp_credentials
 
+  forwarder_secure_sessions? =
+    System.get_env("EDGEHOG_FORWARDER_SECURE_SESSIONS", "true") == "true"
+
+  forwarder_hostname = System.get_env("EDGEHOG_FORWARDER_HOSTNAME")
+
+  if forwarder_hostname != nil &&
+       (String.starts_with?(forwarder_hostname, "http://") ||
+          String.starts_with?(forwarder_hostname, "https://")) do
+    raise """
+    environment variable EDGEHOG_FORWARDER_HOSTNAME must contain only the
+    hostname without the http:// or https:// scheme.
+    """
+  end
+
+  forwarder_port = System.get_env("EDGEHOG_FORWARDER_PORT")
+
+  forwarder_port =
+    cond do
+      forwarder_port != nil -> String.to_integer(forwarder_port)
+      forwarder_secure_sessions? -> 443
+      true -> 80
+    end
+
+  config :edgehog, :edgehog_forwarder, %{
+    hostname: forwarder_hostname,
+    port: forwarder_port,
+    secure_sessions?: forwarder_secure_sessions?,
+    enabled?: forwarder_hostname != nil
+  }
+
   # ## Using releases
   #
   # If you are doing OTP releases, you need to instruct Phoenix
--- a/backend/lib/edgehog/astarte/device/forwarder_session.ex
+++ b/backend/lib/edgehog/astarte/device/forwarder_session.ex
@@ -1,0 +1,123 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Astarte.Device.ForwarderSession do
+  @type t :: %__MODULE__{
+          token: String.t(),
+          status: :connecting | :connected,
+          secure: boolean(),
+          forwarder_hostname: String.t(),
+          forwarder_port: integer()
+        }
+
+  @enforce_keys [:token, :status, :secure, :forwarder_hostname, :forwarder_port]
+  defstruct @enforce_keys
+
+  @behaviour Edgehog.Astarte.Device.ForwarderSession.Behaviour
+
+  alias Astarte.Client.AppEngine
+
+  @session_request_interface "io.edgehog.devicemanager.ForwarderSessionRequest"
+  @sessions_state_interface "io.edgehog.devicemanager.ForwarderSessionState"
+
+  @impl true
+  def list_sessions(%AppEngine{} = client, device_id) do
+    with {:ok, %{"data" => data}} <-
+           AppEngine.Devices.get_properties_data(client, device_id, @sessions_state_interface) do
+      sessions = parse_session_list(data)
+
+      {:ok, sessions}
+    end
+  end
+
+  @impl true
+  def fetch_session(%AppEngine{} = client, device_id, session_token)
+      when is_binary(session_token) do
+    with {:ok, sessions} <- list_sessions(client, device_id) do
+      case Enum.find(sessions, &(&1.token == session_token)) do
+        nil -> {:error, :forwarder_session_not_found}
+        session -> {:ok, session}
+      end
+    end
+  end
+
+  @impl true
+  def request_session(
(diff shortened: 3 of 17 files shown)


backend/lib/edgehog/astarte/device/forwarder_session.ex
```diff
@@ -0,0 +1,129 @@
...
+    end
+  end
+
+  @impl true
+  def request_session(%AppEngine{} = client, device_id, session_token)
+      when is_binary(session_token) do
+    forwarder_hostname = forwarder_hostname()
+    forwarder_port = forwarder_port()
+
+    data = %{session_token: session_token, host: forwarder_hostname, port: forwarder_port}
```
szakhlypa, line 73: `ForwarderSessionRequest` interface was updated  with `/request/secure` endpoint

szakhlypa (changes requested):

backend/lib/edgehog/astarte/device/forwarder_session.ex
davidebriani: Nice catch, thanks. I added support for a `EDGEHOG_FORWARDER_SECURE_SESSIONS` boolean environment variable that define whether the sessions are requested as secure or not.

backend/lib/edgehog/astarte.ex
```diff
@@ -600,6 +606,46 @@ defmodule Edgehog.Astarte do
     @led_behavior_module.post(client, device_id, behavior)
   end
 
+  defp fetch_forwarder_sessions(%Device{} = device) do
```
rbino, line 609: If you look at all the other functions above, they accept an `%AppEngine{}` client, not a `%Device{}`.

Keep the same convention by using `Devices.appengine_client_from_device/1` in the resolver (check out the [Astarte resolver](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog_web/resolvers/astarte.ex#L28) for the general pattern)

backend/lib/edgehog/astarte.ex
```diff
@@ -600,6 +606,46 @@ defmodule Edgehog.Astarte do
...
+    with {:ok, client} <- appengine_client_from_device(device) do
+      @forwarder_session_module.fetch_session_states(client, device.device_id)
```
rbino, line 611: Let's do `list_session_states` so the two function names jump a little more to the eye w.r.t. `fetch_session_states` and `fetch_session_state`

rbino (changes requested):

szakhlypa (changes requested):

backend/config/config.exs
```diff
@@ -88,6 +88,13 @@ config :edgehog, Edgehog.PromEx,
   grafana: :disabled,
   metrics_server: :disabled
 
+config :edgehog, :edgehog_forwarder, %{
+  protocol: "http",
+  hostname: "localhost",
+  port: 4001,
+  enabled?: true
+}
+
```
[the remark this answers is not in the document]
davidebriani: Yes, the configuration is actually used in the tests that verify the behaviour of the module itself, i.e. where the module is not mocked.

backend/lib/edgehog/astarte.ex
davidebriani: Thanks, fixed

backend/lib/edgehog/astarte.ex
davidebriani: Legit. I renamed the functions to `list_sessions` + `fetch_session`, and `list_forwarder_sessions` + `fetch_forwader_session`.

backend/config/runtime.exs
```diff
@@ -146,6 +146,20 @@ if config_env() == :prod do
...
+  forwarder_port = forwarder_port && String.to_integer(forwarder_port)
+
+  forwarder_secure_sessions? =
+    System.get_env("EDGEHOG_FORWARDER_SECURE_SESSIONS", "true") == "true"
+
+  config :edgehog, :edgehog_forwarder, %{
+    hostname: forwarder_hostname,
+    port: forwarder_port,
+    secure_sessions?: forwarder_secure_sessions?,
+    enabled?: Enum.all?([forwarder_hostname, forwarder_port, forwarder_secure_sessions?])
```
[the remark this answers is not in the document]
davidebriani: Nice catch! I fixed it

backend/config/runtime.exs
```diff
@@ -146,6 +146,20 @@ if config_env() == :prod do
     disabled: !use_google_cloud_storage,
     json: s3.gcp_credentials
 
+  forwarder_hostname = System.get_env("EDGEHOG_FORWARDER_HOSTNAME")
+  forwarder_port = System.get_env("EDGEHOG_FORWARDER_PORT")
+  forwarder_port = forwarder_port && String.to_integer(forwarder_port)
```
rbino, line 151: Should we default this to 443 give it will be the majority of the cases?

backend/config/runtime.exs
```diff
@@ -146,6 +146,20 @@ if config_env() == :prod do
     disabled: !use_google_cloud_storage,
     json: s3.gcp_credentials
 
+  forwarder_hostname = System.get_env("EDGEHOG_FORWARDER_HOSTNAME")
```
rbino, line 149: We've been bitten by this in the past, so do a very simple check here: if it starts with `http`, raise an error detailing the fact that you must pass an hostname, not a URL

doc/pages/admin/deploying_with_kubernetes.md
```diff
@@ -364,6 +370,9 @@ Values to be replaced
 - `S3-ASSET-HOST`: the asset host for the S3 storage, e.g. `storage.googleapis.com/<S3-BUCKET>` for
   GCP or `<S3-BUCKET>.s3.amazonaws.com` for AWS.
 - `S3-REGION`: the region where the S3 storage resides.
+- `EDGEHOG-FORWARDER-HOSTNAME`: the host for the instance of [Edgehog Device Forwarder](https://github.com/edgehog-device-manager/edgehog_device_forwarder).
```
rbino, line 373: Also add details here to ensure users don't pass a URL instead of a host

rbino (changes requested):

szakhlypa:
Should the `:remote_terminal` capability be removed from the device capabilities list when `:edgehog_forwarder` is disabled?

E.g. device has required interfaces for `:remote_terminal` capability in introspection, but backend has unconfigured `EDGEHOG_FORWARDER_HOSTNAME`.

doc/pages/admin/deploying_with_kubernetes.md
davidebriani: Updated

backend/config/runtime.exs
davidebriani: Legit, I added the check

backend/config/runtime.exs
davidebriani: I updated the code so that the port is inferred as 443 or 80 depending on the `EDGEHOG_FORWARDER_SECURE_SESSIONS` variable.
However if a `EDGEHOG_FORWARDER_PORT` is specified, it takes precedence.

backend/config/runtime.exs
```diff
@@ -146,6 +146,36 @@ if config_env() == :prod do
...
+      forwarder_port != nil -> String.to_integer(forwarder_port)
+      forwarder_secure_sessions? -> 443
+      true -> 80
+    end
+
+  config :edgehog, :edgehog_forwarder, %{
+    hostname: forwarder_hostname,
+    port: forwarder_port,
+    secure_sessions?: forwarder_secure_sessions?,
+    enabled?: Enum.all?([forwarder_hostname, forwarder_port])
```
szakhlypa, line 176: Nitpicking: after recent changes, `EDGEHOG_FORWARDER_PORT` env variable is no longer required, `forwarder_port` will always have a valid value, so the `enabled?` value should only depend on `forwarder_hostname`

davidebriani: Thanks, now `enabled?` is computed as `forwarder_hostname != nil`.

rbino (approved):

szakhlypa (approved):

PR #463, opened 2024-03-08
Refactor create update channel's form info

harisbikovic:
Create GraphQL fragment for data into CreateUpdateChannel form and use it on UpdateChannelCreate page. 
This PR also gets rid of eslint relay/unused-field warnings in UpdateChannelCreate page.

### DIFF
--- a/frontend/src/pages/UpdateChannelCreate.tsx
+++ b/frontend/src/pages/UpdateChannelCreate.tsx
@@ -42,11 +42,7 @@ import { Route, useNavigate } from "Navigation";
 const GET_DEVICE_GROUPS_QUERY = graphql`
   query UpdateChannelCreate_getDeviceGroups_Query {
     deviceGroups {
-      id
-      name
-      updateChannel {
-        name
-      }
+      ...CreateUpdateChannel_DeviceGroupFragment
     }
   }
 `;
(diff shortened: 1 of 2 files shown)


frontend/src/pages/UpdateChannelCreate.tsx
```diff
@@ -42,11 +42,7 @@ import { Route, useNavigate } from "Navigation";
 const GET_DEVICE_GROUPS_QUERY = graphql`
   query UpdateChannelCreate_getDeviceGroups_Query {
     deviceGroups {
```
szakhlypa, line 44: Is there any reason to query the `deviceGroups` field  in `pages/UpdateChannelCreate.tsx`, when when it is only used in `forms/CreateUpdateChannel.tsx`?

szakhlypa (changes requested):

frontend/src/pages/UpdateChannelCreate.tsx
harisbikovic: Do you think that query should be transferred and used in  `forms/CreateUpdateChannel.tsx` instead? If not, what do you suggest?

szakhlypa: I suggest you update `CREATE_UPDATE_CHANNEL_FRAGMENT` to query for the `deviceGroups` field `on RootQueryType` and use the fragment in `GET_DEVICE_GROUPS_QUERY`. You can also update component propetry name, like `targetGroupsRef` -> `createUpdateChannelRef`

szakhlypa:
Closing as superseded by #505

PR #465, opened 2024-03-14
Disable REMOTE_TERMINAL capability for devices when Edgehog Forwarder is disabled

davidebriani:
Currently, if devices supports the Astarte interfaces for Forwarder sessions but a Forwarder instance is not configured in Edgehog, the devices still reports a REMOTE_TERMINAL capability.

Since the presence of the capability controls whether the remote terminal functionality is enabled on Edgehog frontend, and it would be odd to display it enabled without a supporting Forwarder instance, this PR disables the capability on devices if a Forwarder instance is not configured.


davidebriani:
I'm closing this as superseded by #467, which provides a more compelling solution.

PR #467, opened 2024-03-15
Disable Remote Terminal feature on frontend when Edgehog Forwarder is disabled

davidebriani:
The PR introduces the following changes:
- Define a GraphQL query to fetch the configuration of the forwarder instance, if available. The presence of a configuration indicates that the forwarding functionalities are supported, otherwise they are disabled.
- When there is no forwarder config available from the APIs avoid displaying the button to open remote terminal sessions, since that functionality requires a configured forwarder.

### DIFF
--- a/backend/test/edgehog_web/schema/query/forwarder_config_test.exs
+++ b/backend/test/edgehog_web/schema/query/forwarder_config_test.exs
@@ -1,0 +1,118 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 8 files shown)


backend/test/edgehog_web/schema/query/forwarder_config_test.exs
```diff
@@ -0,0 +1,118 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule EdgehogWeb.Schema.Query.ForwarderConfigTest do
+  use EdgehogWeb.ConnCase, async: true
```
rbino, line 22: This can't be `async: true` since it modifies global configuration

rbino (changes requested):

rbino (approved):

szakhlypa (approved):

Issue #447, opened 2024-02-13
Add an Open Remote Terminal button to access remote terminal sessions on devices


davidebriani:
Building on top of #444, #445 and #446, the frontend can implement a button that allows users to open remote terminal sessions on devices that support the functionality.

The frontend can display the "Open remote terminal" button on each device page, if the device has the `REMOTE_TERMINAL` capability.
The button is disabled if the device is offline.

When the button is pressed, the frontend performs the `requestForwarderSession` GraphQL mutation:
- If the `sessionStatus` in the mutation’s response is not `CONNECTED`, it means that the device did not connect to the Forwarder yet and the session is not ready. The frontend then polls the GraphQL query `forwarderSession` until the status of the session is `CONNECTED`. In the mean time, it will display a loading state.
- If the session status is `CONNECTED`, the frontend composes the URL for the remote terminal session: `https://<forwarder_host>:<forwarder_port>/v1/<session_token>/http/7681`. The frontend then opens a new browser tab and navigates to the URL of the remote terminal session.

davidebriani:
Closing as implemented by #472

PR #470, opened 2024-03-18
Document the implemented support for Edgehog Device Forwarder

### LINKED ISSUE #444: Expose a `REMOTE_TERMINAL` capability for devices that support remote terminal sessions
In the scope of supporting the possibility of establishing remote terminal sessions on devices, we need a way to distinguish devices that have this functionality.

A new `REMOTE_TERMINAL` capability should be defined for devices.
Specifically, devices that support the Astarte interfaces [`io.edgehog.devicemanager.RemoteTerminalRequest`](https://github.com/edgehog-device-manager/edgehog-astarte-interfaces/blob/main/io.edgehog.devicemanager.RemoteTerminalRequest.json) and [`io.edgehog.devicemanager.ForwarderSessionsState`](https://github.com/edgehog-device-manager/edgehog-astarte-interfaces/blob/main/io.edgehog.devicemanager.ForwarderSessionsState.json) supports remote terminal sessions.

Note: the two interfaces have more general purposes and could be used to communicate with different device services, but for the sake of simplicity we will now assume that if a device supports those interfaces then it has a TTYD instance up and running. Future developments will evaluate how a device can advertise specific services such as TTYD.

davidebriani:
Note down a summary of the implemented functionality in the changelog file, and expand the documentation page on deployments to describe how the [Device Forwarder](https://github.com/edgehog-device-manager/edgehog_device_forwarder) can be configured alongside Edgehog.

Closes #444, #445, #446, #447.

### DIFF
--- a/doc/pages/admin/deploying_with_kubernetes.md
+++ b/doc/pages/admin/deploying_with_kubernetes.md
@@ -81,10 +81,9 @@ Controllers are outside the scope of this guide.
 
 Once you have the Load Balancer IP (obtained in the [previous
 step](#installing-nginx-ingress-controller-and-cert-manager-example)), head to your DNS provider and
-point two domains (one for the backend and one for the frontend) to that IP address.
+point three domains (one for the backend, one for the frontend, and one for the [Device Forwarder](https://github.com/edgehog-device-manager/edgehog_device_forwarder)) to that IP address.
 
-Save the two hosts (e.g. `api.edgehog.example.com` and `edgehog.example.com`) since they're going to
-be needed for the following steps.
+Save the three hosts (e.g. `api.edgehog.example.com`, `edgehog.example.com`, and `forwarder.edgehog.example.com`) since they're going to be needed for the following steps.
 
 ### Secrets
 
@@ -108,10 +107,17 @@ Values to be replaced
 
 #### Secret key base
 
-This command creates the secret key base used by Phoenix:
+This command creates the secret key base used by Phoenix for the backend:
 
 ```bash
 $ kubectl create secret generic -n edgehog edgehog-secret-key-base \
+  --from-literal="secret-key-base=$(openssl rand -base64 48)"
+```
+
+Another secret key base can be generated for the device forwarder:
+
+```bash
+$ kubectl create secret generic -n edgehog edgehog-device-forwarder-secret-key-base \
   --from-literal="secret-key-base=$(openssl rand -base64 48)"
 ```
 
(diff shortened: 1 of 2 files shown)


doc/pages/admin/deploying_with_kubernetes.md
```diff
@@ -562,6 +648,8 @@ metadata:
     cert-manager.io/cluster-issuer: letsencrypt
     kubernetes.io/ingress.class: nginx
     nginx.ingress.kubernetes.io/proxy-body-size: <MAX-UPLOAD-SIZE>
+    nginx.ingress.kubernetes.io/proxy-read-timeout: <CONNECTION-TIMEOUT>
+    nginx.ingress.kubernetes.io/proxy-send-timeout: <CONNECTION-TIMEOUT>
```
rbino, line 652: Given we're not configuring the Forwarder timeout, I'd just put `120` here for both to minimize the _tl;dr_ factor (also given that right now there's only one good default and it's 120s)

rbino (changes requested):

doc/pages/admin/deploying_with_kubernetes.md
davidebriani: > there's only one good default and it's 120s

I'm not sure about that. It could be 61s or 180s, only experience will tell.
The timeout value also has implications on connections to Edgehog's frontend and backend.
In addition, a high value might facilitate DoS attacks.

I agree about the TL;DR, although I would still note down the reason why we are messing with these timeouts somewhere.

rbino: works for me

davidebriani: Brilliant. I edited it so that it's more copy&paste-friendly while keeping the explanation on the succeeding notes.

doc/pages/admin/deploying_with_kubernetes.md
```diff
@@ -421,6 +427,58 @@ Values to be replaced
...
+        - name: PORT
+          value: "4000"
+        - name: PHX_HOST
+          value: <DEVICE-FORWARDER-HOST>
+        - name: SECRET_KEY_BASE
+          valueFrom:
+            secretKeyRef:
+              key: secret-key-base
+              name: edgehog-device-forwarder-secret-key-base
+        image: edgehogdevicemanager/edgehog-device-forwarder:snapshot
```
rbino, line 470: Let's use `edgehog-device-forwarder:0.1.0` since it's there now

rbino (changes requested):

doc/pages/admin/deploying_with_kubernetes.md
davidebriani: Done

rbino (approved):

PR #475, opened 2024-03-21
Add v0.8.0-dev to version dropdown

rbino:

### DIFF
--- a/doc/versions.js
+++ b/doc/versions.js
@@ -1,5 +1,9 @@
 var versionNodes = [
   {
+    version: "v0.8.0-dev",
+    url: "https://docs.edgehog.io/snapshot"
+  },
+  {
     version: "v0.7.1",
     url: "https://docs.edgehog.io/0.7"
   }


doc/versions.js
```diff
@@ -1,4 +1,8 @@
 var versionNodes = [
+  {
+    version: "v0.8.0-dev",
```
davidebriani, line 3: I would add `version: "snapshot"` in the `main` branch, branch out with `release-0.8`, and add `version: "v0.8.0-dev"` there.
Or maybe directly add `version: "v0.8.0-rc.0"` if you use the same PR to prepare the `release-0.8` for tagging.
What do you think?

davidebriani (changes requested):

doc/versions.js
rbino: I've used `v0.8.0-dev` because that's what we have everywhere else in `main`. The idea of the steps is:
- Merge this as it is
- Branch-out `release-0.8`
- In `release-0.8` `s/v0.8.0-dev/v0.8.0-rc.0` everywhere
- Tag `v0.8.0-rc.0`
- In main `s/v0.8.0-dev/v0.9.0-dev` everywhere, new release cycle

davidebriani (approved):

PR #478, opened 2024-03-21
Forward port `main` into `from-ash-we-rise`

rbino:
Include new developments from earlier branches like Tenant cleanup and Device Forwarder integration


rbino:
Closing in favor of a new PR doing the same thing

PR #481, opened 2024-03-25
Port Device capabilities to Ash

rbino:
Expose them as a calculation which in turn depends on the `DeviceStatus` calculation


backend/lib/edgehog/devices/device/calculations/capabilities.ex
```diff
@@ -18,18 +18,24 @@
...
-    else
-      _ -> {:ok, []}
-    end
+  @impl true
+  def load(_query, _opts, _context) do
+    [:device_status]
+  end
+
+  @impl true
+  def calculate(devices, opts, _context) do
```
davidebriani, line 32: `opts` is unused

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -147,6 +147,11 @@ defmodule Edgehog.Devices.Device do
       filterable? false
     end
 
+    calculate :capabilities, {:array, Types.Capability} do
+      allow_nil? false
+      calculation Calculations.Capabilities
+    end
```
davidebriani, line 153: I'm not sure if it gets exposed via GraphQL, but as a best practice in general we could also add descriptions to calculations. Basically, the same description that was previously present in the GraphQL schema:

```elixir
description: "The capabilities that the device can support".
```

backend/test/edgehog_web/schema/query/device_test.exs
```diff
@@ -349,6 +349,95 @@ defmodule EdgehogWeb.Schema.Query.DeviceTest do
...
+
+      assert length(capabilities) == length(all_capabilities)
+
+      for capability <- all_capabilities do
+        assert capability in capabilities
+      end
+    end
+
+    test "contain only geolocation for empty introspection", ctx do
+      %{tenant: tenant, id: id, device_id: device_id} = ctx
```
davidebriani, line 420: I'm not sure if it's clear what `id` stands for. How about `graphql_device_id` or something similar?
Otherwise, since you're only using `id` to call `device_query/1`, you could pass the `device` instead and let `device_query/1` do the rest:

```elixir
defp device_query(opts) do
  ...
  device = Keyword.fetch!(opts, :device)
  device_id = AshGraphql.Resource.encode_relay_id(device)
  ...
end
```

davidebriani (changes requested):

davidebriani (approved):

PR #482, opened 2024-03-28
Port Device Tags features to Ash

rbino:


backend/lib/edgehog/labeling/tag.ex
```diff
@@ -19,21 +19,40 @@
...
-    field :name, :string
+  actions do
+    defaults [:read, :destroy]
 
-    timestamps()
+    create :create do
+      primary? true
+      upsert? true
+      upsert_identity :name_tenant_id
+    end
```
davidebriani, line 34: We will probably need some normalization on the `name` attribute so that `tag`, `Tag`, `TAG`, and `      TAG     ` result in the same tag `tag` being applied.

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -117,6 +118,18 @@ defmodule Edgehog.Devices.Device do
                use_identities: [:name_tenant_id]
              )
     end
+
+    action :list_assigned_tag_names, {:array, :string} do
+      # TODO: this should probably live in the Tag resource, but currently is blocked
+      # by https://github.com/ash-project/ash_graphql/issues/119
+      # Move it there once it's fixed
```
davidebriani, line 125: It seems *someone* already fixed that and we can just update our Ash.GraphQL dependency.

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -117,6 +118,18 @@ defmodule Edgehog.Devices.Device do
...
+      description "Fetches the list of all tags currently assigned to some device."
+
+      run fn _input, context ->
+        opts = Ash.context_to_opts(context, action: :assigned_to_devices)
+        Ash.list(Edgehog.Labeling.Tag, :name, opts)
+      end
```
davidebriani, line 131: Does this respect the tenant that was set in the context or it will list all tags from all tenants?

davidebriani (changes requested):

backend/lib/edgehog/devices/device/device.ex
rbino: The tenant is inside the context, `context_to_opts` + passing those opts to `list` does exactly that job.

In fact, this should be enforced by Ash by returning an error if you don't pass a tenant (it is, for example, when using `Ash.read/1`), but I have to check because from a quick test it seems it's not enforced on aggregates.

Still, the code as is currently passes the tenant.

rbino: Oh ok we just have to update, [someone](https://github.com/ash-project/ash/pull/952) has already fixed that :eyes:

backend/lib/edgehog/labeling/device_tag.ex
```diff
@@ -19,31 +19,50 @@
...
-  alias Edgehog.Labeling.Tag
-
-  @primary_key false
-  schema "devices_tags" do
-    field :tenant_id, :integer, autogenerate: {Edgehog.Repo, :get_tenant_id, []}
-    field :tag_id, :id, primary_key: true
-    field :device_id, :id, primary_key: true
+  use Edgehog.MultitenantResource,
+    api: Edgehog.Labeling,
+    tenant_id_in_primary_key?: true
```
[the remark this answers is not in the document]
rbino: I think it already was here
https://github.com/edgehog-device-manager/edgehog/blob/main/backend/priv/repo/migrations/20220517063737_create_devices_tags.exs#L28

Anyway, at the end of the porting there will for sure be a pass to exactly match the tables from the current migrations with the ones generated by Ash; right now we're still not generating them from Ash but reusing the old ones but in the end we have to make sure they match so we can start to use Ash generators from that point onwards.

backend/lib/edgehog/labeling/tag/tag.ex
szakhlypa, line 1: The `Tag` type in the latest GraphQL schema has a `tenant` field, is this intentional?

```graphql
type Tag {
  id: ID!
  name: String!
  tenant: TenantInfo!
}
```

backend/lib/edgehog/labeling/tag/tag.ex
```diff
@@ -0,0 +1,89 @@
...
+    read :assigned_to_devices do
+      description "Returns Tags currently assigned to some device."
+      prepare build(filter: expr(exists(device_tags, true)))
+    end
+  end
+
+  attributes do
+    integer_primary_key :id
+
+    attribute :name, :string, allow_nil?: false
```
davidebriani, line 65: Users might provide `        ` or an empty string as the tag, and they would end up persisted as an empty string tag, which doesn't seem so valuable. 
We might want to add `min_length: 1` here.

backend/test/edgehog_web/schema/mutation/add_device_tags_test.exs
```diff
@@ -0,0 +1,146 @@
...
+      result =
+        add_device_tags_mutation(tenant: tenant, id: id, tags: ["FOO", "bAr", "  BaZ  ", "baz"])
+
+      assert %{"tags" => tags} = extract_result!(result)
+      assert length(tags) == 3
+      tag_names = Enum.map(tags, &Map.fetch!(&1, "name"))
+      assert "foo" in tag_names
+      assert "bar" in tag_names
+      assert "baz" in tag_names
+    end
```
davidebriani, line 57: Do we want to support normalizing the user input also when trying to remove tags?
For example, a user might first add a `FOO` tag and then try to remove the `FOO` tag: I fear that the tag won't be removed, since `FOO` will not match the persisted `foo` tag.

backend/lib/edgehog/labeling/device_tag.ex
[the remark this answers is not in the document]
davidebriani: It can actually make sense to have it in the primary key, since foreign keys that reference device tags should also match on the tenant_id anyways, to properly link multi-tenant resources.
Ok, let's review this whole topic at the end of the migration to Ash.

davidebriani (approved):

PR #487, opened 2024-04-03
Port Device Groups to Ash

rbino:


backend/lib/edgehog/groups/device_group/manual_relationships/devices.ex
```diff
@@ -0,0 +1,53 @@
...
+  require Ash.Query
+
+  @impl true
+  def select(_opts) do
+    [:selector]
+  end
+
+  @impl true
+  def load(groups, _opts, %{query: query}) do
+    # We're doing N+1 queries here, but it's probably inevitable at this point
```
rbino, line 34: Note that we were doing N+1 queries also before. I might have an idea to maybe skip the N+1 here, but I'll think about that in a second pass.

backend/lib/edgehog/groups/device_group/device_group.ex
```diff
@@ -0,0 +1,138 @@
...
+      The identifier of the device group.
+
+      It should start with a lower case ASCII letter and only contain \
+      lower case ASCII letters, digits and the hyphen - symbol.
+      """
+
+      allow_nil? false
+    end
+
+    # TODO: custom type here
```
rbino, line 94: Left this for a second pass, the idea being parsing the selector on load instead of doing it inside the `devices` manual relationship

backend/test/edgehog_web/schema/mutation/delete_device_group_test.exs
```diff
@@ -19,22 +19,52 @@
...
 
   import Edgehog.GroupsFixtures
 
-  alias Edgehog.Groups
   alias Edgehog.Groups.DeviceGroup
 
-  describe "deleteDeviceGroup field" do
-    setup do
-      {:ok, device_group: device_group_fixture()}
+  @moduletag :ported_to_ash
```
davidebriani, line 30: We don't need to specify `@moduletag :ported_to_ash` twice for the same module.

davidebriani (changes requested):
The ported code is missing some functionalities, you might want to track them as TODOs:
- retrieving the groups of a device
- relationship of a group with the update channel

davidebriani (approved):

PR #488, opened 2024-04-10
Ash 3.0 update

rbino:
- Update Elixir and Erlang versions
- Update Ash and all Ash components to 3.0 compatible versions
- Rename Api to Domain
- Remove legacy breaking-change compatibility options
- Explicitly accept attributes in actions
- Mark all public attributes, calculations and relationships
- Add attribute_public? false to relationships
- Add the context argument to module based validations and calculations
- Explicitly load relationship fields in calculations
- Change Ash.Calculation to Ash.Resource.Calculation
- Remove hide_fields almost everywhere, relying instead on the public? option
- Mark non atomic actions with require_atomic? false
- Remove unnecessary domain from cross-domain relationships
- Replace Ash.Query.expr with Ash.Expr.expr
- Implement Ash.ToTenant for the Tenant resource
- Remove parse_attribute from multitenancy, superseded by Ash.ToTenant
- Replace calls to domain modules with calls to the Ash module
- Handle new nested errors in JSON API tests


.tool-versions
szakhlypa, line 1: We should also update `backend/Dockerfile`

szakhlypa (changes requested):

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -158,105 +162,113 @@ defmodule Edgehog.Devices.Device do
...
-      attribute_writable? true
+      allow_nil? false
     end
 
     belongs_to :system_model_part_number, Edgehog.Devices.SystemModelPartNumber do
+      public? true
       attribute_type :string
       source_attribute :part_number
       destination_attribute :part_number
       attribute_writable? true
```
davidebriani, line 173: `attribute_writable? true` on `system_model_part_number` should not be needed, since when not set it will default to the relationship's `writable?` setting, which defaults to `true`.

backend/lib/edgehog/devices/system_model_part_number.ex
```diff
@@ -20,25 +20,34 @@
...
+
+    create :create do
+      primary? true
+      accept [:part_number]
+    end
+
+    update :update do
+      primary? true
+      accept [:part_number]
+    end
```
davidebriani, line 43: Since `create` and `update` now only accepts `[:part_number]`, how does the `system_model_id` get populated?
The same doubt stands for the HardwareTypePartNumber.

davidebriani (changes requested):

backend/lib/edgehog/devices/system_model_part_number.ex
rbino: That gets populated from the other using `manage_relationship` (see e.g. [here](https://github.com/edgehog-device-manager/edgehog/pull/488/files/2b31c47af1c6e07b59589a0fb8460369274073fb#diff-d891c3746b552f30751597445010d266903425bcebc4544bdbd5dd6206aa9709R99)) which is always able to modify the join attributes of the two resources

backend/lib/edgehog/astarte/realm/realm.ex
```diff
@@ -40,19 +40,20 @@ defmodule Edgehog.Astarte.Realm do
...
     integer_primary_key :id
 
-    attribute :name, :string, allow_nil?: false
+    attribute :name, :string do
+      public? true
+      allow_nil? false
+    end
 
     attribute :private_key, :string do
+      public? true
```
szakhlypa, line 56: Should [`sensitive? true`](https://hexdocs.pm/ash/dsl-ash-resource.html#attributes-attribute-sensitive?) be added as well?

szakhlypa:
new GraphQL schema adds few fields like
```graphql
type HardwareTypePartNumber {
  hardwareTypeId: Int
}
type SystemModelPartNumber {
  systemModelId: Int
  devices(sort: [DeviceSortInput], filter: DeviceFilterInput, limit: Int, offset: Int): [Device!]!
}
type SystemModel implements Node {
  hardwareTypeId: Int
}
type Device implements Node {
  systemModelPartNumber: SystemModelPartNumber
}
```
Is this intentional?

szakhlypa:
Do we want to have nullability on "relationships" fields even if we expect them to be not nullable?
```graphql
type SystemModel implements Node {
  hardwareType: HardwareType
}
```

rbino:
Regarding the first, what is exactly the difference w.r.t pre-3.0?

Regarding the second, it derives from the fact that not all relationships have `allow_nil? false`. This was already like this and will be fixed when we do the pass to make sure that the initial DB schema matches the existing one.

szakhlypa:
> Regarding the first, what is exactly the difference w.r.t pre-3.0?

There are new fields in GraphQL schema types (commit hash `442275c` -> `1f047b9`):
| Type  | new field | new field type | 
| ------------- | ------------- | ------------- |
| HardwareTypePartNumber  | hardwareTypeId  | Int |
| SystemModelPartNumber  | systemModelId  | Int |
| SystemModelPartNumber  | devices  | [Device!]! |
| SystemModel  | hardwareTypeId  | Int |
| Device  | systemModelPartNumber  | SystemModelPartNumber |



expanded example for `HardwareTypePartNumber` type below:

GraphQL schema pre (`442275c` commit's hash)
```graphql
type HardwareTypePartNumber {
  id: ID!
  partNumber: String!
  hardwareType: HardwareType
}
```

GraphQL schema after (`1f047b9` commit's hash)
```graphql
type HardwareTypePartNumber {
  id: ID!
  partNumber: String!
  hardwareTypeId: Int
  hardwareType: HardwareType
}
```
So the type `HardwareTypePartNumber` gets new field `hardwareTypeId` of type `Int`. Same for other fields from the table above.

rbino:
@szakhlypa fixed

szakhlypa (approved):

PR #490, opened 2024-04-12
Port Base Images to Ash

davidebriani:


backend/lib/edgehog/base_images/base_image/changes/handle_file_deletion.ex
```diff
@@ -0,0 +1,54 @@
...
+                    BucketStorage
+                  )
+
+  @impl true
+  def change(%Ash.Changeset{valid?: false} = changeset, _opts, _context), do: changeset
+
+  def change(changeset, _opts, _context) do
+    file_url = Ash.Changeset.get_data(changeset, :url)
+
+    if file_url != nil do
```
rbino, line 38: `:url` is `allow_nil? false` so you can assume it's not nil and remove the conditional check

backend/test/support/fixtures/base_images_fixtures.ex
```diff
@@ -64,20 +69,29 @@ defmodule Edgehog.BaseImagesFixtures do
...
 
-    # Stub StorageMock since create_base_image will call it
-    Mox.stub_with(Edgehog.BaseImages.StorageMock, Edgehog.Mocks.BaseImages.Storage)
+    {file_url, opts} = Keyword.pop(opts, :url, "https://example.com/ota.bin")
 
-    attrs =
-      Enum.into(opts, %{
+    # Stub StorageMock with provided :url since create_base_image will call it
+    Edgehog.BaseImages.StorageMock
+    |> Mox.stub(:store, fn _, _ -> {:ok, file_url} end)
```
rbino, line 83: Instead of doing this you could have an alternative `create` action (e.g. `create :no_upload`) that accepts the `:file_url` and gets called by the fixture. This way you don't have to deal with Mox here (and leave the calling code "pure", as in no stubs implicitly defined)

backend/test/edgehog_web/schema/query/base_image_test.exs
```diff
@@ -19,210 +19,97 @@
...
+    fixture = base_image_fixture(tenant: tenant)
+    id = AshGraphql.Resource.encode_relay_id(fixture)
 
-  defp description_fixture(locales) when is_list(locales) do
-    for locale <- locales, into: %{} do
-      {locale, "#{locale} description"}
-    end
-  end
+    Edgehog.BaseImages.StorageMock
+    |> expect(:delete, fn _ -> :ok end)
```
rbino, line 109: Same here, you can define a utility `destroy` action that you can use in tests without worrying about stubbing the storage

rbino (changes requested):
Some nitpicking but overall a very good start with Ash!

backend/test/edgehog_web/schema/mutation/create_base_image_collection_test.exs
```diff
@@ -19,22 +19,144 @@
...
+      system_model_id = AshGraphql.Resource.encode_relay_id(system_model)
+      _ = Ash.destroy!(system_model)
+
+      result =
+        create_base_image_collection_mutation(
+          tenant: tenant,
+          system_model_id: system_model_id
+        )
+
+      # TODO: wrong fields returned by AshGraphql
```
rbino, line 68: What is the upstream problem here?

szakhlypa (changes requested):

backend/lib/edgehog/base_images/base_image_collection/base_image_collection.ex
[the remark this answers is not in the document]
davidebriani: You're right, I kept just the single file without the wrapping folder for now.

PR #491, opened 2024-04-12
Port Device Trigger handling to Ash

rbino:

### DIFF
--- a/backend/lib/edgehog/devices/device/changes/initialize_from_device_status.ex
+++ b/backend/lib/edgehog/devices/device/changes/initialize_from_device_status.ex
@@ -1,0 +1,58 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Devices.Device.Changes.InitializeFromDeviceStatus do
+  use Ash.Resource.Change
+
+  @device_status_attributes [:last_connection, :last_disconnection, :online]
+
+  @impl true
+  def change(changeset, _opts, _context) do
+    # After we handle a trigger, we check if any of the fields we can populate from the device
+    # status is nil. If there is any, we try to initialize them from the device status.
+    Ash.Changeset.after_action(changeset, fn _changeset, device ->
+      if uninitialized?(device) do
+        device
+        |> Ash.load!(:device_status, reuse_values?: true)
+        |> initialize_from_status()
+      else
+        {:ok, device}
+      end
+    end)
+  end
+
+  defp uninitialized?(device) do
+    Enum.any?(@device_status_attributes, &(Map.get(device, &1) == nil))
+  end
+
+  defp initialize_from_status(%{device_status: nil} = device) do
+    # If device status retrieval fails, we just return the device as is. It's going to be
+    # retried during the next event.
+    {:ok, device}
+  end
+
+  defp initialize_from_status(device) do
+    attributes = Map.take(device.device_status, @device_status_attributes)
+
+    device
+    |> Ash.Changeset.for_update(:from_device_status, attributes)
+    |> Ash.update()
+  end
+end
--- a/backend/lib/edgehog/devices/device/device.ex
+++ b/backend/lib/edgehog/devices/device/device.ex
@@ -26,6 +26,7 @@ defmodule Edgehog.Devices.Device do
     ]
 
   alias Edgehog.Devices.Device.Calculations
+  alias Edgehog.Devices.Device.Changes
   alias Edgehog.Devices.Device.ManualRelationships
   alias Edgehog.Devices.Device.Types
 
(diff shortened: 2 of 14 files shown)


backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -20,281 +20,274 @@
...
-  alias Edgehog.Devices
+  alias Edgehog.Devices.Device
   alias Edgehog.OSManagement
 
   import Edgehog.AstarteFixtures
   import Edgehog.DevicesFixtures
   import Edgehog.OSManagementFixtures
 
-  @system_info_interface "io.edgehog.devicemanager.SystemInfo"
+  describe "process_event for device events" do
```
rbino, line 32: Note that _before_ changing all tests I've verified the old ones passed (apart from the changed error codes), to make sure that we're still compatible with the existing triggers

szakhlypa (changes requested):

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -0,0 +1,141 @@
...
+      event: %Ash.Union{value: event}
+    } = input.arguments.trigger_payload
+
+    # We only need the realm id and the tenant id, so we only read those
+    read_query =
+      Realm
+      |> Ash.Query.select([:id])
+      |> Ash.Query.load(tenant: [:tenant_id])
+
+    with {:ok, realm} <- Realm.fetch_by_name(realm_name, query: read_query) do
```
davidebriani, line 51: If I am not mistaken, different Edgehog tenants could have Astarte realms with the same name.
Instead of fetching the realm only by name and load the tenant as a realm's association, we could instead identify *both* the correct realm and tenant beforehand, thanks to the route params of the API that handles incoming triggers (`/tenants/:tenant_slug/triggers`) and the `Astarte-Realm` HTTP header of the request. The tenant can then be pushed downstream to the database queries.

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -20,281 +20,274 @@
...
-      device = device_fixture(realm)
+      realm = realm_fixture(cluster_id: cluster.id, tenant: tenant)
 
-      {:ok, cluster: cluster, realm: realm, device: device}
+      device =
+        device_fixture(
+          realm_id: realm.id,
+          online: false,
+          last_connection: utc_now_second() |> DateTime.add(-10, :minute),
+          last_disconnection: utc_now_second() |> DateTime.add(-50, :minute),
```
davidebriani, line 44: Nitpicking: the device is not online, but it has a `last_connection` more recent than `last_disconnection`.

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -72,6 +73,98 @@ defmodule Edgehog.Devices.Device do
...
+      upsert? true
+      upsert_identity :device_id_realm_id_tenant_id
+      upsert_fields [:online, :last_connection, :updated_at]
+
+      accept [:realm_id]
+      argument :device_id, :string, allow_nil?: false
+      argument :timestamp, :datetime, allow_nil?: false
+
+      # Only if created
+      change Changes.InitializeFromDeviceStatus
```
davidebriani, line 86: Why the comment says that the change is only run when the resource is created? How does that happen?

davidebriani (changes requested):

backend/lib/edgehog_web/controllers/trigger_controller.ex
```diff
@@ -21,119 +21,23 @@
...
-  alias Edgehog.Astarte
-  alias Edgehog.OSManagement
-  alias EdgehogWeb.FallbackController
+  alias Edgehog.Triggers
 
-  require Logger
+  def process_event(conn, _params) do
+    # We want to raise here so that we can have loud log messages that makes us notice
+    # problems. On the other side we don't have a human, so returning errors more informative
+    # than a 500 does not do much good.
```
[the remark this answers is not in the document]
davidebriani: For instance, we could produce an evident log when the realm does not exist, and return a 404 or 400 error.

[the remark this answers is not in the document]
rbino: True that, I took a shortcut and conflated the two things because even implementing [`Plug.Exception`](https://hexdocs.pm/phoenix/custom_error_pages.html#custom-exceptions) the exception is not written to the log if it has an error in the 4xx range. In this case it's better to restore the Fallback controller and handle the error handling/logging there.

backend/lib/edgehog/devices/device/changes/initialize_from_device_status.ex
```diff
@@ -0,0 +1,60 @@
...
+          |> initialize_from_status(attributes)
+      end
+    end)
+  end
+
+  defp uninitialized_device_status_attributes(device) do
+    # :online is not present here since it has a default, so it can never be nil
+    [:last_connection, :last_disconnection]
+    |> Enum.filter(&(Map.get(device, &1) == nil))
+  end
```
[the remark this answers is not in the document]
rbino: The current implementation is coherent with the [old implementation](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/astarte.ex#L439): the device was created if not existing, retrieving the initial info from Astarte, and _then_ it was updated with the event, so the event changes were winning.

Here we do things differently due to the fact that we're able to do an atomic change (i.e. before we were doing two DB calls, read and then update, for each event, here we do a single one), so we actually retrieve the info from Astarte _after_ we upsert the device, only if some fields are `nil` (otherwise we would do an API call each time there's an event).

All that said, your observation makes sense, theoretically your corner case would be rectified just afterwards since we should receive a `device_disconnected` event, but given we have the info we should consider `Astarte` to be the authority here, I'll change it

backend/lib/edgehog/devices/device/device.ex
davidebriani: II agree with making things more explicit as possible, comments included since as stated right now the comment is not clear.

For instance, in the future we might have some automatic mechanism that creates devices on Edgehog if they exist on Astarte: so a device registered in Astarte would appear in Edgehog even though it never connected yet. Implementing such a feature would subtly change the logic here, where we made the assumption that a created device does not have `nil` fields.

davidebriani (approved):

davidebriani (approved):

PR #506, opened 2024-04-29
Port OSManagement to Ash

### LINKED ISSUE #496: Port OS Management Domain to Ash

rbino:
Close #496


backend/lib/edgehog/os_management/ota_operation/ota_operation.ex
```diff
@@ -0,0 +1,170 @@
...
+      description "The timestamp of the last update to the operation"
+      public? true
+    end
+  end
+
+  relationships do
+    belongs_to :device, Edgehog.Devices.Device do
+      public? true
+      allow_nil? false
+    end
```
davidebriani, line 157: We can add `attribute_public? false` to only expose `:device` and not `:device_id`.

backend/lib/edgehog/os_management/ota_operation/ota_operation.ex
```diff
@@ -0,0 +1,170 @@
...
+  end
+
+  calculations do
+    calculate :final_state?, :boolean, expr(status == :success or status == :failure)
```
davidebriani, line 161: I would keep using the `status` concept insteand of `state`. For example `:terminal_status?`.

davidebriani (changes requested):

backend/lib/edgehog/os_management/os_management.ex
```diff
@@ -0,0 +1,36 @@
...
+    ]
+
+  graphql do
+    root_level_errors? true
+  end
+
+  resources do
+    resource Edgehog.OSManagement.OTAOperation do
+      define :send_update_request, args: [:ota_operation]
+    end
```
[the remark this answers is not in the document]
rbino: I'm still on the fence here. In Ash 2.0, `code_interface` was supported only at the resource level, while Ash 3.0 introduced the possiblity of exposing the code interface in the domain.

Defining at the resource level gives you the advantage of implicit namespacing, i.e. `Domain.Foo.create` and `Domain.Bar.create` instead of `Domain.create_foo` and `Domain.create_bar`.

On the other hand, defining code interfaces on the domain is more similar to Phoenix contexts, moves your API boundary outwards and seems to be the recommended way to go in 3.0.

Anyway, to avoid too much bikeshedding, I think we have to decide how to go with this globally and I've opened an issue to track this: https://github.com/edgehog-device-manager/edgehog/issues/509

I'd leave changes about this to when we take that decision

backend/lib/edgehog/os_management/ota_operation/ota_operation.ex
rbino: Maybe just `finished?`

davidebriani (approved):

Issue #270, opened 2023-03-24
Show the Geolocation Provider that originated the geolocation


rbino:
Edgehog implements Geolocation by trying to query a list of providers (ordered in descending order of accuracy, at least by default) and returning the first valid result. Since some Geolocation providers (e.g. IP based Geolocation) can return a very coarse result, it would be nice to expose informations about the provider which originated the result, to provide more context to the user about the actual accuracy of the data.

PR #511, opened 2024-05-02
Port Geolocation functionalities to Ash

### LINKED ISSUE #499: Port Geolocation to Ash

davidebriani:
The PR adds `location` and `position` calculation fields on the Device resource, which are computed thanks to the business logic contained in the Geolocation context.

The fields also expose some details about the data source that allowed estimating the position/location.

Closes #499, #270


backend/lib/edgehog/astarte/device/geolocation/sensor_position.ex
```diff
@@ -1,7 +1,7 @@
 #
 # This file is part of Edgehog.
 #
-# Copyright 2022 SECO Mind Srl
+# Copyright 2022-24 SECO Mind Srl
```
rbino, line 4: Typo, should be the full year

backend/lib/edgehog/geolocation/providers/device_geolocation.ex
```diff
@@ -21,45 +21,47 @@
...
 
   @impl Edgehog.Geolocation.GeolocationProvider
-  def geolocate(%Device{device_id: device_id} = device) do
-    with {:ok, client} <- Devices.appengine_client_from_device(device),
-         {:ok, sensors_positions} <- Astarte.fetch_geolocation(client, device_id),
-         {:ok, sensors_positions} <- filter_latest_sensors_positions(sensors_positions) do
-      geolocate_sensors(sensors_positions)
+  def geolocate(%Device{} = device) do
+    with {:ok, device} <- Ash.load(device, :tenant),
+         {:ok, device} <- Ash.load(device, [:sensor_positions], tenant: device.tenant),
```
rbino, line 31: Shouldn't you be able to do both of these in a single load?

rbino: Also, I'd use `reuse_values?: true` for this kind of loads, otherwise `load!` re-reads from the database everytime (see motivation [here](https://hexdocs.pm/ash/3.0.0-rc.45/upgrading-to-3-0.html#calculations-do-not-reuse-values-by-default), which shouldn't be relevant in our case).
I'll open an issue to check this in the existing codebase, but let's try to do the correct thing for new code.

backend/lib/edgehog/geolocation/providers/device_geolocation.ex
```diff
@@ -1,7 +1,7 @@
 #
 # This file is part of Edgehog.
 #
-# Copyright 2022 SECO Mind Srl
+# Copyright 2022-24 SECO Mind Srl
```
rbino, line 4: Full year

backend/lib/edgehog/devices/device/calculations/device_location.ex
```diff
@@ -0,0 +1,66 @@
...
+  defp fetch_device_location(device) do
+    with {:ok, position} <- Geolocation.geolocate(device) do
+      coordinates = %Coordinates{
+        latitude: position.latitude,
+        longitude: position.longitude
+      }
+
+      address =
+        case Geolocation.reverse_geocode(coordinates) do
+          {:ok, location} -> location
```
rbino, line 47: `{:ok, address} -> address`

backend/lib/edgehog/devices/device/calculations/device_location.ex
```diff
@@ -0,0 +1,66 @@
...
+          _ -> nil
+        end
+
+      device_location = %DeviceLocation{
```
rbino, line 51: Internal note to be considered before the release: right now we always do both Geolocation and Geocoding. Given we can now easily make calculations depend upon each other, it might be smart to split the two values also at the API level. This way you can ask for Geolocation only (and not "pay" for an API call to the Geocoding provider) or you can ask for the address (which will depend on Geolocation)

backend/lib/edgehog/devices/device/types/device_location.ex
```diff
@@ -0,0 +1,91 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Devices.Device.Types.DeviceLocation do
```
rbino, line 21: Given this is a (embedded) resource I'd move it out of `Types` (which I've created to store `Ash.Type`s, and tbh I probably will rollback)

rbino (changes requested):

backend/lib/edgehog/devices/device/calculations/device_location.ex
davidebriani: Here I am being consistent with the Geocoding providers where `reverse_geocode` returns a `Edgehog.Geolocation.Location`.
Right now `Edgehog.Geolocation.Location` is just a `String.t()`, but it would be nice in the future to expose a structured location instead of a formatted address.

If you feel this overcomplicates things and we should just call all of them *formatted_address*, let me know.

rbino: It just feels strange that something that gets assigned to an `address` variable and the to an `address` field in a struct is internally called `location` (which is always in turn the name given to the whole struct when is returned). It felt like a copy-paste error tbh, I'd keep the naming consistent throughout.

backend/lib/edgehog/geolocation/providers/device_geolocation.ex
davidebriani: The fact is that I need the tenant fist in order to set the `tenant` option in the second `Ash.load`.
If I don't set the tenant option, Ash will complain about the option not being set.

If agree this is not ideal.
Another solution would be to accept a `tenant` argument besides the input `device`: `geolocate(tenant, device)`.
Howvever that feels odd: the `device` should be enough, and why allowing to pass a tenant that is potentially different from the one of the device?

Do you have any insights here?

backend/lib/edgehog/devices/device/calculations/device_location.ex
davidebriani: I had the same thought and it makes sense to me.
That's why I clearly separated the two concerns in the Geolocation context, although I kept the previouus GraphQL API on the outside.

I will experiment with defining two separate calculations to see if we like the APIs.
It could be something like:

```graphql
device {
  position {
    latitude
    longitude
    accuracy
    altitude
    altitudeAccuracy
    heading
    speed
    timestamp
  }
  location {
    formattedAddress
    timestamp
  }
}
```

where in the future we could expand the Location type to also hold the address parts (ZIP code, state, region, city, etc.).

backend/lib/edgehog/geolocation/providers/device_geolocation.ex
rbino: That does seem strange since `Ash.load/3` should [pick up the tenant from the record](https://github.com/ash-project/ash/blob/main/lib/ash.ex#L1680). Do you have a stack trace of how Ash complains?

backend/lib/edgehog/devices/device/calculations/device_location.ex
rbino: Might be a good occasion to maybe incorporate https://github.com/edgehog-device-manager/edgehog/issues/270?

backend/lib/edgehog/geolocation/providers/device_geolocation.ex
davidebriani: The error is
```
{:error,
 %Ash.Error.Invalid{
   path: [:realm], 
   query: "#Query<>", 
   errors: [
     %Ash.Error.Invalid.TenantRequired{
       resource: Edgehog.Astarte.Realm,
       splode: Ash.Error,
       bread_crumbs: [],
       vars: [],
       path: [:realm],
       stacktrace: #Splode.Stacktrace<>,
       class: :invalid
     }
   ]
 }}
```

The device does not have a tenant set.
So that's why I first have to load it with `Ash.load(device, :tenant)`.
However note that I must also set the tenant explicitly in `Ash.load(device, [:sensor_positions], tenant: device.tenant)`, since the device has now a populated `device.tenant` but `device.__metadata__.tenant` is still `nil`: and `Ash.load/3` seems to read from __metadata__ if you don't specify any option.

backend/lib/edgehog/astarte/device/geolocation/sensor_position.ex
davidebriani: Fixed

backend/lib/edgehog/geolocation/providers/device_geolocation.ex
davidebriani: Fixed

backend/lib/edgehog/devices/device/calculations/device_location.ex
davidebriani: This isn't relevant anymore, since I clearly separated a Position struct, which contains GPS coordinates, from a Location struct, which contains a formatted address.

backend/lib/edgehog/devices/device/calculations/device_location.ex
davidebriani: I separated a Position struct, which contains GPS coordinates, from a Location struct, which contains a formatted address.
I added a `source` field on both structs that can contain details about how the fields are estimated, as suggested by #270

backend/lib/edgehog/devices/device/types/device_location.ex
davidebriani: Fixed

backend/lib/edgehog/geolocation/providers/device_geolocation.ex
davidebriani: Basically, for some reason, the calculation provides a list of records that do not have the `tenant` set in `__metadata__`.
We can either try to dig deeper into Ash, or fix it later if we don't come up with a good solution right away.

backend/lib/edgehog/devices/device/calculations/location.ex
```diff
@@ -18,18 +18,29 @@
...
   end
 
   @impl true
-  def reverse_geocode(_invalid_coordinates) do
-    {:error, :address_not_found}
+  def calculate(devices, _opts, _context) do
+    Enum.map(devices, fn device ->
+      if device.position do
+        get_location(device.position)
+      end
```
rbino, line 36: Not a fan of the fact that `if` without `else` implicitly return `nil`.

I'd do:
```elixir
Enum.map(devices, &(get_location(&1.position))

defp get_location(nil), do: nil

# ...
```

rbino (changes requested):

backend/lib/edgehog/devices/device/calculations/location.ex
davidebriani: Not a fan of accepting `nil` as function arguments, but the implicit `nil` for the missing `else` case definitely stinks more, I agree.
It's fixed.

PR #517, opened 2024-05-08
[frontend] standardize the structure of update forms

szakhlypa:
- add missing action buttons and sort them in order of: `Update`, `Reset`, `Delete` 
- track field changes and enable/disable action buttons (`Update` and `Reset` buttons are enabled when form has changes and mutation is not in progress, disabled otherwise)
- fix: do not reset form on first render

First render, `Update` and `Reset` buttons are disabled.
HardwareType_first-render

Form has changes, buttons are active
HardwareType_buttons-active

Update request in progress
HardwareTypes_update-in-progress

### DIFF
--- a/frontend/src/forms/UpdateBaseImage.tsx
+++ b/frontend/src/forms/UpdateBaseImage.tsx
@@ -18,7 +18,7 @@
   SPDX-License-Identifier: Apache-2.0
 */
 
-import React, { useEffect, useMemo } from "react";
+import { useMemo, useState } from "react";
 import { useForm } from "react-hook-form";
 import { FormattedMessage } from "react-intl";
 import { graphql, useFragment } from "react-relay/hooks";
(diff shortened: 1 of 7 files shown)


frontend/src/forms/UpdateBaseImage.tsx
```diff
@@ -267,23 +281,35 @@ const UpdateBaseImage = ({
...
+            <FormattedMessage
+              id="forms.UpdateBaseImage.submitButton"
+              defaultMessage="Update"
+            />
+          </Button>
+          <Button
+            variant="secondary"
+            disabled={!canSubmit}
+            onClick={() => reset()}
+          >
```
davidebriani, line 300: We could define a `canReset` variable to use here: it isn't obvious that one can reset the form when `canSubmit` is true. For instance, if we have some validation on the form fields, it might be that the form is dirty and could be reset, but it cannot be submitted if field validation fails.

Same argument for the other forms.

frontend/src/forms/UpdateSystemModel.tsx
```diff
@@ -386,6 +386,13 @@ const UpdateSystemModelForm = ({
         </Row>
         <div className="d-flex justify-content-end align-items-center">
           <Stack direction="horizontal" gap={3}>
+            <Button variant="primary" type="submit" disabled={isLoading}>
+              {isLoading && <Spinner size="sm" className="me-2" />}
+              <FormattedMessage
+                id="components.UpdateSystemModelForm.submitButton"
+                defaultMessage="Update"
+              />
```
davidebriani, line 394: I saw you're also changing the ID of i18n messages to being with `forms.` instead of `components.`.
I reckon you'd want to do the same here.

davidebriani (changes requested):

davidebriani (approved):

PR #521, opened 2024-05-20
Port Update Channels to Ash

davidebriani:


backend/lib/edgehog/update_campaigns/update_channel/changes/relate_target_groups.ex
```diff
@@ -0,0 +1,58 @@
...
+      DeviceGroup
+      |> Ash.Query.filter(id in ^target_group_ids)
+      |> Ash.Query.filter(is_nil(update_channel_id))
+      |> Ash.Query.set_tenant(tenant)
+      |> Ash.bulk_update(:update_update_channel, %{update_channel_id: update_channel.id},
+        return_records?: true
+      )
+
+    if status == :success and length(records) == length(target_group_ids) do
+      {:ok, update_channel}
```
rbino, line 48: Possible idea (let's open an issue to iterate on this since it's not critical and I don't want this to become a blocker).

Here you're filtering on `is_nil(update_channel)` and then setting the update channel. If the length of records is less then what you expect, then you know some of the target groups already had an update channel.

This works fine but this way you can't say _which_ target groups failed. My idea is that you could use `if` _inside_ an atomic update expression (see [here](https://elixirforum.com/t/conditions-on-atomic-changes/61200/6))

```elixir
atomic_update(:update_channel_id, if is_nil(update_channel_id) do
  # This is the one you're passing
  ^update_channel_id
else
  # This is the existing one
  update_channel_id
end)
```

and then on the returned records you can check if all of them have the correct `update_channel_ids`. If some of them don't, you can indicate to the caller _which_ ones.

backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
```diff
@@ -0,0 +1,218 @@
...
+    an UpdateCampaign.
+    """
+  end
+
+  graphql do
+    type :update_channel
+
+    queries do
+      get :update_channel, :get
+      list :update_channels, :list
```
rbino, line 45: Let's not do this on new resources, this is something I did initially to have different descriptions for the two queries, but now the description can be passed [inline](https://hexdocs.pm/ash_graphql/dsl-ashgraphql-resource.html#graphql-queries-get-description), and `get? true` doesn't really have any effect on GraphQL, so we should just use the default read for both.

Note that the default read paginates by default, so you have to pass `paginate_with: nil` to the `list` query until we handle pagination project-wide.

```elixir
queries do
  get :update_channel, :read do
    description "blabla"
  end
  
  list :update_channels, :list do
    description "blablab"
    paginate_with nil
  end
end

# ...
actions do
  defaults [:read]
end
```

rbino: I've also opened an issue to track this for existing resources

rbino (changes requested):

backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
```diff
@@ -0,0 +1,218 @@
...
+      description """
+      The devices targeted by the update channel that can be updated with the \
+      provided base image.
+      Note that this only checks the compatibility between the device and the \
+      system model targeted by the base image. The starting version \
+      requirement will be checked just before the update and will potentially \
+      result in a failed operation.\
+      """
+
+      public? true
```
rbino, line 190: This is public so it should be exposed on GraphQL. Is this tested somewhere? I'd add a test in `update_channel_test.exs`, with a custom document to just return updatable devices

rbino (changes requested):

backend/lib/edgehog/update_campaigns/update_channel/calculations/updatable_devices.ex
```diff
@@ -0,0 +1,49 @@
...
+  @impl true
+  def calculate(update_channels, _opts, context) do
+    %{arguments: %{base_image: base_image}} = context
+
+    base_image = Ash.load!(base_image, base_image_collection: [:system_model_id])
+
+    system_model_id = base_image.base_image_collection.system_model_id
+
+    update_channels
+    |> Ash.load!(target_groups: [devices: :system_model])
```
rbino, line 37: Why are you not filtering at the database level?

```elixir
devices_query =
  Device
  |> Ash.Query.filter(device_id == ^system_model_id)
  
update_channels
|> Ash.load!(target_groups: [devices: devices_query])
# ...
```

I _think_ this should work because the `Devices` ManualRelationship in Groups [considers the load query](https://github.com/edgehog-device-manager/edgehog/blob/from-ash-we-rise/backend/lib/edgehog/groups/device_group/manual_relationships/devices.ex#L33) and then [applies additional filters](https://github.com/edgehog-device-manager/edgehog/blob/from-ash-we-rise/backend/lib/edgehog/groups/device_group/manual_relationships/devices.ex#L44) on top of it

rbino (changes requested):

backend/lib/edgehog/update_campaigns/update_channel/calculations/updatable_devices.ex
davidebriani: Good point.
Ash complains about `Edgehog.Devices.Device.system_model is not filterable`.
So I would need to dig deeper into why that happens.

backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
davidebriani: Nice catch.
I added the tests to the PR.
But I also made the field private: it would be better to expose device lists with pagination, so I am still not sure that a calculation is the best way to compute this field, a custom read action might be more appropriate. I'll leave to our future selves to figure out if we really need to expose this info in the first place.

backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
davidebriani: Very true.
My only concern is that by adding the documentation on GraphQL queries I would also have to add documentation to other APIs / code interfaces, while I could add the documentation just once directly on actions.

backend/lib/edgehog/update_campaigns/update_channel/changes/relate_target_groups.ex
davidebriani: This is the draft I ended up with, which is still not working.
We can carry on with the current implementation and maybe review this approach in the future.

```elixir
  defp relate_target_groups(tenant, update_channel, target_group_ids) do
    update_channel_id = update_channel.id

    update_expr =
      Ash.Expr.expr(
        if is_nil(update_channel_id) do
          # This is the one you're passing
          ^update_channel_id
        else
          # This is the existing one
          update_channel_id
        end
      )

    %Ash.BulkResult{status: status, records: records} =
      DeviceGroup
      |> Ash.Query.filter(id in ^target_group_ids)
      |> Ash.Query.set_tenant(tenant)
      |> Ash.bulk_update(:update_update_channel, %{},
        atomic_update: %{update_channel_id: update_expr},
        return_records?: true
      )

    group_ids_with_preexisting_channel =
      records
      |> Enum.filter(&(&1.update_channel_id != update_channel_id))
      |> Enum.map(& &1.id)

    group_ids_not_found =
      Enum.filter(target_group_ids, fn id -> not Enum.any?(records, &(&1.id != id)) end)

    cond do
      not Enum.empty?(group_ids_not_found) ->
        {:error,
         Ash.Error.Changes.InvalidArgument.exception(
           field: :target_group_ids,
           message: "some target groups were not found: #{inspect(group_ids_not_found)}"
         )}

      not Enum.empty?(group_ids_with_preexisting_channel) ->
        {:error,
         Ash.Error.Changes.InvalidArgument.exception(
           field: :target_group_ids,
           message:
             "some target groups are already associated with an update channel: #{inspect(group_ids_with_preexisting_channel)}"
         )}

      status != :success or length(records) != length(target_group_ids) ->
        {:error,
         Ash.Error.Changes.InvalidArgument.exception(
           field: :target_group_ids,
           message: "is invalid"
         )}

      true ->
        {:ok, update_channel}
    end
  end
```

backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
davidebriani: As discussed in person, it's better to maintain a single `:read` action so that if we need to support low-level functionalities we would need to only implement them once.

I updated the queries as you suggested.

backend/test/support/fixtures/devices_fixtures.ex
```diff
@@ -89,7 +89,14 @@ defmodule Edgehog.DevicesFixtures do
   Generate a %Devices.Device{} compatible with a specific %BaseImages.BaseImage{}, passed as argument.
   """
   def device_fixture_compatible_with(opts \\ []) do
-    {base_image, opts} = Keyword.pop!(opts, :base_image)
+    {base_image_id, opts} = Keyword.pop!(opts, :base_image_id)
```
rbino, line 92: This was changed but it doesn't get used anywhere, maybe it's from the next PR? I'd like to see it in context to check the reason for the change

rbino (changes requested):

backend/test/support/fixtures/devices_fixtures.ex
davidebriani: I agree, although the fixture is already being used for this PR in [`backend/test/edgehog/update_campaigns/update_channel_test.exs`](https://github.com/edgehog-device-manager/edgehog/blob/3f7296c8fb182c51b04b66965e139071a98a530d/backend/test/edgehog/update_campaigns/update_channel_test.exs#L53).

rbino: Missed that, good to go

PR #524, opened 2024-05-24
Port localized attributes to Ash

rbino:
Expose localized descriptions in SystemModel and localized descriptions and release display names in BaseImage.

Change the localization logic to be more explicit: the caller can pass a list of preferred language tags to filter the available localized attributes, and no implicit filtering (e.g. with the default tenant locale) is done behind the scenes.


backend/lib/edgehog/base_images/base_image/base_image.ex
```diff
@@ -39,6 +40,9 @@ defmodule Edgehog.BaseImages.BaseImage do
   graphql do
     type :base_image
 
+    field_names localized_description: :description,
+                localized_release_display_name: :release_display_name
```
davidebriani, line 44: Are these field_names intentional or a leftover?

davidebriani (changes requested):

davidebriani (approved):

PR #527, opened 2024-05-24
tools: allow creating admin JWT with gen-edgehog-jwt

### LINKED ISSUE #508: Provide a tool to generate admin REST API JWTs
Right now `gen-edgehog-jwt` is only able to emit JWTs for the GraphQL API. We need to provide an easy way to generate tokens for the admin API.

rbino:
Remove `auth-paths` flag since it's not currently used and add `token-type` flag to pass `tenant` or `admin`

Close #508


szakhlypa:
We should update the docs (`doc/pages/tutorials/edgehog_in_5_minutes.md` and `doc/pages/admin/deploying_with_kubernetes.md` files) with new required `-t` (token-type) argument.

szakhlypa (approved):

Issue #267, opened 2023-03-23
Standardize test fixtures


rbino:
With the latest pull requests, we started experimenting with a form of fixtures inspired by [this talk](https://www.youtube.com/watch?v=6sNmJtoKDCo). Namely, all fixtures should accept just a keyword list for attributes, and should implicitly create all their nested associations if they are not passed explicitly. This allow to just focus on the important part of the fixture for the test under hand (e.g. an invalid name) without being surrounded by noise.

Additionally, similar private helper functions that provide reasonable defaults should be created (see e.g. [this](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/test/edgehog/update_campaigns_test.exs#L284)) to, again, reduce noise in tests and allow to concentrate on what's important for the test case.

This format (which can be found, as an example, in [UpdateCampaignsFixtures](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/test/support/fixtures/update_campaigns_fixtures.ex) and [UpdateCampaignsTest](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/test/edgehog/update_campaigns_test.exs)) should be propagated to all existing fixtures, fixing tests as needed.

Note that, as can be seen above, fixtures and private helper functions are actually quite similar, so it's worth evaluating some kind of deduplication strategy to avoid repeating the same code twice.

rbino:
This is being handled as part of the Ash porting

PR #530, opened 2024-05-27
Port Update Campaigns to Ash

### LINKED ISSUE #343: Change rollout defaults for `ota_request_retries` and `ota_request_timeout`
Right now, the Edgehog Device Runtime has an MQTT keepalive interval of 120 seconds (the Rust one has a keepalive of 30 seconds). This means that if a device disconnects, it can take up to 180 seconds (1.5 * keepalive) for the broker to mark the device as offline.

Since by default we're setting 0 retries and 60 seconds as OTA Request timeout, this makes it very easy for an Update Target to fail, because the Executor could see it as online when in fact it has just disconnected and with 0 retries the Update Target would immediately be marked as failed. A more sensible default would probably be 3 retries with 300 seconds timeout, to cover the uncertainty window given by the keepalive. Users would still be free to change it, but at least the default would be less prone to failure.

davidebriani:
Closes #343


szakhlypa (changes requested):

backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/core.ex
```diff
@@ -114,32 +127,34 @@ defmodule Edgehog.UpdateCampaigns.PushRollout.Core do
...
   end
 
   @doc """
-  Increases the retry count for the target and saves it to the Repo.
+  Increases the retry count for the target and saves it to the data layer.
   """
   def increase_retry_count!(target) do
-    Ecto.Changeset.change(target, retry_count: target.retry_count + 1)
-    |> Repo.update!()
+    Ash.update!(target, %{retry_count: target.retry_count + 1})
```
rbino, line 157:
```elixir
Ash.update!(target, action: :increase_retry_count)
```
i.e. let's have specific update actions that more or less map to the various updates we do from Core. In some cases (like this one) you don't even need attributes and you can make them atomic by increasing the retry count at the database level with `atomic_update` (which shouldn't be a problem in our case since only a single process updates the target at a time, but why not)

backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/executor.ex
```diff
@@ -138,13 +149,13 @@ defmodule Edgehog.UpdateCampaigns.PushRollout.Executor do
     }
 
     timeout_actions =
-      Core.list_targets_with_pending_ota_operation(update_campaign_id)
+      Core.list_targets_with_pending_ota_operation(tenant_id, update_campaign_id)
       |> Enum.map(fn pending_target ->
         # Side effect: receive updates for the OTA Operation so we can track it
-        Core.subscribe_to_ota_operation_updates!(pending_target.ota_operation_id)
+        Core.subscribe_to_ota_operation_updates!(pending_target.ota_operation.id)
```
rbino, line 155: It's the same, but `target` should still have an `ota_operation_id` attribute (generated from the relationship)

backend/test/edgehog/update_campaigns/resumer/core_test.exs
```diff
@@ -25,53 +25,63 @@ defmodule Edgehog.UpdateCampaigns.Resumer.CoreTest do
...
+  @moduletag :ported_to_ash
+
   describe "stream_resumable_update_campaigns/0" do
     test "returns an empty stream if no UpdateCampaigns are present" do
       assert [] = Core.stream_resumable_update_campaigns() |> stream_to_list()
     end
 
     test "returns an empty stream if terminated UpdateCampaigns are present" do
-      _update_campaign = update_campaign_fixture()
+      tenant = tenant_fixture()
```
rbino, line 40: Nit: maybe just put the tenant creation in the setup

backend/lib/edgehog/update_campaigns/executor_supervisor.ex
```diff
@@ -36,9 +34,12 @@ defmodule Edgehog.UpdateCampaigns.ExecutorSupervisor do
   end
 
   def start_executor!(update_campaign) do
+    update_campaign =
+      Ash.load!(update_campaign, [:id, :rollout_mechanism, :tenant_id], reuse_values?: true)
```
rbino, line 38: This should be fixed by the dedicated action (see other comment), but I'd expect `:id` and `:tenant_id` to always be loaded since they are attributes

backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/push_rollout.ex
```diff
@@ -0,0 +1,112 @@
...
+      description """
+      The timeout (in seconds) Edgehog has to wait before considering an OTA \
+      Request lost (and possibly retry). It must be at least 30 seconds.
+      """
+
+      public? true
+      allow_nil? false
+      default 60
+      constraints min: 30
+    end
```
rbino, line 110: While we're changing this, let's deal with #343 (those defaults should be already pushed from the frontend IIRC)

rbino (changes requested):
Great (and massive) work.
The main comments have been on using more semantic actions instead of manually using `Ash.Query` etc in the `Core` module.
We don't have to do everything now, I'd go for the low hanging fruits and then iterate. The idea is that if we can encode stuff in the action, the action can then become a code interface, so the endgame would be calling `UpdateCampaigns.increase_retry_count_for_target(target, tenant: tenant)`.

So let's try to do step 1 towards that (which is trying to keep semantics inside actions) and see where we go from there.

rbino:
@davidebriani given I think we've now touched all available fixtures, check if #267 can be considered closed and if so add it to the "Closes" list of your PR

backend/lib/edgehog/os_management/ota_operation/ota_operation.ex
```diff
@@ -90,6 +99,18 @@ defmodule Edgehog.OSManagement.OTAOperation do
...
+      change {Edgehog.Changes.PublishNotification, name: :ota_operation_created}
+    end
+
+    update :update do
+      primary? true
+
+      accept [:status, :status_progress, :status_code, :message]
+
+      # Needed because Edgehog.Changes.PublishNotification is not atomic
+      require_atomic? false
```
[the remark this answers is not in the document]
davidebriani: Because Ash

> cannot atomically run a changeset with an after_transaction hook

At least that's the error I'm getting if I don't specify `require_atomic? false`.

davidebriani:
Issue #267 is almost done, but a few more changes are required and it's better to include them in a separate PR.

backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/core.ex
```diff
@@ -114,32 +127,34 @@ defmodule Edgehog.UpdateCampaigns.PushRollout.Core do
   @doc """
   Resends an OTARequest on Astarte for an existing OTA Operation
   """
-  def retry_target_update(target, base_image) do
-    OSManagement.send_update_request(target.device, target.ota_operation_id, base_image.url)
+  def retry_target_update(target) do
+    target = Ash.load!(target, :ota_operation)
```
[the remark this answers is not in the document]
davidebriani: Right!
Although I would also need to specify the tenant:

```elixir
OTAOperation
|> Ash.get!(target.ota_operation_id, tenant: target.tenant_id)
|> OSManagement.send_update_request()
```

So I am trying this version to make Ash respect the tenant automatically:
```elixir
target
|> Ash.load!(:ota_operation)
|> Map.fetch!(:ota_operation)
|> OSManagement.send_update_request()
```

Another option would be the following, although I find it less readable:
```elixir
target
|> Ash.load!(:ota_operation)
|> then(& OSManagement.send_update_request(&1.ota_operation))
```

backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/core.ex
```diff
@@ -18,30 +18,36 @@
...
   # structure will emerge from the requirements.
 
   @doc """
-  Gets an UpdateCampaign. Differently from `UpdateCampaigns.fetch_update_campaign/1`, this
-  doesn't preload associations, since they're not needed here.
+  Lists the default associations that should be preloaded for an update campaign.
+  """
+  def default_preloads_for_update_campaign do
+    [:base_image]
+  end
```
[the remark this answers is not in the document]
davidebriani: There is now a single place where I need to fetch the base image from an update campaign's id.
So I just manually loaded the relationship instead of defining a `for_rollout` read action, which didn't suit my case properly anymore.

backend/lib/edgehog/update_campaigns/update_campaign/update_campaign.ex
```diff
@@ -0,0 +1,253 @@
...
+
+      accept [:status, :outcome, :start_timestamp, :completion_timestamp]
+    end
+
+    update :mark_as_in_progress do
+      # TODO: Needed otherwise set_new_attribute/2 is called before accepted attributes
+      require_atomic? false
+
+      accept [:start_timestamp]
+      change set_new_attribute(:start_timestamp, &DateTime.utc_now/0)
```
rbino, line 115:
```elixir
argument :start_timestamp, :utc_datetime_usec do
  default &DateTime.utc_now/0
end
change set_attribute(:start_timestamp, arg(:start_timestamp))
```

rbino (changes requested):

backend/lib/edgehog/update_campaigns/update_campaign/update_campaign.ex
davidebriani: Fixed, thanks

rbino (approved):

PR #534, opened 2024-06-05
Tests: remove redundant tests

davidebriani:
These tests verify functionalities already tested by other tests, i.e. those on Triggers.


backend/test/edgehog/astarte_test.exs
```diff
@@ -77,131 +77,6 @@ defmodule Edgehog.AstarteTest do
...
-      device = astarte_device_fixture(realm)
-      device_id = device.device_id
-      event = %{"type" => "unknown"}
-      timestamp = DateTime.utc_now() |> DateTime.to_iso8601()
-      assert :ok = Astarte.process_device_event(realm, device_id, event, timestamp)
-
-      assert ^device = Astarte.get_device!(device.id)
-    end
-
-    test "process_device_event/4 updates online and last_connection on device_connected event", %{
```
Pavinati, line 91: we have a test for creating a new device from a connection event, but we're missing a test where we update the device with data from a connection event

davidebriani: I'm guessing Riccardo just implemented the case for disconnection events and avoided the case for connection events since we could expect them to be practically equal.

https://github.com/edgehog-device-manager/edgehog/blob/ac1ce857c9eba80379ebe83787049fbc150f6eda/backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs#L147C14-L147C53

However, better safe than sorry, and adding a new test is not a great endeavor. So I will do just that.

Pavinati: Considering they are different events and contain different data (for example the connection event also contains the device public IP) i would add a specific test.
It should be pretty straight forward to do anyway

davidebriani: Done, I split the test I referenced above into two cases: for connection and disconnection triggers.

davidebriani: Note that the device IP in connection triggers is currently unused: we don't read it or persist it anywhere in Edgehog.

PR #536, opened 2024-06-10
Migrate set let behavior

Pavinati:
Migrate the "set LED behavior" to Ash using manual updates.
Ash updates are used because this mutation updated the status of a device. The only difference from a standard update is that this `state` is not persisted in the Edgehog data layer (the DB) but rather in Astarte.


backend/lib/edgehog/devices/device/types/led_behavior.ex
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Devices.Device.Types.LedBehavior do
```
davidebriani, line 21: Riccardo initially defined a `Types` folder for Astarte values of devices, however we eventually avoided this approach for further developments.
It's probably not necessary to indicate that this file contains a *type*, distinguishing it from other fields that might be implemented as embedded resources or in other ways.

We could just keep `led_behavior.ex` next to `device.ex` for now.
 
```suggestion
defmodule Edgehog.Devices.Device.LedBehavior do
```

backend/lib/edgehog/devices/device/types/led_behavior.ex
```diff
@@ -0,0 +1,30 @@
...
+  use Ash.Type.Enum,
+    values: [
+      blink: "Blink for 60 seconds.",
+      double_blink: "Double blink for 60 seconds.",
+      slow_blink: "Slow blink for 60 seconds."
+    ]
+
+  def graphql_type(_), do: :device_blink_behavior
```
davidebriani, line 29:
```suggestion
  def graphql_type(_), do: :device_led_behavior
```

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
```diff
@@ -19,63 +19,134 @@
...
-    setup do
-      device =
-        cluster_fixture()
-        |> realm_fixture()
-        |> device_fixture()
+  describe "SetLedBehavior mutation" do
+    test "sets behavior with valid data", %{tenant: tenant} do
+      device = device_fixture(tenant: tenant)
+      id = AshGraphql.Resource.encode_relay_id(device)
+      astarte_device_id = device.device_id
```
davidebriani, line 33: Let's move this ceremony to a `setup` block, so tests can focus on the actual logic being tested.

Tests can expect already populated variables: `tenant`, `id`, `astarte_device_id`.

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
```diff
@@ -19,63 +19,134 @@
...
+
+      Edgehog.Astarte.Device.LedBehaviorMock
+      |> expect(:post, 1, fn _client, ^astarte_device_id, _behavior -> :ok end)
+
+      result = set_led_behavior_mutation(tenant: tenant, id: id, behavior: "BLINK")
+      assert %{"id" => ^id} = extract_result!(result)
```
davidebriani, line 39: Let's also match the exact behavior that is sent to Astarte, to ensure the mutation actually works as intended.

This test could then be renamed `sets a BLINK behavior`, since this test is specific on the BLINK behavior.

We can add similar changes to the following two tests: i.e. for DOUBLE_BLINK and SLOW_BLINK.

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
```diff
@@ -19,63 +19,134 @@
...
+
+    test "does not send data to Astarte with invalid data", %{tenant: tenant} do
+      device = device_fixture(tenant: tenant)
+      id = AshGraphql.Resource.encode_relay_id(device)
+      astarte_device_id = device.device_id
+
+      Edgehog.Astarte.Device.LedBehaviorMock
+      |> expect(:post, 0, fn _client, _device_id, _behavior -> raise "Error!" end)
+
+      assert %{errors: [%{message: msg}]} =
```
davidebriani, line 90: `msg` is unused

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
```diff
@@ -19,63 +19,134 @@
...
+  defp set_led_behavior_mutation(opts) do
+    default_document = """
+    mutation SetLedBehavior($id: ID!, $input: SetLedBehaviorInput!) {
+      setLedBehavior(id: $id, input: $input) {
+        result {
+          id
+        }
+        errors {
+          message
+        }
```
davidebriani, line 104: We are usually have `root_level_errors`, so this field should be empty and you can avoid querying it.
The errors you are getting from the mutation should actually be the root level ones.

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
```diff
@@ -19,63 +19,134 @@
...
+  defp extract_error!(result) do
+    assert %{errors: [error]} = result
 
-      assert %{
-               "data" => %{
-                 "setLedBehavior" => %{
-                   "behavior" => "BLINK"
-                 }
+    error
+  end
```
davidebriani, line 126: I would add a check on the data as well

```elixir
defp extract_error!(result) do
  assert %{
           data: %{"setLedBehavior" => nil},
           errors: [error]
         } = result

  error
end
```

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
```diff
@@ -19,63 +19,134 @@
 #
 
 defmodule EdgehogWeb.Schema.Mutation.SetLedBehaviorTest do
-  use EdgehogWeb.ConnCase, async: true
-  use Edgehog.AstarteMockCase
+  use EdgehogWeb.GraphqlCase, async: true
+
+  @moduletag :ported_to_ash
 
   import Edgehog.AstarteFixtures
```
davidebriani, line 26: Is this still needed?

davidebriani (changes requested):

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -53,6 +54,7 @@ defmodule Edgehog.Devices.Device do
       update :update_device, :update
       update :add_device_tags, :add_tags
       update :remove_device_tags, :remove_tags
+      update :set_led_behavior, :set_led_behavior
```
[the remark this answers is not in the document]
Pavinati: i've kept the same name as before to minimize the API changes. I do agree with you here, if we can change the mutation name i'll update the PR accordingly

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
Pavinati: GraphQL errors do not return `data`. For example when the we use an invalid behavior

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
Pavinati: It's not used anymore. I'll remove it and convert the other to an `alias` to avoid confusion

backend/test/edgehog_web/schema/mutation/set_led_behavior_test.exs
Pavinati: In all the test we don't really need the whole device. I created an utility function for tests that require an Astarte device id and the corresponding device GraphQL id

PR #537, opened 2024-06-11
Port tests on OTA operations to Ash

davidebriani:
The PR converts the old tests that were run on the function of the OSManagement context and run them on the available Ash actions.

It also adds a missing piece of logic on OTA operations before porting the relative tests: cleaning up the ephemeral image of manual OTA operation once they reach a terminal status and the rollout is ended.

### DIFF
--- a/backend/lib/edgehog/os_management/ota_operation/changes/handle_ephemeral_image_deletion.ex
+++ b/backend/lib/edgehog/os_management/ota_operation/changes/handle_ephemeral_image_deletion.ex
@@ -1,0 +1,51 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.OSManagement.OTAOperation.Changes.HandleEphemeralImageDeletion do
+  use Ash.Resource.Change
+
+  alias Edgehog.OSManagement.EphemeralImage
+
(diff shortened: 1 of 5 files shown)


backend/test/edgehog/os_management_test.exs
```diff
@@ -189,80 +71,91 @@ defmodule Edgehog.OSManagementTest do
...
     } do
       device_id = device.device_id
-      ota_request_id = "715cb3d6-a24d-4cb5-bc86-5c85f5a906f2"
-      image_url = "https://my.bucket.com/my_image.bin"
+      ota_operation_id = ota_operation.id
+      base_image_url = ota_operation.base_image_url
 
       Edgehog.Astarte.Device.OTARequestV1Mock
-      |> expect(:update, fn _client, ^device_id, ^ota_request_id, ^image_url ->
+      |> expect(:update, fn _client, ^device_id, ^ota_operation_id, ^base_image_url ->
```
Pavinati, line 125: here we're testing against a specific OTA operation ID and base image URL, i would expect those components to be parts of the test itself instead of relying to external initialization
```elixir
ota_operation = manual_ota_operation_fixture(tenant: tenant,
  device_id: "mspcBscfW7KaYjpy7GAO7A",
  operation_id: "some-id",
  base_image_url: "some-url"
)

Edgehog.Astarte.Device.OTARequestV1Mock
|> expect(:update, fn _client, "mspcBscfW7KaYjpy7GAO7A", "some-id", "some-url" -> ... end)
```

davidebriani: Sounds right, I moved the setup for the OTA operation within the test.
Actually, there's no ota operation in the `setup` block anymore: each test creates one when needed.

Pavinati (approved):

szakhlypa (changes requested):

backend/lib/edgehog/os_management/ota_operation/changes/handle_ephemeral_image_deletion.ex
```diff
@@ -0,0 +1,47 @@
...
+                            :edgehog,
+                            :os_management_ephemeral_image_module,
+                            EphemeralImage
+                          )
+
+  @impl true
+  def change(%Ash.Changeset{valid?: false} = changeset, _opts, _context), do: changeset
+
+  def change(changeset, _opts, _context) do
+    Ash.Changeset.after_action(changeset, &cleanup_base_image_url/2)
```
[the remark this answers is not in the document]
davidebriani: It makes sense, I changed it to `after_transaction/3`.

PR #541, opened 2024-06-19
Move code interfaces from resources to domains

### LINKED ISSUE #509: Move all Code Interfaces on Domains
Ash 3.0 introduced the possibility of defining code interfaces at the domain level.

We need to decide if we want to have `Domain.Resource.create!` or `Domain.create_resource!` in our codebase.

Pavinati:
Use Ash domains as the default module to define code interfaces

Closes #509

### DIFF
--- a/backend/lib/edgehog/astarte/astarte.ex
+++ b/backend/lib/edgehog/astarte/astarte.ex
@@ -1,7 +1,7 @@
 #
 # This file is part of Edgehog.
 #
-# Copyright 2023 SECO Mind Srl
+# Copyright 2023-2024 SECO Mind Srl
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


backend/lib/edgehog/astarte/astarte.ex
```diff
@@ -22,7 +22,14 @@ defmodule Edgehog.Astarte do
...
 
   resources do
-    resource Edgehog.Astarte.Cluster
-    resource Edgehog.Astarte.Realm
+    resource Edgehog.Astarte.Cluster do
+      define :create_cluster, action: :create
+    end
+
+    resource Edgehog.Astarte.Realm do
+      define :fetch_realm_by_name, action: :by_name, args: [:name]
```
davidebriani, line 30: Given that we're touching this definition, you should also be able to delete the `:by_name` action on the resource and rely on the default `:read` instead.
In addition, I would set `not_found_error?: true`, also for uniformity with the code interfaces already defined in other domains.

```suggestion
      define :fetch_realm_by_name, action: :read, get_by: :name, not_found_error?: true
```

backend/lib/edgehog/tenants/tenants.ex
```diff
@@ -36,6 +36,12 @@ defmodule Edgehog.Tenants do
   end
 
   resources do
-    resource Edgehog.Tenants.Tenant
+    resource Edgehog.Tenants.Tenant do
+      define :create_tenant, action: :create
+      define :provision_tenant, action: :provision
+      define :fetch_tenant_by_slug, action: :by_slug, args: [:slug]
```
davidebriani, line 42: Same here, you should be able to delete the `by_slug` action and rely on the default `read`.

```suggestion
      define :fetch_tenant_by_slug, action: :read, get_by: :slug, not_found_error?: true
```

davidebriani (changes requested):

davidebriani (approved):

PR #542, opened 2024-06-21
Use Relay connections for listing resources and relationships

### LINKED ISSUE #522: Remove get/list reads and just use the default read
This is something we initially did to have different descriptions for the two queries, but now the description can be passed [inline](https://hexdocs.pm/ash_graphql/dsl-ashgraphql-resource.html#graphql-queries-get-description), and `get? true` doesn't really have any effect on GraphQL, so we should just use the default read for both.

Note that the default read paginates by default, so we have to pass paginate_with: nil to the list query until we handle pagination project-wide.

```elixir
queries do
  get :resource, :read do
    description "blabla"
  end
  
  list :resources, :read do
    description "blablab"
    paginate_with nil
  end
end

# ...
actions do
  defaults [:read]
end
```

davidebriani:
This PR paginates the "list" GraphQL queries and relationships using Relay connections.

All GraphQL query definitions are updated by specifying that pagination and Relay connections should be used.
The same approach is used to expose the relationships between resources.

At the same time, the default `read` action is preferred as the backing bone of the functionality, removing the distinction of "get" and "list" actions: this distinction only becomes relevant in code interfaces or GraphQL APIs.

Close #522

Note: the relationships between devices and groups are left unpaginated: since it's a manual relationship based on the dynamic nature of group selectors, supporting paginated queries is not straightforward and will be evaluated in a future PR.


davidebriani:
I'm reverting the PR to a draft to highlight that we might want to avoid merging it now.

We can instead focus on getting both the backend and the frontend to a working status without pagination, and later incrementally introduce pagination with PRs that encompass both of them at the same time.
This will allow to keep the development branch stable and introduce other features without waiting for pagination to be fully implemented.

PR #546, opened 2024-07-01
Handle device tags with the new mutations

### LINKED ISSUE #545: Update Device page to use new GraphQL mutations for tags
Adding/removing device tags should be done with `addDeviceTags`/`removeDeviceTags` mutations in latest GraphQL schema.

Pavinati:
use addDeviceTags and removeDeviceTags mutatations to edit tags assigned to a device.

Closes #545

### DIFF
--- a/frontend/src/pages/Device.tsx
+++ b/frontend/src/pages/Device.tsx
@@ -38,7 +38,7 @@ import {
   useRelayEnvironment,
 } from "react-relay/hooks";
 import type { PreloadedQuery } from "react-relay/hooks";
-import type { Subscription } from "relay-runtime";
+import type { PayloadError, Subscription } from "relay-runtime";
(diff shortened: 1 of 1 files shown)


frontend/src/pages/Device.tsx
```diff
@@ -298,10 +301,41 @@ const UPDATE_DEVICE_MUTATION = graphql`
...
+        id
         tags {
+          id
           name
         }
-        deviceGroups {
+      }
+    }
+  }
+`;
```
davidebriani, line 328: The new mutations to add or remove tags may also modify the groups that the device belongs to: I believe that was the reason why `deviceGroups` was queried in the response for the `updateDevice` mutation.

We should query `deviceGroups` on the responses for the new mutations too.

davidebriani (approved):

davidebriani (approved):

Issue #266, opened 2023-03-23
Implement update canceling


bettio:
This will allow to (tentatively) cancel an update that is in progress on a device using the GraphQL API

Issue #225, opened 2022-11-24
Allow canceling pending OTAOperations


rbino:
Sometimes an `OTAOperation` can remain in a pending state (e.g. the response is not sent from the device, or the trigger gets lost for any reason) and currently the frontend disallows creating a new `OTAOperation` if a pending one si present, thus leaving users in an unrecoverable state without manual intervention.

Edgehog should instead expose a way to forcefully terminate a pending `OTAOperation`. Ideally, after the first "manual" implementation, this could also be handled automatically with a (possibly optional) timeout.

PR #547, opened 2024-07-02
Fix or remove all missing tests, remove `ported_to_ash` tag

rbino:
Restore the CI


davidebriani (approved):
The PR looks good for its scope and I'm merging it.

The PR also removes tests on Astarte features that raise some points:
- Requests to cancel OTA updates on `io.edgehog.devicemanager.OTARequest` v1.0 are not implemented or tested anymore. We may want to open an issue now and reintroduce this in a future PR.
- We don't verify whether a device supports `io.edgehog.devicemanager.OTARequest` v1.0 anymore. This is just optimization but we might open an issue and reintroduce it in a future PR.
- Requests to `io.edgehog.devicemanager.OTARequest` v0.1 are not implemented or tested anymore. I guess this is expected and we plan to not maintain it anymore.

rbino:
Just for reference:
- OTA Operation canceling is tracked by #225 and #266 
- 2 and 3 are handled by the fact that we drop support for v0.1, so we assume v1.0

PR #553, opened 2024-07-03
Remove unused code and fix warnings

davidebriani:
- Remove `astarte/device.ex` file since it is unused.
- Fix all compilation warnings about unused functions, aliases and variables.
- Fix all warnings emitted by dialyzer.
- Restore CI checks on compilation and dialyzer warnings.


backend/lib/edgehog/devices/device/manual_relationships/system_model.ex
```diff
@@ -36,15 +35,16 @@ defmodule Edgehog.Devices.Device.ManualRelationships.SystemModel do
...
-    device_id_to_system_model =
+    device_id_to_system_models =
       related_system_models
       |> Enum.flat_map(fn system_model ->
         system_model.part_numbers
         |> Enum.flat_map(fn part_number ->
           Enum.map(part_number.devices, &{&1.id, system_model})
         end)
       end)
+      |> Enum.group_by(&elem(&1, 0), &elem(&1, 1))
```
rbino, line 46: Shouldn't you use `Map.new/1` here? I think that this expects a map as return type

davidebriani: `Enum.group_by/3` indeed returns a map.

PR #564, opened 2024-07-08
Try a fix for a recurrently flaky test

rbino:
Increase the timeout before it flunks


rbino:
It doesn't actually fix the flakyness so let's close this

PR #571, opened 2024-07-08
Specify Postgres version requirement in docs

davidebriani:
- When detailing the prerequisites for deploying an Edgehog instance on the documentation, specify that only Postgres v13+ is supported.
- Update development tooling to use the latest Postgres version, v16, and update the CI workflows to test all supported and advertised versions for production environments.

### DIFF
--- a/.github/workflows/backend-test.yaml
+++ b/.github/workflows/backend-test.yaml
@@ -125,10 +125,13 @@ jobs:
 
   test-coverage:
     name: Build and Test
+    strategy:
+      matrix:
+        postgres: [13, 14, 15, 16]
     runs-on: ubuntu-22.04
     services:
       postgres:
-        image: postgres:14
+        image: postgres:${{ matrix.postgres }}
         env:
           POSTGRES_PASSWORD: postgres
         ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -21,7 +21,7 @@
 version: "3.8"
 services:
   postgresql:
-    image: postgres:14.0
+    image: postgres:16.3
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -21,7 +21,7 @@
 version: "3.8"
 services:
   postgresql:
-    image: postgres:14.0
+    image: postgres:13.0
```
rbino, line 24: I think for local development/testing we want to use the most recent thing we have.
Put `postgres:16` (mind the missing `.0`)

.github/workflows/backend-test.yaml
```diff
@@ -128,7 +128,7 @@ jobs:
     runs-on: ubuntu-22.04
     services:
       postgres:
-        image: postgres:14
+        image: postgres:13
```
rbino, line 131: I think we're better off with a test matrix here, if we want to do things properly. This way we can test we don't regress for older versions _and_ keep checking newer versions so we don't find out too late we have some incompatibilities. Let's test postgres 13, 14, 15 and 16

davidebriani: Added the matrix

docker-compose.yml
davidebriani: I left the minor version specified (`16.3`) since we're not just using this for development but also for the [Edgehog in 5 minutes guide](https://docs.edgehog.io/0.8/edgehog_in_5_minutes.html): we probably don't want the guide to break if a new incongruent Postgres minor gets released.

rbino (approved):

PR #573, opened 2024-07-09
Fix: wrong input values used when creating a base image

### LINKED ISSUE #574: Align Base Image creation page with new GraphQL Schema
The Base Image creation page is still using the old localized fields, preventing a Base Image creation since the backend returns an error.
It should switch to using `localizedDescriptions` and `localizedReleaseDisplayNames`.

davidebriani:
Some input fields were not updated during the migration to the newest version of the GraphQL API.
This change fixes how localized fields are provided in the mutation when creating a base image.

Fixes #574

### DIFF
--- a/frontend/src/forms/CreateBaseImage.tsx
+++ b/frontend/src/forms/CreateBaseImage.tsx
@@ -80,14 +80,14 @@ type BaseImageData = {
   file: File;
   version: string;
   startingVersionRequirement: string;
-  releaseDisplayName: {
-    locale: string;
-    text: string;
-  };
-  description: {
-    locale: string;
-    text: string;
-  };
+  localizedReleaseDisplayNames?: {
+    languageTag: string;
+    value: string;
+  }[];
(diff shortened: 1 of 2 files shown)


frontend/src/forms/CreateBaseImage.tsx
```diff
@@ -129,20 +129,31 @@ const transformOutputData = (
...
-    text: data.description,
-  },
-});
+): BaseImageData => {
+  const baseImage: BaseImageData = {
+    baseImageCollectionId: baseImageCollection.id,
+    file: data.file[0],
+    version: data.version,
+    startingVersionRequirement: data.startingVersionRequirement,
+    localizedReleaseDisplayNames: [
```
rbino, line 138: If I understand the logic below, this should be behind an `if` too because it's optional

davidebriani: Makes sense, fixed it.

PR #577, opened 2024-07-09
Fix: OTA operation success event not being handled

davidebriani:
The code crashed in accessing the `device` field of the OTA operation while trying to handle the event about the OTA operation.

The change fixes the issue by ensuring that the device field is preloaded on the OTA operation resource before trying to access it.

### DIFF
--- a/backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/executor.ex
+++ b/backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/executor.ex
@@ -398,6 +398,7 @@ defmodule Edgehog.UpdateCampaigns.RolloutMechanism.PushRollout.Executor do
           [internal_event({:ota_operation_failure, ota_operation})]
 
         Core.ota_operation_acknowledged?(ota_operation) ->
+          ota_operation = Ash.load!(ota_operation, :device)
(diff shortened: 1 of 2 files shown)


backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/executor.ex
```diff
@@ -398,6 +398,7 @@ defmodule Edgehog.UpdateCampaigns.RolloutMechanism.PushRollout.Executor do
           [internal_event({:ota_operation_failure, ota_operation})]
 
         Core.ota_operation_acknowledged?(ota_operation) ->
+          ota_operation = Ash.load!(ota_operation, :device)
```
rbino, line 401: Shouldn't we just load once at the beginning of the function so that `ota_operation` alway has the `device` when we push it downward (in `internal_event` and here)?

rbino: On a second thought, it's better this way so that loads for independent codepath are already independent and the load is closer to where it's used (they are mutually exclusive so `:device` gets loaded once anyways)

rbino (approved):

PR #585, opened 2024-08-14
Support redirection to a specific page

Emina-M:
Support redirection to a specific page after successful authentication. For example, if the user requested `Update Campaigns` page in the url, instead of going to `Devices` page after authentication, go to `Update Campaigns`.

### DIFF
--- a/frontend/src/App.tsx
+++ b/frontend/src/App.tsx
@@ -1,7 +1,7 @@
 /*
   This file is part of Edgehog.
 
-  Copyright 2021-2023 SECO Mind Srl
+  Copyright 2021-2024 SECO Mind Srl
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -49,6 +49,7 @@ import UpdateCampaignCreate from "pages/UpdateCampaignCreate";
 import UpdateCampaigns from "pages/UpdateCampaigns";
 import Login from "pages/Login";
 import Logout from "pages/Logout";
+import AttemptLogin from "pages/AttemptLogin";
 
 import { version, repository, bugs } from "../package.json";
 
@@ -58,11 +59,13 @@ type RouterRule = {
 };
 
 const publicRoutes: RouterRule[] = [
+  { path: Route.auth, element: <AttemptLogin /> },
   { path: Route.login, element: <Login /> },
   { path: "*", element: <Navigate to={Route.login} /> },
 ];
 
 const authenticatedRoutes: RouterRule[] = [
+  { path: Route.auth, element: <AttemptLogin /> },
   { path: Route.devices, element: <Devices /> },
   { path: Route.devicesEdit, element: <Device /> },
   { path: Route.deviceGroups, element: <DeviceGroups /> },
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.9.0-dev] - Unreleased
 ### Added
 - Allow generating admin JWT using `gen-edgehog-jwt`.
+- Support redirection to a specific page after successful authentication
 ### Changed
 - Change logo and brand images with the latest brand revision.
 
--- a/frontend/src/Navigation.tsx
+++ b/frontend/src/Navigation.tsx
@@ -51,6 +51,7 @@ enum Route {
   updateCampaigns = "/update-campaigns",
   updateCampaignsNew = "/update-campaigns/new",
   updateCampaignsEdit = "/update-campaigns/:updateCampaignId",
+  auth = "/auth",
   login = "/login",
   logout = "/logout",
 }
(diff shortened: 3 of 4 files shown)


frontend/src/contexts/Auth.tsx
```diff
@@ -56,7 +56,7 @@ function saveAuthConfig(
   const cookieOptions = {
     secure: window.location.protocol === "https:",
     expires: persistConfig ? 365 : undefined,
-    sameSite: "strict",
+    sameSite: "none",
```
Pavinati, line 59: This doesn't feel related to this PR

frontend/src/Router.tsx
```diff
@@ -0,0 +1,110 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2021-2023 SECO Mind Srl
```
Pavinati: newly created components should report the year of their creation

frontend/src/Router.tsx
```diff
@@ -0,0 +1,110 @@
...
+  { path: Route.updateCampaignsEdit, element: <UpdateCampaign /> },
+  { path: Route.logout, element: <Logout /> },
+  { path: "*", element: <Navigate to={Route.devices} /> },
+];
+
+export default ({
+  redirectTo,
+}: {
+  redirectTo: string;
+}): React.ReactElement | undefined => {
```
Pavinati, line 95: This file adds a lot of noise to the PR, just to accept a single `redirectTo` parameter.
We could just define this component inside the `App.tsx` file, keeping the diff nice and clean

Pavinati (changes requested):

frontend/src/App.tsx
```diff
@@ -92,10 +99,49 @@ const authenticatedRoutes: RouterRule[] = [
...
+  useEffect(() => {
+    if (auth.isAuthenticated) {
+      redirectTo ? navigate(redirectTo as Route) : navigate(Route.devices);
+    } else {
+      navigate(Route.login);
+    }
+  }, []);
+
+  return <>{RouterElement}</>;
+};
```
Pavinati, line 121: We don't even need this component. We can leave everything as it was in `App`, and just `navigate` if the user completed the login.

```javascript
      auth.login(authConfig, false).then((isValidLogin) => {
        setAttemptingLogin(false);
        if (isValidLogin) {
          navigate(redirectTo)
        }
      });
```

Pavinati (changes requested):

Pavinati (approved):

frontend/src/App.tsx
```diff
@@ -96,6 +103,39 @@ function App() {
   const auth = useAuth();
   const routes = auth.isAuthenticated ? authenticatedRoutes : publicRoutes;
   const RouterElement = useRoutes(routes);
+  const { search } = useLocation();
+  const searchParams = new URLSearchParams(search);
+  const tenantSlug = searchParams.get("tenantSlug") || "";
+  const authToken = searchParams.get("authToken") || "";
+  const redirectTo = searchParams.get("redirectTo") || "";
```
davidebriani, line 110: Adding the logic in the `App.tsx` file means that I can use any URL to attempt the login, it just has to include the needed search params.
For instance `localhost:3000/hardware-types?tenantSlug=<tenant-slug>&authToken=<auth-token>` works, but it will first try to load the page of the Hardware Types, even tough it will not be displayed to the user after the login attempt: either the Devices or the Login page will be shown depending on whether the login was successful.
In addition, we could avoid always reading the search params: e.g. even when they are used because the user is simply navigating throughout the app.

We could use a dedicated route and page for this logic. Let's say we have:
- a `/try-login` route that correspond to a TryLogin page. This route should always be available, regardless of the user already being authenticated or not: i.e. both `publicRoutes` and `authenticatedRoutes`.
- the TryLogin page reads the search params and attempts the login. It redirects to `"/"` in case of failure, or it redirects to the `redirectTo` param in case of success.
- besides this logic, the TryLogin just renders a loading spinner (instead of displaying it from the App file).

cc @Pavinati

frontend/src/App.tsx
```diff
@@ -96,6 +103,39 @@ function App() {
...
+  useEffect(() => {
+    if (attemptLogin) {
+      const authConfig: AuthConfig = {
+        tenantSlug: tenantSlug,
+        authToken: authToken,
+      };
+      auth.login(authConfig, false).then((isValidLogin) => {
+        setAttemptingLogin(false);
+        if (isValidLogin) {
+          navigate(redirectTo);
```
davidebriani, line 124: When navigating to a different route, it would be better to replace the current route with the new one in the browser's history instead of just adding the new route to the browser's history.
The reason is that if the user pushed the Back button of the browser to go back in history, it would reach again this route where the login is attempted and the user is automatically redirected to a new route. So the user wouldn't be able to go back in history, always being redirected to the latest page.

```suggestion
          navigate(redirectTo, { replace: true });
```

davidebriani (changes requested):

frontend/src/App.tsx
davidebriani: Of course, I have no strong opinion, `/auth` is fine.

Pavinati (approved):

davidebriani (approved):

PR #587, opened 2024-09-02
base_image: correctly scope upload to the BaseImageCollection bucket

### LINKED ISSUE #586: Base Image file uploads are not correctly scoped by their Base Image Collection id
This happens in v0.9.0-rc.1 due to the fact that `base_image_collection_id` gets set _after_ the upload

rbino:
The extracted base_image_collection_id was nil in the current implementation, leading to the image being put in the wrong bucket directory. This makes the storage module behaviour more explicit in its requirement (instead of having to pass the whole BaseImage) and adds a test to verify that the base_image_collection_id received by the storage module actually matches the one from the Base Image Collection.

Close #586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.0] - Unreleased
+### Fixed
+- Correctly scope Base Image uploads to their Base Image Collection bucket.
+
 ## [0.9.0-rc.1] - 2024-07-10
 ### Fixed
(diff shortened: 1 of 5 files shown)


backend/lib/edgehog/base_images/base_image/base_image.ex
```diff
@@ -71,7 +71,7 @@ defmodule Edgehog.BaseImages.BaseImage do
         description "A list of release display names in different languages."
       end
 
-      change Changes.HandleFileUpload
+      change Changes.HandleFileUpload, only_when_valid?: true
```
davidebriani, line 74: The Change already performs a check on `%Ash.Changeset{valid?: false}` and in that case avoids to upload the file.
We probably can remove one of these checks.

If you want to stick with `, only_when_valid?: true` we could also make a separate PR for that and update all relevant Changes that might need it (and this change is not really relevant for this PR).

davidebriani (changes requested):

davidebriani (approved):

PR #597, opened 2024-09-12
Fix attempt login for authenticated sessions

### LINKED ISSUE #596: Dashboard doesn't login with provided URL params if already authenticated
When an authenticated session is already present, opening the `/login` path with the `tenantSlug` and `authToken` query parameters does not result in being authenticated with the new parameters: the frontend remains authenticated with the previous tenant / credentials.

Since the same dashboard may be used with different tenants, for scenarios where programmatic login is attempted by being redirected from third party services, the third party services cannot know whether the frontend is already authenticated: the dashboard should try to login with the provided credentials regardless of any previously existing sessions.

davidebriani:
Correctly support automatic login attempts on the frontend regardless of existing auth sessions.

Fixes #596

### DIFF
--- a/frontend/src/pages/AttemptLogin.tsx
+++ b/frontend/src/pages/AttemptLogin.tsx
@@ -21,6 +21,7 @@
 import { useEffect } from "react";
 import { Spinner } from "react-bootstrap";
 import { useLocation, useNavigate } from "react-router-dom";
+import { commitLocalUpdate, useRelayEnvironment } from "react-relay/hooks";
 
 import { Route } from "Navigation";
 import { AuthConfig, useAuth } from "../contexts/Auth";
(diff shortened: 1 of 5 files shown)


frontend/src/pages/AttemptLogin.tsx
```diff
@@ -31,24 +32,35 @@ const AttemptLogin = () => {
   const searchParams = new URLSearchParams(search);
   const tenantSlug = searchParams.get("tenantSlug") || "";
   const authToken = searchParams.get("authToken") || "";
-  const redirectTo = searchParams.get("redirectTo") || "";
+  const redirectTo = searchParams.get("redirectTo") || "/";
   const navigate = useNavigate();
+  const relayEnvironment = useRelayEnvironment();
 
   useEffect(() => {
+    if (tenantSlug === "" || authToken === "") {
```
eddbbt, line 40: this only match exactly an empty string, a more wide check like 
```
if (!tenantSlug  || !authToken)
```

would catch also invalid/null/false values, it's worth to evaluate if we expect a 'false' value as a valid one

frontend/src/pages/AttemptLogin.tsx
```diff
@@ -31,24 +32,35 @@ const AttemptLogin = () => {
...
     };
+    commitLocalUpdate(relayEnvironment, (store) => store.invalidateStore());
     auth
       .login(authConfig, false)
       .then((isValidLogin) => {
         if (isValidLogin) {
           navigate(redirectTo, { replace: true });
         } else {
+          // Logout if credentials were not valid
+          auth.logout();
```
eddbbt, line 57: storage is already invalidated, so I expect credentials to not be valid anymore, do we still need to force a logout?

frontend/src/pages/AttemptLogin.tsx
davidebriani: TypeScript already guarantees that those are strings, so we don't need to check other values other than string values.
However I like your version better, so I took your advice. Thanks!

frontend/src/pages/AttemptLogin.tsx
davidebriani: We have:
- a Relay cache that may contain resources fetched using previous credentials (on a different tenant).
- a React context that holds a state about the current auth session.
We need to invalid both when attempting a login with different credentials.

eddbbt (approved):

PR #602, opened 2024-09-23
BaseImage: fix missing 'version' when uploading an image

### LINKED ISSUE #600: Missing `version` in scope leads to crash when uploading an image
https://github.com/edgehog-device-manager/edgehog/commit/0e4d17b0cdb432236ecf04632153fe4bdc75642d refactored how the scope gets passed to the `BaseImages.Storage` behaviour, but in doing so it left out the `version` field from the scope, which gets used [here](https://github.com/edgehog-device-manager/edgehog/blob/e782ab52377ce9ec13df7a1c1a9949c74649378e/backend/lib/edgehog/base_images/uploaders/base_image.ex#L48).

This was not covered by tests because they mock out the S3 upload part (see #601).

davidebriani:
Commit 0e4d17b refactored how the scope gets passed to the BaseImages.Storage behaviour, but in doing so the base image's version was not provided anymore and thus the upload logic failed for a missing 'version' field, which is used to determine the filename for the uploaded file.

This change adds the base image's version between the required parameters by the upload logic.

Closes #600.

### DIFF
--- a/backend/lib/edgehog/base_images/base_image/changes/handle_file_upload.ex
+++ b/backend/lib/edgehog/base_images/base_image/changes/handle_file_upload.ex
@@ -48,10 +48,12 @@ defmodule Edgehog.BaseImages.BaseImage.Changes.HandleFileUpload do
   defp upload_file(changeset, file) do
     tenant_id = changeset.to_tenant
 
+    {:ok, base_image} = Ash.Changeset.apply_attributes(changeset)
+
     {:ok, base_image_collection_id} =
       Ash.Changeset.fetch_argument(changeset, :base_image_collection_id)
 
-    case @storage_module.store(tenant_id, base_image_collection_id, file) do
+    case @storage_module.store(tenant_id, base_image_collection_id, base_image.version, file) do
       {:ok, file_url} ->
         changeset
         |> Ash.Changeset.force_change_attribute(:url, file_url)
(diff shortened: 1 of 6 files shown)


backend/lib/edgehog/base_images/base_image/changes/handle_file_upload.ex
```diff
@@ -48,10 +48,12 @@ defmodule Edgehog.BaseImages.BaseImage.Changes.HandleFileUpload do
   defp upload_file(changeset, file) do
     tenant_id = changeset.to_tenant
 
+    {:ok, base_image} = Ash.Changeset.apply_attributes(changeset)
```
rbino, line 51: The point of splitting the arguments was avoiding to apply attributes here (since it's failable), use `Ash.Changset.fetch_argument(changeset, :version)` below like it's done for `base_image_collection_id`

rbino (changes requested):

backend/lib/edgehog/base_images/base_image/changes/handle_file_upload.ex
rbino: Merging this anyway and implementing this fix in a separate PR

davidebriani: I didn't understand your point, since `fetch_argument` is failable too.
But I changed it to `fetch_change` (`version` is not an argument) so Ash doesn't have to apply all attributes just to get the version.

PR #605, opened 2024-10-02
refactor(nix): use flake-parts and nixfmt

noaccOS:
elixir-utils needed an update to add support for elixir 1.17/otp27

as elixir-utils is now using flake-parts, we do so as well to  shrink the size of the lock file. the behavior should be exactly the same.

use nixfmt as a formatter as it is now the official nix formatter.

### DIFF
--- a/backend/flake.lock
+++ b/backend/flake.lock
@@ -2,19 +2,20 @@
   "nodes": {
     "elixir-utils": {
       "inputs": {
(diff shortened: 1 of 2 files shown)


rbino:
Ah yes clearly

*nods head*

rbino (approved):

Issue #606, opened 2024-10-02
Initialize Edgehog.Containers domain with ImageCredentials resource


rbino:
The resource should have those attributes:
- `:id` (UUID)
- `:name` (string, non-null, unique)
- `:username` (string, non-null)
- `:password` (string, non-null, sensitive, only accepted when creating but not exposed when reading via public APIs)
- create/update timestamps named coherently with the rest of Edgehog

The resource should be multitenant like all resources.

The `Edgehog.Containers` domain should expose create, get, list and destroy operations for the resource.  All those query/mutations should be tested.

rbino:
Hint: the new Ash generators might help (`mix help ash.gen.resource`).

For now, manually generate the migration to initialize the db table (`mix ecto.gen.migration` and populate accordingly), but I'm currently working on #500 so in the end we should be able to replace that with Ash generated migrations once rebased.

PR #607, opened 2024-10-04
Initialize Containers.ImageCredentials resource

lusergit:
Added the `Container` domain with the `ImageCredentials` resource. The resource has the attributes decribed with issue #606 and the domain exposes the required operations. All operations are tested with Absinthe.


backend/lib/edgehog_web/schema.ex
```diff
@@ -31,7 +31,8 @@ defmodule EdgehogWeb.Schema do
       Edgehog.Labeling,
       Edgehog.OSManagement,
       Edgehog.Tenants,
-      Edgehog.UpdateCampaigns
+      Edgehog.UpdateCampaigns,
+      Edgehog.Containers
```
rbino, line 35: Nit: let's maintain alphabetical order

backend/lib/edgehog/containers/image_credentials.ex
```diff
@@ -0,0 +1,80 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2021-2024 SECO Mind Srl
```
rbino, line 4: New files should have only `2024`

backend/lib/edgehog/containers/image_credentials.ex
```diff
@@ -0,0 +1,80 @@
...
+    attribute :password, :string do
+      sensitive? true
+    end
+
+    create_timestamp :inserted_at
+    update_timestamp :updated_at
+  end
+
+  identities do
+    identity :name_tenant_id, :name
```
rbino, line 73: The `_tenant_id` suffix format has been superseded by #608, it was used just to be consistent with existing indexes.
It should be `identity :name, :name` (does the second parameter also accepts a single atom instead of a list of atoms? TIL)

backend/lib/edgehog/containers/image_credentials.ex
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  postgres do
+    table("image_credentials")
+    repo(Edgehog.Repo)
```
rbino, line 78: Extra parentheses, they shouldn't be there

backend/lib/edgehog/containers/image_credentials.ex
rbino: Note that I think you have to run again the generator after you change this. Don't do two different migrations for the rename, just recreate the migration from scratch. You have to throw away both the generated migration _and_ its relative resource snapshot JSON to do so.

backend/lib/edgehog/containers/image_credentials.ex
```diff
@@ -0,0 +1,80 @@
...
+      public? true
+    end
+
+    attribute :username, :string do
+      allow_nil? false
+      public? true
+    end
+
+    attribute :password, :string do
+      sensitive? true
```
rbino, line 65: `allow_nil? false` here too I guess, unless there's some usecase where you want to authenticate with a registry just with a username (I don't think there is)

backend/test/edgehog_web/schema/query/image_credentials_test.exs
```diff
@@ -0,0 +1,102 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule EdgehogWeb.Schema.Query.ImageCredentialsTest do
+  use EdgehogWeb.GraphqlCase, async: true
+
+  import Edgehog.ImageCredentialsFixtures
+
+  require IEx
```
rbino, line 26: Leftover

backend/test/edgehog_web/schema/query/image_credentials_test.exs
```diff
@@ -0,0 +1,102 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2021-2024 SECO Mind Srl
```
rbino, line 4: Year

backend/test/edgehog_web/schema/query/image_credentials_test.exs
```diff
@@ -0,0 +1,102 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule EdgehogWeb.Schema.Query.ImageCredentialsTest do
```
rbino, line 21: We usually do one-file-per-query so split into `ImageCredentialsTest` and `ListImageCredentialsTest`

backend/config/config.exs
```diff
@@ -92,6 +92,7 @@ config :edgehog, EdgehogWeb.Endpoint,
   live_view: [signing_salt: "aiSLZVyY"]
 
 config :edgehog, :ash_domains, [
+  Edgehog.Containers,
```
rbino, line 95: See other comment, let's keep these in alphabetical order

rbino (changes requested):

rbino (approved):

backend/lib/edgehog/containers/containers.ex
```diff
@@ -0,0 +1,53 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2021-2024 SECO Mind Srl
```
noaccOS, line 4:
```suggestion
# Copyright 2024 SECO Mind Srl
```

backend/lib/edgehog/containers/image_credentials.ex
```diff
@@ -1,24 +1,81 @@
...
+    update_timestamp :updated_at
+  end
+
+  identities do
+    identity :name, [:name]
+  end
+
+  postgres do
+    table "image_credentials"
+    repo Edgehog.Repo
```
noaccOS, line 79:
```suggestion
```

I believe the repo is already defined in [`Edgehog.MultitenantResource`](https://github.com/edgehog-device-manager/edgehog/blob/1cd7df0671e9ad40c0bf17c4fb5316ad96e497b0/backend/lib/edgehog/multitenant_resource.ex#L50)

backend/lib/edgehog/containers/containers.ex
```diff
@@ -0,0 +1,53 @@
...
+        description "Returns all available image credentials."
+        paginate_with nil
+      end
+    end
+
+    mutations do
+      create ImageCredentials, :create_image_credentials, :create
+      destroy ImageCredentials, :delete_image_credentials, :destroy
+    end
+  end
```
noaccOS, line 48: In general, how do we want to approach image credentials? 
imo we should use the same guidelines as the images:

> `Let's not add any GraphQL queries for now and just add them as we go to satisfy the frontend needs.`

How about we split this PR in two: one for the resources, one draft pr for the graphql mutations?

backend/lib/edgehog/containers/image_credentials.ex
```diff
@@ -0,0 +1,81 @@
...
+  graphql do
+    type :image_credentials
+  end
+
+  actions do
+    defaults [:destroy, :read]
+
+    create :create do
+      accept [:name, :username, :password]
+    end
```
noaccOS, line 48:
```suggestion
    create :create do
      primary? true
      accept [:name, :username, :password]
    end
```

how about we mark this as primary?

backend/lib/edgehog/containers/image_credentials.ex
rbino: Indeed it is, but I think this is the status quo for basically all existing resource. Let's keep it consistent for now and decide later if want to remove it from the MultitenantResource or from the single resources in a separate PR.

backend/lib/edgehog/containers/containers.ex
rbino: Here it makes sense to have GraphQL functionalities from the get-go (as requested in #606) because in the initial flow that was sketched credentials are added in an explicit separate step (while, e.g., Images, currently are not). I'd keep the GraphQL mutations/queries, we can always change them later if there are different requirements from the frontend.

PR #613, opened 2024-10-09
feat: azure support

### LINKED ISSUE #233: Support for Azure Blob Storage
Edgehog should have either integrated native support for Azure Blob Storage in addition to S3 APIs, or there should be a documented procedure about how to use it.

noaccOS:
add azure support by adding a waffle compatible storage module based on azurex.

the configuration mostly maps to existing s3 configuration, so for now I did not deem it necessary to separate the configuration. I am however open to implementing a separate configuration if this was preferred though.

Two new environment variables were added for configuration:
- STORAGE_TYPE, which can be either s3 or azure
- AZURE_API_URL, because it needs more flexibility than just scheme, host and port.

also add integration tests and allow runtime configuration of s3 storage

closes #233


rbino (changes requested):
Not urgent, but before merging this needs:
- Changelog entry
- Documentation about how to use Azure Blob Storage in the "Deploying with Kubernetes" page

lusergit (approved):

backend/lib/edgehog/base_images/bucket_storage.ex
```diff
@@ -42,8 +42,20 @@ defmodule Edgehog.BaseImages.BucketStorage do
...
+      version: base_image_version,
+      base_image_collection_id: base_image_collection_id
+    } = base_image
+
+    scope = %{
+      tenant_id: tenant_id,
+      base_image_collection_id: base_image_collection_id,
+      base_image_version: base_image_version
+    }
+
```
davidebriani, line 58: This was broken from previous releases.
It would be nice to have this fix in v0.9 and forward-port it to main.

davidebriani (changes requested):

backend/lib/edgehog/base_images/bucket_storage.ex
noaccOS: Done in #764

davidebriani (changes requested):
Brilliant PR!
If we manage to use a more informative changelog line, it's a merge

davidebriani (approved):

PR #636, opened 2024-10-14
chore(container): set up deployment event interface handling

noaccOS:
initialize infrastructure for the DeploymentEvent interface, similarly to OTAEvent.

the actual handling of the state change will be done at a later date, once the actual deployment process is implemented.

### DIFF
--- a/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
+++ b/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
@@ -31,6 +31,7 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
   alias Edgehog.Triggers.IncomingData
   alias Edgehog.Triggers.TriggerPayload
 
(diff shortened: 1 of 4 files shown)


backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -103,6 +104,21 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
   end
 
+  defp handle_event(%IncomingData{interface: @deployment_event} = event, tenant, _realm_id, _device_id, _timestamp) do
+    "/" <> deployment_id = event.path
+    status = event.value["status"]
+    message = event.value["message"]
+
+    status_attrs = %{status: status, mesage: message}
+
+    # TODO: update deployment status
```
rbino, line 114: Add an issue for this with the correct milestone and project so we don't lose track of it

rbino (changes requested):

PR #637, opened 2024-10-15
feat(containers): add auth repo signing

shinnokdisengir:
initial commit


frontend/src/schema/ImageCredential.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export const createImageCredentialSchema = object({
+  password: string()
+    .required()
+    .matches(/^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[!@#$%^_]).{8,12}$/, {
+      message: messages.passwordStronger.id,
+    }),
+})
+  .concat(commonImageCredentialSchema)
+  .required();
```
davidebriani, line 68: We can skip validation and remove this and other files related to validation.
Specifically, the password should not have any validation: the user is not choosing a new password on Edgehog, but just inputs the password as already configured in the container registry's service.

davidebriani (changes requested):

frontend/src/i18n/langs/en.json
```diff
@@ -600,6 +612,15 @@
...
   },
+  "components.Sidebar.applicationsLabel": {
+    "defaultMessage": "Applications"
+  },
+  "components.Sidebar.applications.credentialsLabel": {
+    "defaultMessage": "Credentials"
+  },
+  "components.Sidebar.applications.managementLabel": {
+    "defaultMessage": "Management"
+  },
```
davidebriani, line 623: This is not used anywhere in the code, we can remove it.

frontend/src/i18n/langs/en.json
```diff
@@ -600,6 +612,15 @@
   "components.Sidebar.updateChannelsLabel": {
     "defaultMessage": "Update Channels"
   },
+  "components.Sidebar.applicationsLabel": {
+    "defaultMessage": "Applications"
+  },
+  "components.Sidebar.applications.credentialsLabel": {
+    "defaultMessage": "Credentials"
+  },
```
davidebriani, line 620: This is simple and fine.

I am just wondering if a user can quickly understand what the `Credentials` section is about. Since it has a very generic name, even though it's placed within the Applications menu, it is not straightforward to see that these credentials are those for container registries and are supposed to be used when defining the containers of an application.

Maybe we could find a simple way (adjusting the labels a bit) to make the UX more intuitive.
What's your take on this?

davidebriani (changes requested):

frontend/src/forms/CreateImageCredential.tsx
```diff
@@ -0,0 +1,158 @@
...
+        autoComplete="username"
+        style={{ display: "none" }}
+      />
+      <input
+        id="password"
+        type="password"
+        name="fakepasswordremembered"
+        autoComplete="new-password"
+        style={{ display: "none" }}
+      />
```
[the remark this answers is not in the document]
shinnokdisengir: Nope :( unfortunately, to bypass the chrome's password autocompletation I had to resort to this trick

frontend/src/forms/CreateImageCredential.tsx
```diff
@@ -0,0 +1,158 @@
...
+          controlId="image-credential-form-password"
+          label={
+            <FormattedMessage id="components.CreateImageCredentialForm.passwordLabel" />
+          }
+        >
+          <Form.Control
+            {...register("password")}
+            className="security"
+            isInvalid={!!errors.password}
+            autoComplete="new-password"
```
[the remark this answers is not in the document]
shinnokdisengir: see above

frontend/src/schema/ImageCredential.ts
shinnokdisengir: I had already discussed this at the time with @rbino  and we agreed it was a plus. At most, and consistent with what was said earlier, I can just comment on the regex for future use

frontend/src/i18n/langs/en.json
shinnokdisengir: right, it doesn't exist anymore

frontend/src/Navigation.tsx
```diff
@@ -58,6 +58,9 @@ enum Route {
   application = "/applications/:applicationId",
   release = "/applications/:applicationId/release/:releaseId",
   releaseNew = "/applications/:applicationId/release/new",
+  imageCredentials = "/image-credentials",
+  imageCredentialsEdit = "/image-credentials/:imageCredentialId/edit",
+  imageCredentialsNew = "/image-credentials/new",
```
OmarBrbutovic, line 63: Update breadcrumbs navigation logic by adding the necessary cases in `matchPath()` and `breadcrumbRoutes` to handle the new routes:

1. In `matchPath()` (in this same file) add :

    ```ts
    case Route.imageCredentials:
    case Route.imageCredentialsNew:
    case Route.login:
    case Route.logout:
      return { route } as ParametricRoute;
    ```

    ```ts
    case Route.imageCredentialsEdit:
      return params && typeof params["imageCredentialId"] === "string"
        ? {
            route,
            params: { imageCredentialId: params.imageCredentialId },
          }
        : null;
    ```

2. In `Page.tsx`, inside the `breadcrumbRoutes` add:

    ```ts
    case Route.imageCredentials:
    case Route.login:
    case Route.logout:
      return [currentRoute];
    ```

    ```ts
    case Route.imageCredentialsEdit:
    case Route.imageCredentialsNew:
      return [{ route: Route.imageCredentials }, currentRoute];
    ```

These changes ensure proper breadcrumb navigation for image credentials-related routes.

frontend/src/Navigation.tsx
```diff
@@ -365,6 +370,18 @@ const routeTitles: Record<Route, MessageDescriptor> = defineMessages({
...
+    defaultMessage: "Image Credentials",
+  },
+  [Route.imageCredentialsNew]: {
+    id: "navigation.routeTitle.ImageCredentialNew",
+    defaultMessage: "Image Credentials New",
+  },
+  [Route.imageCredentialsEdit]: {
+    id: "navigation.routeTitle.ImageCredentialEdit",
+    defaultMessage: "Image Credentials Edit",
+  },
```
OmarBrbutovic, line 384: i suggest following the current convention of naming Routes:
- `Create Image Credentials`
- `Image Credential Details`

frontend/src/components/ImageCredentialsTable.tsx
```diff
@@ -0,0 +1,84 @@
...
+    # eslint-disable-next-line relay/unused-fields
+    username
+  }
+`;
+
+const columnHelper = createColumnHelper<ImageCredential>();
+const columns = [
+  columnHelper.accessor("label", {
+    header: () => (
+      <FormattedMessage id="components.ImageCredentialsTable.labelTitle" />
```
OmarBrbutovic, line 47: For all instances where you use `FormattedMessage`, I recommend specifying the `defaultMessage` inline. This makes the text easier to read and update without needing to search for it in `en.json`. 

To streamline the use of `FormattedMessage`, after setting the `defaultMessage` values, you can run:

```bash
npm run i18n:extract
```

This command extracts all messages into `frontend/translations.generated.json`. You can then copy its contents and paste them into `en.json` to keep translations up to date.

After this, run:

```bash
npm run i18n:compile
```

This will compile the translations and ensure everything is ready for use.

frontend/src/components/Sidebar.tsx
```diff
@@ -208,12 +208,23 @@ const Sidebar = () => (
...
-          id="components.Sidebar.applicationsLabel"
+          id="components.Sidebar.applications.applicationsLabel"
           defaultMessage="Applications"
         />
       }
       icon="applications"
     >
+      <SidebarItem
+        label={
+          <FormattedMessage id="components.Sidebar.applications.credentialsLabel" />
```
OmarBrbutovic, line 219: I personally prefer the `defaultMessage` for this to be `Image Credentials` rather than `Image Registries Credentials` for simplicity and clarity.

frontend/src/components/Sidebar.tsx
```diff
@@ -208,12 +208,23 @@ const Sidebar = () => (
...
       label={
         <FormattedMessage
-          id="components.Sidebar.applicationsLabel"
+          id="components.Sidebar.applications.applicationsLabel"
           defaultMessage="Applications"
         />
       }
       icon="applications"
     >
+      <SidebarItem
```
OmarBrbutovic, line 217: I would recommend placing this `SidebarItem` after `Applications` , it improves the flow and feels more intuitive.

OmarBrbutovic (changes requested):

frontend/src/Navigation.tsx
shinnokdisengir: Yes, these were already done 2 months ago, but too much time has passed and the merge with your code did not help...

frontend/src/components/Table.tsx
```diff
@@ -56,7 +56,7 @@ const SortDirectionIndicator = ({
   </span>
 );
 
-type TableProps<T extends RowData> = {
+export type TableProps<T extends RowData> = {
```
OmarBrbutovic, line 59: What’s the purpose of exporting this if it’s not being used anywhere?

OmarBrbutovic (approved):

frontend/src/schema/ImageCredential.ts
```diff
@@ -0,0 +1,54 @@
...
+*/
+
+import { defineMessages } from "react-intl";
+import { object, setLocale, string } from "yup";
+
+const messages = defineMessages({
+  required: {
+    id: "validation.required",
+  },
+});
```
OmarBrbutovic, line 28: Duplicate message id: "validation.required", but the `description` and/or `defaultMessage` are different.

frontend/src/schema/ImageCredential.ts
OmarBrbutovic: If you need to use `yup` validation, there's no need to create a new file. You can place the validation in `frontend/src/forms/index.ts`, which is intended for this purpose. This approach will also resolve the duplicate message ID issue also.

frontend/src/forms/CreateImageCredential.tsx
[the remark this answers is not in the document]
OmarBrbutovic: To remove the `Hidden Dummy Input`, as Davide suggested, you can use `type="password"` instead of `className="security"`. This eliminates the need for custom CSS, allowing you to remove `CreateImageCredential.scss`, and it also resolves the `autoComplete` issue. 

However, if you use `autoComplete="new-password"`, Google’s password manager may prompt to suggest a new password. To avoid this, set `autoComplete="off"`.

You can simplify `Form.Control` for passwords like this:

```tsx
<Form.Control
  {...register("password")}
  type="password"
  isInvalid={!!errors.password}
  autoComplete="off"
  onCopy={(e) => e.preventDefault()}
  onCut={(e) => e.preventDefault()}
/>
```

davidebriani (approved):

PR #640, opened 2024-10-16
feat(containers): implemented create image request interface

### LINKED ISSUE #630: containers: add support for interface "io.edgehog.devicemanager.apps.CreateImageRequest"
ref: #614 
interface [here](https://github.com/edgehog-device-manager/edgehog-astarte-interfaces/pull/64/files#diff-33bbcd4723777662206f11b5f0f24a6d22ef942a1e72c198279c73531e5cf6a6)

lusergit:
Closes #630

- Implemented the `create_image` behaviour
- Implemented the `base64_json` calculation in the `image_credentials` resource
- Implemented the manual action `send_create_image_request` in the `device` resource. This is inteded to be internally called with an `Image` and an optional `ImageCredentials` struct to send the appropriate astarte call to the device.

### DIFF
--- a/backend/lib/edgehog/containers/image_credentials.ex
+++ b/backend/lib/edgehog/containers/image_credentials.ex
@@ -75,6 +75,10 @@ defmodule Edgehog.Containers.ImageCredentials do
     has_many :images, Edgehog.Containers.Image
   end
 
+  calculations do
+    calculate :base64_json, :string, Base64JsonEncoding
+  end
+
   identities do
     identity :name, [:name]
   end
--- a/backend/lib/edgehog/containers/image_credentials/base64_json_encoding.ex
+++ b/backend/lib/edgehog/containers/image_credentials/base64_json_encoding.ex
@@ -1,0 +1,35 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Containers.ImageCredentials.Base64JsonEncoding do
+  @moduledoc false
+  use Ash.Resource.Calculation
+
+  alias AshGraphql.Types.JSONString
+
+  @impl Ash.Resource.Calculation
+  def calculate(records, _opts, _context) do
+    for record <- records do
+      %{username: record.username, password: record.password}
+      |> JSONString.encode()
+      |> Base.encode64()
+    end
+  end
+end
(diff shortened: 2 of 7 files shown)


backend/lib/edgehog/devices/device/manual_actions/send_create_image_request.ex
```diff
@@ -0,0 +1,60 @@
...
+                                       )
+
+  @impl Ash.Resource.ManualUpdate
+  def update(changeset, _opts, _context) do
+    device = changeset.data
+
+    image_id = changeset.arguments.image.id
+    reference = changeset.arguments.image.reference
+
+    credentials =
```
shinnokdisengir, line 42: no check is done on `credentials` (e.g. that it is %{el1, el2}). In the case wouldn't it be better to put it inside the `with`?

lusergit: Yess now the code checks for correct credentials calculation thanks 👍🏻

backend/lib/edgehog/containers/image_credentials/base64_json_encoding.ex
```diff
@@ -0,0 +1,38 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Containers.ImageCredentials.Base64JsonEncoding do
+  @moduledoc false
+  use Ash.Resource.Calculation
+
+  alias AshGraphql.Types.JSON
+
+  def calculate(records, _opts, _context) do
```
noaccOS, line 27: Let's use `@impl` to mark this function as part of `Ash.Resource.Calculation`

backend/lib/edgehog/containers/image_credentials/base64_json_encoding.ex
```diff
@@ -0,0 +1,38 @@
...
+    Enum.map(records, fn record ->
+      with {:ok, username} <- Map.fetch(record, :username),
+           {:ok, password} <- Map.fetch(record, :password) do
+        data = [username: username, password: password]
+        json = JSON.encode(data)
+
+        Base.encode64(json)
+      end
+    end)
```
noaccOS, line 36: Just like in #645, we're dealing with structs so we already know their attributes.
nit/tip: pipelines are your friend
nit/tip 2: sometimes the code is (arguably) clearer using for comprehensions

```suggestion
    for record <- records do
      %{username: record.username, password: record.password}
      |>  JSON.encode()
      |> Base.encode64()
    end
```

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -224,6 +226,22 @@ defmodule Edgehog.Devices.Device do
...
+
+      argument :image, :struct,
+        constraints: [instance_of: Image],
+        description: "The image the device will pull.",
+        allow_nil?: false
+
+      argument :credentials, :struct,
+        constraints: [instance_of: ImageCredentials],
+        description: "The credentials to use.",
+        allow_nil?: true
```
noaccOS, line 241: let's use the do notation as that's what's being used in the codebase when there is more than one option

backend/lib/edgehog/containers/image_credentials.ex
```diff
@@ -75,6 +75,10 @@ defmodule Edgehog.Containers.ImageCredentials do
     has_many :images, Edgehog.Containers.Image
   end
 
+  calculations do
+    calculate :base64_json, :string, {Base64JsonEncoding}
```
rbino, line 79: No need for tuple (not sure it even works to be quite frank). Should be either [`module` or `{module, args}`](https://hexdocs.pm/ash/dsl-ash-resource.html#calculations-calculate-calculation) (or possibly `expr(...)`)

rbino (changes requested):

rbino (approved):

PR #642, opened 2024-10-16
chore(containers): add release resource

### LINKED ISSUE #633: containers: Add Containers.Release resource
Attributes:
```
id (uuid primary key)
version (string, non-null, must be a valid semantic version)
application_id (belongs_to, references Containers.Application, see #632)
```

unique index on application_id + version

noaccOS:
Depends on #641 
Closes #633


lusergit (approved):

backend/lib/edgehog/containers/release.ex
```diff
@@ -0,0 +1,59 @@
...
+  relationships do
+    belongs_to :application, Edgehog.Containers.Application, attribute_type: :uuid
+  end
+
+  identities do
+    identity :application_version, [:application_id, :version]
+  end
+
+  postgres do
+    table "releases"
```
rbino, line 57: I'd do `application_releases` since release is a generic enough term that we might end up reusing it somewhere else

rbino (approved):

PR #645, opened 2024-10-17
Feature/container resource

### LINKED ISSUE #644: containers: Add `Containers.Container` resource
Attributes

- `id` (uuid primary key)
- `restart_policy` (enum with possible values `:nil`, `:no`, `:always`, `:unless_stopped` and `:on_failure`)
- `hostname` (string)
- `env` (map)
- `privileged` (boolean)

fields come from the [`io.edgehog.devicemanager.apps.CreateContainerRequest` interface](https://github.com/edgehog-device-manager/edgehog-astarte-interfaces/blob/055d6266c4f0b7612dd5838f71b062ba49999641/io.edgehog.devicemanager.apps.CreateContainerRequest.json)

lusergit:
Closes #644 

Implements the `Container` resource


backend/config/config.exs
```diff
@@ -58,7 +58,10 @@ resource_section_order = [
   :postgres
 ]
 
-config :ash, :custom_types, id: Edgehog.Types.Id
+config :ash, :custom_types,
+  id: Edgehog.Types.Id,
+  restart_policy: Edgehog.Containers.Types.RestartPolicy
```
noaccOS, line 63: is this needed? we don't do that for eg [led behavior](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/devices/device/led_behavior.ex)

backend/lib/edgehog/containers/container.ex
```diff
@@ -0,0 +1,73 @@
...
+defmodule Edgehog.Containers.Container do
+  @moduledoc false
+  use Edgehog.MultitenantResource,
+    domain: Edgehog.Containers
+
+  alias Edgehog.Containers.Container.EnvJsonEncoding
+  alias Edgehog.Containers.Image
+
+  actions do
+    defaults [:read, :destroy, :create, :update]
```
noaccOS, line 30: if you use `:create` or `:update`, they won't accept any argument.
consider using `create: :*` to accept all public fields or `create: [:restart_policy, :hostname, ...]` to manually specify the (public or private) accepted fields

backend/lib/edgehog/containers/container/env_json_encoding.ex
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Containers.Container.EnvJsonEncoding do
+  @moduledoc false
+  use Ash.Resource.Calculation
+
+  alias AshGraphql.Types.JSONString
+
+  def calculate(records, _opts, _context) do
```
noaccOS, line 27:
```suggestion
  @impl Ash.Resource.Calculation
  def calculate(records, _opts, _context) do
```

noaccOS (changes requested):

backend/config/config.exs
lusergit: Indeed it was not needed 👍🏻

noaccOS (approved):
I'm not 100% convinced we should mark all fields as public by default, but since this is a feature branch I'm not too concerned with making some breaking changes later if we need to

noaccOS (approved):
LGTM

rbino (approved):

PR #646, opened 2024-10-17
chore(containers): create container request 

### LINKED ISSUE #628: containers: add support for interface "io.edgehog.devicemanager.apps.CreateContainerRequest"
ref: #614 
interface [here](https://github.com/edgehog-device-manager/edgehog-astarte-interfaces/pull/64/files#diff-48c9921b362a1acc00dc72c0a8840f9424e87189f45ada4d3146b431b8cf3f40)

lusergit:
Depends on #645 ,
closes #628

### DIFF
--- a/backend/lib/edgehog/astarte/device/create_container_request.ex
+++ b/backend/lib/edgehog/astarte/device/create_container_request.ex
@@ -1,0 +1,50 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Astarte.Device.CreateContainerRequest do
+  @moduledoc false
+  @behaviour Edgehog.Astarte.Device.CreateContainerRequest.Behaviour
+
+  alias Astarte.Client.AppEngine
+  alias Edgehog.Error.AstarteAPIError
+
+  @interface "io.edgehog.devicemanager.apps.CreateContainerRequest"
+
+  @impl Edgehog.Astarte.Device.CreateContainerRequest.Behaviour
+  def send_create_container_request(%AppEngine{} = client, device_id, request_data) do
+    request_data_map = Map.from_struct(request_data)
+
+    api_call =
+      AppEngine.Devices.send_datastream(
+        client,
+        device_id,
+        @interface,
+        "/container",
+        request_data_map
+      )
+
+    with {:error, api_error} <- api_call do
+      reason =
+        AstarteAPIError.exception(status: api_error.status, response: api_error.response)
+
+      {:error, reason}
+    end
+  end
+end
--- a/backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
+++ b/backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
@@ -1,0 +1,66 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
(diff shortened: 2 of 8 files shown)


backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
```diff
@@ -0,0 +1,58 @@
...
+  def update(changeset, _opts, _context) do
+    device = changeset.data
+    container = changeset.container
+
+    data = %{
+      container_id: container.id,
+      image_id: container.image_id,
+      networks_ids: [],
+      volume_ids: [],
+      hostname: container.hostname,
```
rbino, line 41: Is this (and all other fields) guaranteed to be non-nil? Otherwise I think we have to default to empty string at some level (e.g. `container.hostname || ""`) because IIRC the Rust SDK is not happy if an object aggregation sends a null field (cc @joshuachp for confirmation)

rbino (changes requested):

backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
joshuachp: Yes, it was agreed that we will never receive incomplete object aggregates from Astarte.

lusergit: Makes sense, I think is best to place the constraints in the resource with `allow_nil? false` and `constraints [allow_empty?: true]`

backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
```diff
@@ -0,0 +1,66 @@
...
+          device.appengine_client,
+          device.device_id,
+          data
+        )
+      end
+    end
+  end
+
+  defp to_correct_string(:unless_stopped), do: "unless-stopped"
+  defp to_correct_string(:on_failure), do: "unless-stopped"
```
rbino, line 64: Typo I guess.

Also, I'd probably handle this either replacing `_` with `-` or using something like [Recase](https://github.com/wemake-services/recase) (probably the former, given you know that you always have a snake case atom as input)

The reasoning is that if you add a new multi-word restart policy you'd have to change the code in two different places with the current solution

rbino (changes requested):

backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
```diff
@@ -0,0 +1,64 @@
...
+
+  @impl Ash.Resource.ManualUpdate
+  def update(changeset, _opts, _context) do
+    device = changeset.data
+
+    with {:ok, container} <- Ash.load(changeset.container, :env_encoding) do
+      env_encoding = container.env_encoding
+      restart_policy = to_correct_string(container.restart_policy)
+
+      data = %{
```
noaccOS, line 39:
```suggestion
      data = %RequestData{
```

I think dyalizer wouldn't like this as it is. Let's give it "proper" typing

lusergit: The `send_datastream` function accepts a map as input, not a `RequestData` instance. For simplicity I intially just wrote a map but proer typing is better I agree

noaccOS: It works, but it breaks the contract: [`9f02250` (#646)](https://github.com/edgehog-device-manager/edgehog/pull/646/commits/9f02250a6df6a69c2a9ac62d7624ba485dfc78e8#diff-b705cafe3b60fe47455496a41accaa2ac6316f9c2ac96d2ddd0dd70fb6b94db7R29)

backend/lib/edgehog/astarte/device/create_container_request.ex
```diff
@@ -0,0 +1,42 @@
...
+  @behaviour Edgehog.Astarte.Device.CreateContainerRequest.Behaviour
+
+  alias Astarte.Client.AppEngine
+  alias Edgehog.Astarte.Device.CreateContainerRequest.RequestData
+
+  @interface "io.edgehog.devicemanager.apps.CreateContainerRequest"
+
+  @impl Edgehog.Astarte.Device.CreateContainerRequest.Behaviour
+  def send_create_container_request(%AppEngine{} = client, device_id, %RequestData{} = request_data) do
+    request_data_map = Map.from_struct(request_data)
```
noaccOS, line 32: Is this needed?

rbino: Probably yes since `send_datastream` expects a map

PR #648, opened 2024-10-17
chore(containers): add deployment resource

### LINKED ISSUE #643: containers: add Containers.Deployment resource
Attributes:

```sql
id (uuid primary key)
device_id (belongs_to, references Edgehog.Devices.Device)
release_id (belongs_to, references Edgehog.Containers.Release)
```

unique index on device_id + release_id

noaccOS:
depends on #642 
closes #643


lusergit (approved):

backend/lib/edgehog/devices/device/device.ex
```diff
@@ -302,6 +302,11 @@ defmodule Edgehog.Devices.Device do
       description "The existing OTA operations for this device"
       writable? false
     end
+
+    many_to_many :releases, Edgehog.Containers.Release do
+      through Edgehog.Containers.Deployment
+      join_relationship :deployments
```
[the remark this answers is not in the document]
noaccOS: because the deployment table actually holds meaning and it's not just for the many-to-many relationship. all interfaces require the deployment id and not the interface id.
nothing is **needed**, but `Ash.load(device, :deployments)` conveys more meaning than `Ash.load(device, :releases_join_assoc)`

backend/lib/edgehog/containers/deployment.ex
```diff
@@ -0,0 +1,51 @@
...
+
+    timestamps()
+  end
+
+  relationships do
+    belongs_to :device, Edgehog.Devices.Device
+
+    belongs_to :release, Edgehog.Containers.Release do
+      attribute_type :uuid
+    end
```
lusergit, line 41: Shouldn't this also include a reference to the containers?
```suggestion
      has_many :containers, Edgehog.Containers.Container
    end
```

rbino: General note: it's not mandatory to have _all_ `has_many` relationships everywhere, they can be added if/when they're actually needed/used

rbino (approved):

PR #653, opened 2024-10-21
feat(containers): deployment command interface

noaccOS:
allows sending a command to a release

closes #620

PR #654, opened 2024-10-21
feat(containers): application upgrade interface

### LINKED ISSUE #619: containers: add support for interface "io.edgehog.devicemanager.apps.DeploymentUpdate"
ref: #614
interface: https://github.com/edgehog-device-manager/edgehog-astarte-interfaces/pull/64/files#diff-0482701a7ded078199e43f0fcfc2291ca0b384dc5be49ac52969a3d3938e1e17

noaccOS:
expose a way to upgrade an application to a newer release.

- does not allow downgrading
- the releases must be of the same application

~depends on #653 and is drafted because the last commit (which must be removed before merging) depends on it~
closes #619

### DIFF
--- a/backend/lib/edgehog/astarte/device/deployment_update.ex
+++ b/backend/lib/edgehog/astarte/device/deployment_update.ex
@@ -1,0 +1,42 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Astarte.Device.DeploymentUpdate do
+  @moduledoc false
+
+  @behaviour Edgehog.Astarte.Device.DeploymentUpdate.Behaviour
+
+  alias Astarte.Client.AppEngine
+  alias Edgehog.Error.AstarteAPIError
+
+  @interface "io.edgehog.devicemanager.apps.DeploymentUpdate"
+
+  @impl Edgehog.Astarte.Device.DeploymentUpdate.Behaviour
+  def update(client, device_id, data) do
+    api_call =
+      AppEngine.Devices.send_datastream(client, device_id, @interface, "/deployment", data)
+
+    with {:error, api_error} <- api_call do
+      reason = AstarteAPIError.exception(status: api_error.status, response: api_error.response)
+
+      {:error, reason}
+    end
+  end
+end
--- a/backend/lib/edgehog/astarte/device/deployment_update/request_data.ex
+++ b/backend/lib/edgehog/astarte/device/deployment_update/request_data.ex
@@ -1,0 +1,30 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Astarte.Device.DeploymentUpdate.RequestData do
+  @moduledoc false
+
+  defstruct [:from, :to]
+
+  @type t :: %__MODULE__{
+          from: String.t(),
+          to: String.t()
+        }
+end
--- a/backend/lib/edgehog/devices/device/manual_actions/update_application.ex
+++ b/backend/lib/edgehog/devices/device/manual_actions/update_application.ex
@@ -1,0 +1,52 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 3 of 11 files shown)


backend/lib/edgehog/devices/device/manual_actions/send_application_command.ex
```diff
@@ -0,0 +1,69 @@
...
+      |> case do
+        :ok ->
+          {:ok, device}
+
+        {:error, api_error} ->
+          reason =
+            AstarteAPIError.exception(status: api_error.status, response: api_error.response)
+
+          {:error, reason}
+      end
```
lusergit, line 54: maybe nit: Wouldn't it be better to move this behaviour (specifically the error handling in the `case do ... end`) in the `Edgehog.Astarte.Device.DeploymentCommand` (and do the same for all those kinds of modules?) this way we can handle this as
```elixir
with :ok <- @deployment_command.send_deployment_command(...) do
  {:ok, device}
end
```
while still ensuring that the error is correctly handled by `send_deployment_command` itself
```elixir
@impl Edgehog.Astarte.Device.DeploymentCommand.Behaviour
  def send_deployment_command(client, device_id, deployment_id, command) do
    path = "/#{deployment_id}/command"
    case AppEngine.Devices.send_datastream(client, device_id, @interface, path, command) do
      {:error, api_error} -> ...
      :ok -> :ok
  end
```

backend/lib/edgehog/astarte/device/deployment_update.ex
```diff
@@ -0,0 +1,35 @@
...
+  @behaviour Edgehog.Astarte.Device.DeploymentUpdate.Behaviour
+
+  alias Astarte.Client.AppEngine
+
+  @interface "io.edgehog.devicemanager.apps.DeploymentUpdate"
+
+  @impl Edgehog.Astarte.Device.DeploymentUpdate.Behaviour
+  def update(client, device_id, from, to) do
+    data = %{from: from, to: to}
+    AppEngine.Devices.send_datastream(client, device_id, @interface, "/release", data)
```
lusergit, line 33: Maybe we should be consistent with this kind of modules; choosing whether we make these calls with the whole `request_data` (see #646 [`create_container_request` request data](https://github.com/edgehog-device-manager/edgehog/pull/646/files#diff-2b04344ec44f5192aab41c44036af5f1407b9fea3c58315f3a621cf0b2ea8558) ) or if we want to pass the individual fields

backend/lib/edgehog/devices/device/manual_actions/send_application_command.ex
noaccOS: I did it this way because that how it was done elsewhere and I didn't give it much thought, but yeah I also think doing it at the API level makes more sense (I'm actually unsure why the astarte client doesn't do that by default)

backend/lib/edgehog/astarte/device/deployment_update.ex
noaccOS: Passing individual fields every time quickly becomes unmaintainable (you don't really want to have functions with 7 parameters)
I agree, it's probably best to have a consistent interface between the api calls though, I'll add a structure for this interface

lusergit (approved):

backend/lib/edgehog/astarte/device/deployment_update.ex
```diff
@@ -0,0 +1,42 @@
...
+  @behaviour Edgehog.Astarte.Device.DeploymentUpdate.Behaviour
+
+  alias Astarte.Client.AppEngine
+  alias Edgehog.Error.AstarteAPIError
+
+  @interface "io.edgehog.devicemanager.apps.DeploymentUpdate"
+
+  @impl Edgehog.Astarte.Device.DeploymentUpdate.Behaviour
+  def update(client, device_id, data) do
+    api_call = AppEngine.Devices.send_datastream(client, device_id, @interface, "/release", data)
```
lusergit, line 33: The interface changed: it now needs to be called on the `/deployment` endpoint (see [interfaces PR](https://github.com//pull/64/files#diff-0482701a7ded078199e43f0fcfc2291ca0b384dc5be49ac52969a3d3938e1e17))
```suggestion
    api_call = AppEngine.Devices.send_datastream(client, device_id, @interface, "/deployment", data)
```

backend/lib/edgehog/astarte/device/deployment_update/request_data.ex
```diff
@@ -0,0 +1,32 @@
...
+defmodule Edgehog.Astarte.Device.DeploymentUpdate.RequestData do
+  @moduledoc false
+
+  alias Edgehog.Containers.Deployment
+
+  defstruct [:from, :to]
+
+  @type t :: %__MODULE__{
+          from: Deployment.t(),
+          to: Deployment.t()
```
lusergit, line 30: I think we want to send ids and not structs
```suggestion
          from: String.t(),
          to: String.t()
```

backend/lib/edgehog/devices/device/manual_actions/update_application.ex
```diff
@@ -0,0 +1,53 @@
...
+                     )
+
+  @impl Ash.Resource.ManualUpdate
+  def update(changeset, _opts, _context) do
+    from = changeset.arguments.from
+    to = changeset.arguments.to
+
+    device = changeset.data
+
+    with {:ok, from} <- fetch_deployment(device, from),
```
lusergit, line 41: I think here is better to stuck with `Ash` intended behaviour and fetch arguments trough `Ash.Changeset` functions
```suggestion
    device = changeset.data

    with {:ok, from} <- Ash.Changeset.fetch_argument(changeset, :from),
         {:ok, to} <- Ash.Changeset.fetch_argument(changeset, :to),
         {:ok, from} <- fetch_deployment(device, from),
```

backend/lib/edgehog/devices/device/manual_actions/update_application.ex
```diff
@@ -0,0 +1,53 @@
...
+         {:ok, to} <- fetch_deployment(device, to),
+         {:ok, device} <- Ash.load(device, :appengine_client),
+         data = %RequestData{from: from, to: to},
```
lusergit, line 44: Related to previous comment we should send ids to the device according to the interface, therefore here we have to extract them
```suggestion
         data = %RequestData{from: from.id, to: to.id},
```

backend/priv/astarte_resources/interfaces/io.edgehog.devicemanager.apps.DeploymentUpdate.json
lusergit, line 1: This interface has changes after IRL discussions, see [interfaces PR](https://github.com//pull/64/files#diff-0482701a7ded078199e43f0fcfc2291ca0b384dc5be49ac52969a3d3938e1e17))

lusergit (changes requested):

lusergit (approved):

lusergit (approved):

PR #655, opened 2024-10-21
chore(containers) release containers resource

### LINKED ISSUE #647: containers: Add `Containers.ReleaseContainers` resource
The resourse should rapresent the `many_to_many`relation between the `Release` and the `Container` resources, see [Ash documentation](https://hexdocs.pm/ash/relationships.html#many-to-many)

lusergit:
Closes #647


backend/lib/edgehog/containers/release_containers.ex
```diff
@@ -12,15 +39,18 @@ defmodule Edgehog.Containers.ReleaseContainers do
   relationships do
     belongs_to :release, Edgehog.Containers.Release do
       primary_key? true
+      allow_nil? false
+      attribute_type :uuid
     end
 
-    belongs_to :container, Edgehog.Containers.Containers do
+    belongs_to :container, Edgehog.Containers.Container do
```
noaccOS, line 46: It'd be better if you could correct the error in 9b22827ce0591f597a7a410419af97ba6af56dc2 instead (there was an extra s in the command)

backend/lib/edgehog/containers/release_containers.ex
```diff
@@ -1,9 +1,36 @@
...
+    domain: Edgehog.Containers
+
+  actions do
+    defaults [
+      :read,
+      :destroy,
+      create: [:release_id, :container_id],
+      update: [:release_id, :container_id]
+    ]
+  end
```
noaccOS, line 33: it doesn't really matter but actions can be added using the generator too: `--default-actions read,destroy,create`

backend/lib/edgehog/containers/release_containers.ex
```diff
@@ -12,15 +39,18 @@ defmodule Edgehog.Containers.ReleaseContainers do
   relationships do
     belongs_to :release, Edgehog.Containers.Release do
       primary_key? true
```
noaccOS, line 41: we usually add a single parameter as primary key, then in cases like this we add a unique index on the two fields (using an ash identity).
I'm not sure it's worth it in this case though, I'd ask @rbino

backend/lib/edgehog/containers/release_containers.ex
```diff
@@ -1,9 +1,36 @@
...
-    data_layer: AshPostgres.DataLayer
+  use Edgehog.MultitenantResource,
+    domain: Edgehog.Containers
+
+  actions do
+    defaults [
+      :read,
+      :destroy,
+      create: [:release_id, :container_id],
+      update: [:release_id, :container_id]
```
noaccOS, line 31: as the two parameters are part of the primary key, this update action does nothing.
but, even if they weren't: do we really want to allow an update of this resource? I personally think not, but right now I suggest we take the same approach as the graphql api: don't add anything right now, and build features as we need them

tl;dr: I suggest removing the default update action

backend/lib/edgehog/containers/release_containers.ex
```diff
@@ -12,15 +39,18 @@ defmodule Edgehog.Containers.ReleaseContainers do
...
-    belongs_to :container, Edgehog.Containers.Containers do
+    belongs_to :container, Edgehog.Containers.Container do
       primary_key? true
+      allow_nil? false
+      attribute_type :uuid
     end
   end
 
   postgres do
     table "release_containers"
```
noaccOS, line 54: table names are not namespaced so let's use a prefix: https://github.com/edgehog-device-manager/edgehog/pull/642#discussion_r1806133546

noaccOS (changes requested):
Remember to also add a commit with the migrations when generating resources: `mix ash_postgres.generate_migrations *migration_name*`

There are a couple issues to tackle (and some nitpicks which can honestly be ignored) but looks ok overall

backend/lib/edgehog/containers/release_containers.ex
lusergit: I think that since no additional fields are important in this resource (is essentially used to rapresent the many to many relationship) the pair  `(release_id, container_id)` can be the primary key

noaccOS: I agree. I've also found that's how we do it in [device_tag](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/labeling/device_tag.ex#L47) :+1:

backend/lib/edgehog/containers/release.ex
```diff
@@ -52,6 +52,12 @@ defmodule Edgehog.Containers.Release do
       through Edgehog.Containers.Deployment
       join_relationship :deployments
     end
+
+    many_to_many :containers, Edgehog.Containers.Release do
+      through Edgehog.Containers.ReleaseContainers
+      source_attribute_on_join_resource :release_id
+      destination_attribute_on_join_resource :container_id
```
noaccOS, line 59: are these needed?

noaccOS (approved):
Good job

PR #659, opened 2024-10-22
feat(containers): init container capability

### LINKED ISSUE #615: Add container_management capability
Add it to the list of capabilities [here](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/devices/device/types/capability.ex) and add it [here](https://github.com/edgehog-device-manager/edgehog/blob/1cd7df0671e9ad40c0bf17c4fb5316ad96e497b0/backend/lib/edgehog/capabilities.ex#L35) with the list of interfaces needed to support it (i.e. the new Astarte interfaces)

noaccOS:
adds the capability `container_management` if a device supports all the required interfaces.

closes #615 
~drafted because we don't currently support all required interfaces~

### DIFF
--- a/backend/lib/edgehog/devices/device/types/capability.ex
+++ b/backend/lib/edgehog/devices/device/types/capability.ex
@@ -26,6 +26,7 @@ defmodule Edgehog.Devices.Device.Types.Capability do
       battery_status: "The device provides information about its battery status.",
       cellular_connection: "The device provides information about its cellular connection.",
       commands: "The device supports commands, for example the rebooting command.",
+      container_management: "The device supports running applications using containers.",
       geolocation: "The device can be geolocated.",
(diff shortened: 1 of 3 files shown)


backend/lib/edgehog/devices/device/types/capability.ex
```diff
@@ -26,6 +26,7 @@ defmodule Edgehog.Devices.Device.Types.Capability do
       battery_status: "The device provides information about its battery status.",
       cellular_connection: "The device provides information about its cellular connection.",
       commands: "The device supports commands, for example the rebooting command.",
+      contoainer_management: "The device supports running applications using containers.",
```
rbino, line 29: Are Contoainers based on Dockoer?

PR #660, opened 2024-10-24
feat(containers): release deployment mechanism

### LINKED ISSUE #656: containers: implement release deployment
The first step consists of sending the required information to the device and then sending the start command once **all** acks have been received.
This can be achieved eg by using a [state machine](https://hexdocs.pm/gen_state_machine/GenStateMachine.html). An example usage would be the [push_rollout executor](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/update_campaigns/rollout_mechanism/push_rollout/executor.ex).

During the initialization of the deployment, all resources are already available to us: they are created first together with the release.

As a first step, we can assume the device does not already have any of the images/containers we need for the deployment. Handling the situation where the device already has partial information can be done later.

```mermaid
sequenceDiagram
    participant E as Edgehog
    participant D as Device

    E -->> D: CreateImageRequest
    E -->> D: CreateContainerRequest
    E -->> D: CreateDeploymentRequest

    D -->> E: AvailableImages /id/pulled(false)
    D -->> E: AvailableContainer /id/status(Received)
    D -->> E: DeploymentEvent /id/status(Idle)

    E ->> D: DeploymentCommand /id/command(start)
```

lusergit:
closes #656

The deployment mechanism consists of an `executor.ex` that handles the stages of the deployment, sending the appropriate calls to the device asyncronously. The `Devices` domain now exposes code interfaces for the `Device` resource

### DIFF
--- a/backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
+++ b/backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
@@ -37,8 +37,9 @@ defmodule Edgehog.Containers.ManualActions.SendDeployRequest do
       images = containers |> Enum.map(& &1.image) |> Enum.uniq()
 
       with :ok <- send_create_image_requests(device, images),
-           :ok <- send_create_container_requests(device, containers) do
-        Devices.send_create_deployment_request(device, deployment)
+           :ok <- send_create_container_requests(device, containers),
+           {:ok, _device} <- Devices.send_create_deployment_request(device, deployment) do
+        {:ok, deployment}
       end
     end
   end
@@ -46,7 +47,7 @@ defmodule Edgehog.Containers.ManualActions.SendDeployRequest do
   defp send_create_image_requests(device, images) do
     Enum.reduce_while(images, :ok, fn image, _acc ->
       case Devices.send_create_image_request(device, image) do
-        {:ok, _image} -> {:cont, :ok}
+        {:ok, _device} -> {:cont, :ok}
         {:error, reason} -> {:halt, {:error, reason}}
       end
     end)
@@ -55,7 +56,7 @@ defmodule Edgehog.Containers.ManualActions.SendDeployRequest do
   defp send_create_container_requests(device, containers) do
     Enum.reduce_while(containers, :ok, fn container, _acc ->
       case Devices.send_create_container_request(device, container) do
-        {:ok, _container} -> {:cont, :ok}
+        {:ok, _device} -> {:cont, :ok}
         {:error, reason} -> {:halt, {:error, reason}}
       end
     end)
--- a/backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
+++ b/backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
@@ -1,0 +1,63 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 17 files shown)


rbino (changes requested):
General comment: in this specific phase I would avoid having to deal with executor processes, waiting for acks and so on.

What I would do is wrap the logic that given a deployment sends all the stuff to Astarte in a generic action. This is similar to what is done [here](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/os_management/ota_operation/ota_operation.ex#L127) with `:send_ota_request`, which given an `OTARequest` database resource sends the appropriate message to Astarte. Of course the implementation of the Deployment would have more moving parts since it needs to send images and containers, but imho for now the idea would be to send everything without worrying about acks, but just short-circuiting (i.e. returning an error) if any request returns an error at any point.

With that, you can create a corresponding change (see the one for the `OTARequest` [here](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/os_management/ota_operation/changes/send_update_request.ex)) that can be called at the end of the `:deploy` action. Note that the change uses `after_action` instead of `after_transaction`, so if any of the requests fails, the whole action returns an error _and_ the `Deployment` is not persisted to the database.

backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
```diff
@@ -0,0 +1,50 @@
...
+    with {:ok, deployment} <- Ash.load(deployment, device: [], release: [containers: [:image]]) do
+      device = deployment.device
+
+      release = deployment.release
+      containers = release.containers
+      images = Enum.map(containers, & &1.image)
+
+      for image <- images do
+        Devices.send_create_image_request(device, image)
+      end
```
noaccOS, line 41: I'm not against this approach as a first step but:

The device returns an error if we send it a create image for an image it already has.
This means that this algorithm will fail if the same image is in use by multiple containers

This could be mitigated by using `containers |> Enum.map(& &1.image) |> Enum.uniq`, which might even be alright for now if we also include a comment explaining why.

Of course the real solution is to actually identify the resources we need to send requests for, but that will be done later :tm:

noaccOS (approved):

backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
```diff
@@ -0,0 +1,50 @@
...
+
+      for container <- containers do
+        Devices.send_create_container_request(device, container)
+      end
+
+      Devices.send_create_deployment_request(device, deployment)
```
rbino, line 47: From my previous long comment:

> [...] but just short-circuiting (i.e. returning an error) if any request returns an error at any point

i.e. even in this phase we can (and should) at least check if there's an error _during the request_ and bubble it up

```elixir
def run(input, _opts, _context) do
  # ...

  with :ok <- send_create_image_requests(device, images),
      :ok <- send_create_container_requests(device, containers) do
    Devices.send_create_deployment_request(device, deployment)
  end
end

defp send_create_image_requests(device, images) do
  # Usually short-circuiting is done using Enum.reduce_while/3, or with manual recursion
  Enum.reduce_while(images, :ok, fn image, _acc ->
    case Devices.send_create_image_request(device, image) do
      :ok -> {:cont, :ok}
      {:error, reason} -> {:halt, {:error, reason}}
    end
  end)
end

# Same for send_create_container_requests
```

rbino (changes requested):

lusergit:
The happy path now works with astarte 1.2.0 (locally tested). I've added the necessary fixes to make it work to this PR as the changes are needed in order to implement the release deployment mechanism feature

noaccOS:
please fixup 06015e0 onto 6e6ac3a

noaccOS (approved):

noaccOS:
I'll merge this as the error propagation was addressed. The code can still be modified with a followup pr in case.

PR #672, opened 2024-10-28
fix(containers): `send_create_image` accepts images and fetches arguments correctly

### LINKED ISSUE #671: containers: fix `send_create_image_request` update action
Currently the update action in [L33-L53 of `send_create_image_request`](https://github.com/edgehog-device-manager/edgehog/blob/6a7c8c0eb4ba5824ec57d8779ccbaa506a1c75bc/backend/lib/edgehog/devices/device/manual_actions/send_create_image_request.ex#L33-L53) fetches the image and the credentials from the changeset. This should not work, the correct approach is to use `Ash.Changeset.fetch_argument` (see [`send_create_deployment` L38-L39](https://github.com/edgehog-device-manager/edgehog/blob/6a7c8c0eb4ba5824ec57d8779ccbaa506a1c75bc/backend/lib/edgehog/devices/device/manual_actions/send_create_deployment.ex#L38-L39)). Moreover the action in [`device.ex` L244-L260](https://github.com/edgehog-device-manager/edgehog/blob/6a7c8c0eb4ba5824ec57d8779ccbaa506a1c75bc/backend/lib/edgehog/devices/device/device.ex#L244-L260) should not need the image credentials as they can be fetched from the `Image` argument itself

lusergit:
Closes #671


backend/lib/edgehog/devices/device/manual_actions/send_create_image_request.ex
```diff
@@ -34,21 +34,23 @@ defmodule Edgehog.Devices.Device.ManualActions.SendCreateImageRequest do
...
+        credentials: credentials
+      }
 
-    with {:ok, device} <- Ash.load(device, :appengine_client) do
-      @send_create_image_request_behaviour.send_create_image_request(
-        device.appengine_client,
-        device.device_id,
-        data
-      )
+      with {:ok, device} <- Ash.load(device, :appengine_client) do
```
noaccOS, line 47: since the order here is not important, we can group this instruction together with the previous ash functions in order to have a single `with` block

noaccOS (approved):

PR #673, opened 2024-10-28
trigger_payload: accept trigger_name key

### LINKED ISSUE #664: Allow receiving `trigger_name` in `TriggerPayload`
Astarte 1.2.0 added the [`trigger_name`](https://github.com/astarte-platform/astarte/pull/923) field to the HTTP trigger payload. Ash returns an error if unexpected inputs are received, so currently it returns an error if a trigger from an Astarte 1.2.0 instance is received.

This can be handled either explicitly adding `trigger_name` to the attributes [here](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/triggers/trigger_payload.ex#L28) or adding an explicit `:create` action with [`skip_unknown_inputs: :*`](https://hexdocs.pm/ash/dsl-ash-resource.html#actions-create-skip_unknown_inputs)

rbino:
Astarte >= 1.2.0 sends an additional trigger_name key in the trigger payload. This was making Edgehog crash since Ash doesn't accept additional input keys by default. Fix this allowing unknown keys to be accepted. This should future-proof the trigger handler in case further additional keys are added in the future.

Since 1.2.0 is the current stable version of Astarte, add the trigger_name key to all existing tests and explicitly test support for the key missing to avoid regressions with old Astarte versions.

Fix #664

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.9.1] - Unreleased
+### Fixed
+- Allow receiving `trigger_name` key in trigger payload, which is sent by Astarte >= 1.2.0.
+
 ## [0.9.0] - 2024-10-25
 ### Fixed
 - Correctly support automatic login attempts on the frontend regardless of existing auth sessions ([#596](https://github.com/edgehog-device-manager/edgehog/pull/596)).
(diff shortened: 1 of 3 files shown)


davidebriani (changes requested):
Thanks for the fix, the PR is very welcome.

Would you care adding references to the originating issue?
It would be nice to have a quick reference to #664 in the commit, PR description, and CHANGELOG line.

Besides that, as noted in the issue we could also use [skip_unknown_inputs](https://hexdocs.pm/ash/dsl-ash-resource.html#actions-create-skip_unknown_inputs) so that future additions to Astarte APIs will not break existing Edgehog deployments.

rbino:
Added the references in both places.
Let me know if you prefer handling this with `skip_unknown_inputs`, either now or in the future (in that case open an issue).

davidebriani (approved):

PR #681, opened 2024-10-30
Feat/applications query

### LINKED ISSUE #665: containers: applications query
a graphql query `applications` should be available. it should return the list of applications

input: _
output: [application]

lusergit:
Add the `applications` query for the frontend, closes #665


backend/test/edgehog_web/schema/query/list_applications_test.exs
```diff
@@ -0,0 +1,69 @@
...
+
+      data = [tenant: tenant] |> list_applications() |> extract_result!()
+
+      assert %{"applications" => %{"results" => applications}} = data
+
+      names = Enum.map(applications, & &1["name"])
+
+      for app <- [app1, app2] do
+        assert app.name in names
+      end
```
noaccOS, line 39: my problem with the current solution is that `applications` _may_ be a superset of `[app1, app2]`
```suggestion
      names = applications |> Enum.map(& &1["name"]) |> Enum.sort()
      expected_names = [app1, app2] |> Enum.map(& &1.name) |> Enum.sort()
      
      assert names == expected_names
```

PR #685, opened 2024-11-04
chore(containers): volumes resource

lusergit:
Add volumes management on the backend side


backend/lib/edgehog/containers/volume.ex
```diff
@@ -0,0 +1,68 @@
...
+      create: [:label, :driver, :options],
+      update: [:label, :driver, :options]
+    ]
+  end
+
+  attributes do
+    uuid_primary_key :id
+
+    attribute :driver, :string do
+      constraints allow_empty?: true
```
noaccOS, line 44: I don't think it can be empty (?)

backend/lib/edgehog/containers/volume.ex
```diff
@@ -0,0 +1,68 @@
...
+
+    attribute :label, :string do
+      allow_nil? false
+    end
+
+    timestamps()
+  end
+
+  identities do
+    identity :volume_label, [:label]
```
noaccOS, line 66: let's just call it `:label` since it's only one field

backend/lib/edgehog/containers/volume.ex
```diff
@@ -0,0 +1,68 @@
...
+    end
+
+    attribute :options, :map do
+      default %{}
+      allow_nil? false
+    end
+
+    attribute :label, :string do
+      allow_nil? false
+    end
```
noaccOS, line 56: I'm not convinced we want or need this, at least for now.

In any case, Docker does have a human-readable identifier for volumes but
1. it's called [name](https://docs.docker.com/reference/api/engine/version/v1.43/#tag/Volume/operation/VolumeCreate). `label` holds another meaning.
2. It's auto generated when not given. This could be implemented later though

backend/lib/edgehog/containers/volume.ex
```diff
@@ -0,0 +1,63 @@
...
+
+defmodule Edgehog.Containers.Volume do
+  @moduledoc false
+  use Edgehog.MultitenantResource,
+    domain: Edgehog.Containers,
+    extensions: [AshGraphql.Resource]
+
+  graphql do
+    type :volume
+  end
```
noaccOS, line 29: I'm not sure we want to expose volumes via graphql, I think the schema has already been finalized as is?

backend/lib/edgehog/containers/volume/options_calculation.ex
```diff
@@ -0,0 +1,35 @@
...
+  use Ash.Resource.Calculation
+
+  @impl Ash.Resource.Calculation
+  def calculate(records, _opts, _context) do
+    Enum.map(records, &encode_options(&1.options))
+  end
+
+  defp encode_options(options) do
+    Enum.map(options, fn {key, value} ->
+      String.upcase(key) <> "=" <> value
```
noaccOS, line 32: After some IRL discussions, we've decided the best approach to be to not modify the casing at all.
It's up to the user to input the correct values

noaccOS (approved):

PR #692, opened 2024-11-06
feat(containers): isolate containers by network

noaccOS:
send one network for each container to the device, so that they cannot talk to each other

depends on #690 
depends on #691

### DIFF
--- a/backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
+++ b/backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
@@ -35,7 +35,7 @@ defmodule Edgehog.Devices.Device.ManualActions.SendCreateContainer do
     device = changeset.data
 
     with {:ok, container} <- Ash.Changeset.fetch_argument(changeset, :container),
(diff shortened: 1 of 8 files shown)


backend/lib/edgehog/devices/device/manual_actions/send_create_container.ex
```diff
@@ -35,7 +35,7 @@ defmodule Edgehog.Devices.Device.ManualActions.SendCreateContainer do
     device = changeset.data
 
     with {:ok, container} <- Ash.Changeset.fetch_argument(changeset, :container),
-         {:ok, container} <- Ash.load(container, [:env_encoding, :image]),
+         {:ok, container} <- Ash.load(container, [:env_encoding, :image, networks: [:id]]),
```
lusergit, line 38: very nit:
```suggestion
         {:ok, container} <- Ash.load(container, [:env_encoding, :image, :networks]),
```
calling `Ash.load` with just `:networks` works, as networks ids are not a nested calculation or relationship.

lusergit (approved):

PR #695, opened 2024-11-08
Deployment: fix GraphQL args for deploy_release mutation

davidebriani:
Instead of accepting the resource IDs used internally, expect Relay IDs as the arguments of the mutation.
Indeed, GraphQL queries expose resources with Relay global IDs, so a translation step should be done when accepting those IDs in GraphQL mutations before they are used with Ash actions.

### DIFF
--- a/backend/lib/edgehog/containers/deployment.ex
+++ b/backend/lib/edgehog/containers/deployment.ex
@@ -87,6 +87,7 @@ defmodule Edgehog.Containers.Deployment do
 
   relationships do
     belongs_to :device, Edgehog.Devices.Device do
+      attribute_type :id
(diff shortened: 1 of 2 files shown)


backend/lib/edgehog/containers/deployment.ex
```diff
@@ -87,6 +87,7 @@ defmodule Edgehog.Containers.Deployment do
 
   relationships do
     belongs_to :device, Edgehog.Devices.Device do
+      attribute_type :id
```
noaccOS, line 90: is this needed? I think this should be the default behavior

noaccOS (approved):

PR #696, opened 2024-11-08
Triggers: fix support for AvailableDeployments trigger

davidebriani:
- Add a trigger definition between the templates of the triggers that Edgehog automatically installs on the Astart realm.
- Fix the event handler relative to the trigger, since it was mistakenly expecting a different shape for the event payload.


lusergit (approved):

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -116,11 +116,16 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+        status = event.value
 
-    with {:ok, deployment} <- Containers.fetch_deployment(deployment_id, tenant: tenant) do
-      Containers.deployment_set_status(deployment, status, tenant: tenant)
+        with {:ok, deployment} <- Containers.fetch_deployment(deployment_id, tenant: tenant) do
+          Containers.deployment_set_status(deployment, status, tenant: tenant)
+        end
+
+      _ ->
+        {:ok, nil}
```
noaccOS, line 128: If the path is not what we expect, I think we should fail

```suggestion
        {:error, :invalid_event_path}
```

noaccOS (approved):

PR #711, opened 2024-11-13
chore(containers): support the latest version of the interfaces

noaccOS:


backend/lib/edgehog/containers/container.ex
```diff
@@ -84,6 +84,12 @@ defmodule Edgehog.Containers.Container do
       public? true
     end
 
+    attribute :network_mode, :string do
+      default "bridge"
+      public? true
+      allow_nil? false
+    end
```
davidebriani, line 107: Let's put this attribute in create/update actions

Let's also:
- specify `networkMode` when communicating with the device (request_data + send_create_container).
- avoid specifying `networks` when communicating with the device  (request_data + send_create_container), and instead use `networkIds`

davidebriani (changes requested):

backend/priv/astarte_resources/interfaces/io.edgehog.devicemanager.apps.CreateNetworkRequest.json
```diff
@@ -43,8 +38,18 @@
...
       "doc": "Enable IPv6 on the network."
+    },
+    {
+      "endpoint": "/network/options",
+      "type": "stringarray",
+      "database_retention_policy": "use_ttl",
+      "database_retention_ttl": 31556952,
+      "reliability": "guaranteed",
+      "description": "Network driver options",
+      "doc": "An array of key=value options to set for the driver."
```
davidebriani, line 52: Related to this interface, we should change the code in the module `Edgehog.Astarte.Device.CreateNetworkRequest` so that it interacts with the interface path `/network`, instead of `/container`.

davidebriani (changes requested):

davidebriani (approved):

PR #712, opened 2024-11-13
feat(containers): send one network for each release

noaccOS:
the network is set so that it does not have internet access and
containers outside the network cannot communicate with containers inside
it

### DIFF
--- a/backend/lib/edgehog/containers/changes/create_default_network.ex
+++ b/backend/lib/edgehog/containers/changes/create_default_network.ex
@@ -1,0 +1,69 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


backend/lib/edgehog/astarte/device/create_network_request/request_data.ex
```diff
@@ -24,16 +24,16 @@ defmodule Edgehog.Astarte.Device.CreateNetworkRequest.RequestData do
...
   ]
 
   @type t() :: %__MODULE__{
           id: String.t(),
           driver: String.t(),
-          checkDuplicate: boolean(),
           internal: boolean(),
-          enableIpv6: boolean()
+          enableIpv6: boolean(),
+          options: [String.t()]
```
lusergit, line 37: i think this should be
```suggestion
          options: list(String.t())
```

lusergit (changes requested):

lusergit (approved):

backend/lib/edgehog/astarte/device/create_network_request/request_data.ex
noaccOS: the two syntaxes are equivalent

backend/priv/repo/migrations/20241113173040_update_interfaces.exs
```diff
@@ -0,0 +1,59 @@
...
+      add :options, {:array, :text}, null: false, default: []
+    end
+
+    alter table(:containers) do
+      add :network_mode, :text, null: false, default: "bridge"
+    end
+
+    alter table(:application_deployments) do
+      modify :device_id, :integer
+    end
```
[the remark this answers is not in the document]
noaccOS: You're right, but unfortunately the change was made in the previous pr (#711) which has already been merged.
I've opened #714 to fix the issue

lusergit (approved):

PR #718, opened 2024-11-14
Fixtures: rework setup of container resources

davidebriani:
This commit changes how the seeds.exs file creates the resources of the Applications feature:
- Avoid using test fixtures, since those modules are only built in the test environment. This fixes the issue where running the seeding routing crashed when using docker compose.
- Fix the Application's default create action so that it correctly reuses existing Image resources, identified by their `reference` field.
- Ensure that `port_bindings` and `network_mode` attributes are accepted when specifying the containers to be created.


noaccOS (approved):
I'd rather have two separate commits (one for the container, one for the seeds) but it's nitpicking

PR #724, opened 2024-11-18
chore(containers): update status

lusergit:
Deployment status update at each container interface trigger, status interface richer, expresses new deployment statuses:
- `:created` when the backend has recived the deployment description
- `:sent` when the deployment description has been sent to the device
- `:puled_images` when the device has created all the image resources
- `:created_networks` when the device has set up all the network resources
- `:created_containers` when the device has created all the container resources
- `:created_deployment` when the device has set up the deployment resource
- `:ready` when all resources have been created and the deployment is ready to be started

### DIFF
--- a/backend/lib/edgehog/containers/deployment.ex
+++ b/backend/lib/edgehog/containers/deployment.ex
@@ -102,12 +102,29 @@ defmodule Edgehog.Containers.Deployment do
     update :set_status do
       accept [:status, :message]
     end
+
+    update :update_status do
+      change Changes.CheckImages
+      change Changes.CheckNetworks
+      change Changes.CheckContainers
+      change Changes.CheckDeployments
+
+      require_atomic? false
+    end
+
+    read :filter_by_release do
+      argument :release_id, :uuid
+
+      filter expr(release_id == ^arg(:release_id))
+    end
   end
 
   attributes do
     uuid_primary_key :id
 
     attribute :status, DeploymentStatus do
+      allow_nil? false
+      default :created
       public? true
     end
 
--- a/backend/lib/edgehog/containers/deployment/changes/check_images.ex
+++ b/backend/lib/edgehog/containers/deployment/changes/check_images.ex
@@ -1,0 +1,55 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Containers.Deployment.Changes.CheckImages do
+  @moduledoc false
+  use Ash.Resource.Change
+
+  alias Edgehog.Devices
+
+  @impl Ash.Resource.Change
+  def change(changeset, _opts, context) do
+    %{tenant: tenant} = context
+    deployment = changeset.data
+
+    with :sent <- deployment.status,
+         {:ok, deployment} <-
+           Ash.load(deployment, [:device, release: [containers: [:image]]], reuse_values?: true),
+         {:ok, available_images_statuses} <-
+           Devices.available_images(deployment.device, tenant: tenant) do
+      available_images_ids =
+        Enum.map(available_images_statuses, & &1.id)
+
+      missing_images =
+        deployment.release.containers
+        |> Enum.map(& &1.image.id)
+        |> Enum.reject(&(&1 in available_images_ids))
+
+      if missing_images == [] do
+        Ash.Changeset.change_attribute(changeset, :status, :pulled_images)
+      else
+        changeset
+      end
+    else
+      _ ->
+        changeset
+    end
+  end
+end
--- a/backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
+++ b/backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
@@ -23,11 +23,13 @@ defmodule Edgehog.Containers.ManualActions.SendDeployRequest do
 
   use Ash.Resource.Actions.Implementation
 
+  alias Edgehog.Containers
   alias Edgehog.Devices
 
   @impl Ash.Resource.Actions.Implementation
-  def run(input, _opts, _context) do
+  def run(input, _opts, context) do
     deployment = input.arguments.deployment
+    %{tenant: tenant} = context
 
     with {:ok, deployment} <-
            Ash.load(deployment, device: [], release: [containers: [:image, :networks]]) do
@@ -46,7 +48,7 @@ defmodule Edgehog.Containers.ManualActions.SendDeployRequest do
            :ok <- send_create_container_requests(device, containers),
            :ok <- send_create_network_requests(device, networks),
            {:ok, _device} <- Devices.send_create_deployment_request(device, deployment) do
-        {:ok, deployment}
+        Containers.deployment_set_status(deployment, :sent, nil, tenant: tenant)
       end
     end
   end
--- a/backend/lib/edgehog/containers/types/deployment_status.ex
+++ b/backend/lib/edgehog/containers/types/deployment_status.ex
@@ -27,7 +27,16 @@ defmodule Edgehog.Containers.Types.DeploymentStatus do
       started: "The deployment is running.",
       starting: "The deployment is starting.",
       stopped: "The deployment has stopped.",
-      stopping: "The deploymen is stopping."
+      stopping: "The deploymen is stopping.",
+      created: "The deployment has been received by the backend and will be sent to the device.",
+      sent: "All the necessary resources have been sent to the device.",
+      ready: "The deployment is ready on the device. All necessary resources are available.",
+      # TODO: these are internal states that should not be exposed.
+      # Remove when reimplementing the deployment and its status as a state machine
+      pulled_images: "The device is currently pulling the necessary images for the deployment.",
+      created_networks: "The device is setting up the networks necessary for the deployment.",
+      created_containers: "The device is setting up the containers necessary for the deployment.",
+      created_deployment: "The device is setting up the release of the the deployment."
     ]
 
   def graphql_type(_), do: :application_deployment_status
--- a/backend/lib/edgehog/devices/devices.ex
+++ b/backend/lib/edgehog/devices/devices.ex
@@ -108,8 +108,10 @@ defmodule Edgehog.Devices do
 
       define :update_application, action: :update_application, args: [:from, :to]
 
-      define_calculation :available_images
-      define_calculation :available_containers
+      define_calculation :available_images, args: [:_record]
+      define_calculation :available_containers, args: [:_record]
+      define_calculation :available_networks, args: [:_record]
+      define_calculation :available_deployments, args: [:_record]
     end
 
     resource HardwareType
--- a/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
+++ b/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
@@ -32,7 +32,10 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
   alias Edgehog.Triggers.IncomingData
   alias Edgehog.Triggers.TriggerPayload
 
+  @available_containers "io.edgehog.devicemanager.apps.AvailableContainers"
   @available_deployments "io.edgehog.devicemanager.apps.AvailableDeployments"
+  @available_images "io.edgehog.devicemanager.apps.AvailableImages"
+  @available_networks "io.edgehog.devicemanager.apps.AvailableNetworks"
   @deployment_event "io.edgehog.devicemanager.apps.DeploymentEvent"
   @ota_event "io.edgehog.devicemanager.OTAEvent"
   @ota_response "io.edgehog.devicemanager.OTAResponse"
@@ -106,6 +109,77 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
     |> Ash.create(tenant: tenant)
   end
 
+  defp handle_event(%IncomingData{interface: @available_images} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", image_id, "pulled"] ->
+        containers = Containers.containers_with_image!(image_id, tenant: tenant)
+
+        releases =
+          containers
+          |> Enum.flat_map(&Containers.releases_with_container!(&1.id, tenant: tenant, load: :release))
+          |> Enum.map(& &1.release_id)
+          |> Enum.uniq()
+
+        deployments =
+          releases
+          |> Enum.flat_map(&Containers.deployments_with_release!(&1, tenant: tenant))
+          |> Enum.uniq_by(& &1.id)
+
+        {:ok, Enum.map(deployments, &Containers.deployment_update_status!/1)}
+
+      _ ->
+        {:error, :invalid_event_path}
+    end
+  end
+
+  defp handle_event(%IncomingData{interface: @available_networks} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", network_id, "pulled"] ->
+        containers =
+          network_id
+          |> Containers.containers_with_network!(tenant: tenant, load: :container)
+          |> Enum.map(& &1.container_id)
+          |> Enum.uniq()
+
+        releases =
+          containers
+          |> Enum.flat_map(&Containers.releases_with_container!(&1, tenant: tenant, load: :release))
+          |> Enum.map(& &1.release_id)
+          |> Enum.uniq()
+
+        deployments =
+          releases
+          |> Enum.flat_map(&Containers.deployments_with_release!(&1, tenant: tenant))
+          |> Enum.uniq_by(& &1.id)
+
+        {:ok, Enum.map(deployments, &Containers.deployment_update_status!/1)}
+
+      _ ->
+        {:error, :invalid_event_path}
+    end
+  end
+
+  defp handle_event(%IncomingData{interface: @available_containers} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", container_id, "status"] ->
+        releases =
+          container_id
+          |> Containers.releases_with_container!(tenant: tenant, load: :release)
+          |> Enum.map(& &1.release_id)
+          |> Enum.uniq()
+
+        deployments =
+          releases
+          |> Enum.flat_map(&Containers.deployments_with_release!(&1, tenant: tenant))
+          |> Enum.uniq_by(& &1.id)
+
+        {:ok, Enum.map(deployments, &Containers.deployment_update_status!/1)}
+
+      _ ->
+        {:error, :invalid_event_path}
+    end
+  end
+
   defp handle_event(%IncomingData{interface: @deployment_event} = event, tenant, _realm_id, _device_id, _timestamp) do
     "/" <> deployment_id = event.path
 
@@ -139,7 +213,7 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
           if status == nil do
             Containers.delete_deployment(deployment)
           else
-            Containers.deployment_set_status(deployment, status, tenant: tenant)
+            Containers.deployment_set_status(deployment, status, deployment.message, tenant: tenant)
           end
         end
 
--- a/backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
+++ b/backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
@@ -26,6 +26,14 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
   import Edgehog.DevicesFixtures
   import Edgehog.OSManagementFixtures
 
+  alias Edgehog.Astarte.Device.AvailableContainers.ContainerStatus
+  alias Edgehog.Astarte.Device.AvailableContainersMock
+  alias Edgehog.Astarte.Device.AvailableDeployments.DeploymentStatus
+  alias Edgehog.Astarte.Device.AvailableDeploymentsMock
+  alias Edgehog.Astarte.Device.AvailableImages.ImageStatus
+  alias Edgehog.Astarte.Device.AvailableImagesMock
+  alias Edgehog.Astarte.Device.AvailableNetworks.NetworkStatus
+  alias Edgehog.Astarte.Device.AvailableNetworksMock
   alias Edgehog.Astarte.Device.DeviceStatusMock
   alias Edgehog.Containers.Deployment
   alias Edgehog.Devices.Device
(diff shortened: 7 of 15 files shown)


backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,72 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+          deployments =
+            releases
+            |> Enum.reduce_while([], &deployments_from_release(&1, &2, tenant))
+            |> Enum.uniq()
+
+          Enum.map(deployments, &Containers.deployment_update_satus(&1))
+        end
+
+      _ ->
+        :ok
```
noaccOS, line 130: the case where we don't know the path should be an error. see https://github.com/edgehog-device-manager/edgehog/pull/696#discussion_r1840131511

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,72 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+    case String.split(event.path, "/") do
+      ["", network_id, "pulled"] ->
+        with {:ok, relations} <- Containers.containers_with_network(network_id, tenant: tenant),
+             {:ok, relations} <- Ash.load(relations, [:container]) do
+          containers = relations |> Enum.map(& &1.container) |> Enum.uniq()
+
+          releases =
+            containers
+            |> Enum.reduce_while([], &releases_from_container(&1, &2, tenant))
+            |> Enum.uniq()
```
noaccOS, line 144: Even if something errors out, I'd say we want to update the status of the containers we were able to gather. And possibly logging the error:

```suggestion
          releases =
            containers
            |> Enum.map(&Containers.releases_with_container(&1.id, tenant: tenant))
            |> Enum.flat_map(fn
              {:ok, releases} -> releases
              {:error, reason} -> 
                Logger.error("Error reading releases from containers: #{inspect(reason)}")
                []
            end)
            |> Enum.uniq()
```

backend/lib/edgehog/containers/manual_actions/update_status.ex
```diff
@@ -0,0 +1,138 @@
...
+defmodule Edgehog.Containers.ManualActions.UpdateStatus do
+  @moduledoc false
+  use Ash.Resource.ManualUpdate
+
+  alias Edgehog.Devices
+
+  @impl Ash.Resource.ManualUpdate
+  def update(changeset, _opts, context) do
+    changeset = check_images(changeset, context)
+    Ash.update(changeset)
```
noaccOS, line 30: I don't like the current approach: reading these lines, it only looks like you're checking the images.
I would rather see all actions explicitly top level:

```suggestion
    changeset
    |> check_images(context)
    |> check_networks(context)
    |> check_containers(context)
    |> check_deployments(context)
    |> Ash.update()
```

Being changesets, it's easy to apply the next action only if the one before has succeeded. I see two possible ways to approach this:
- by using `Ash.Changeset.fetch_argument_or_change` to get the current status of the changeset (which could have been updated by previous actions), and only going forward with the check if the status is the expected status
- adding an error to the changeset in case not all resources are available. this allows removing all intermediate states (pulling_images, creating_networks, ...) if we want to go in that direction, but changes the semantic meaning of a failed `UpdateStatus.update`

backend/lib/edgehog/containers/manual_actions/update_status.ex
```diff
@@ -0,0 +1,138 @@
...
+        |> Enum.filter(& &1.pulled)
+        |> Enum.map(& &1.id)
+
+      missing_images = Enum.reject(images, fn image -> image.id not in pulled_images_ids end)
+
+      if missing_images == [] do
+        check_networks(changeset, context)
+      else
+        changeset
+      end
```
noaccOS, line 56: 1. As a rule of thumb, do not do double negations (`reject` and `not in`)
2. Reason being, I think you only need one negation here (you want to take (ie `filter`) all images `not in` pulled images)

```elixir
iex(1)> images = [1,2,3,4,5]
[1, 2, 3, 4, 5]
iex(2)> pulled = [2,4,1]
[2, 4, 1]
iex(3)> images |> Enum.reject(& &1 not in pulled)
[1, 2, 4]
```

3. Bonus, but you're treating lists like sets: you can use sets
```suggestion
      deployment_images = deployment.release.containers |> Enum.map(& &1.image.id) |> MapSet.new()

      available_images =
        deployment.device
        |> Devices.available_images(tenant: tenant)
        |> Enum.filter(& &1.pulled)
        |> Enum.map(& &1.id)
        |> MapSet.new()
        
      all_images_available? = MapSet.subset?(deployment_images, available_images)

      if all_images_available? do
        changeset
      else
        check_networks(changeset, context)
      end
```

backend/lib/edgehog/containers/types/deployment_status.ex
```diff
@@ -27,7 +27,14 @@ defmodule Edgehog.Containers.Types.DeploymentStatus do
...
       starting: "The deployment is starting.",
       stopped: "The deployment has stopped.",
-      stopping: "The deploymen is stopping."
+      stopping: "The deploymen is stopping.",
+      created: "The deployment has been received by the backend and will be sent to the device.",
+      sent: "All the necessary resources have been sent to the device.",
+      pulling_images: "The device is currently pulling the necessary images for the deployment.",
+      creating_networks: "The device is setting up the networks necessary for the deployment.",
+      creating_containers: "The device is setting up the containers necessary for the deployment.",
+      creating_deployment: "The device is setting up the release of the the deployment.",
```
noaccOS, line 36: I personally don't like the idea of exposing these states because they're internal states and don't necessarily reflect the status for the device (a device could already have all containers created but still missing some images)

also, the device only pulls images once we actually start the deployment, so that would be "creating_images"

still, we plan to change the implementation for good once we have time to do so, so if you need these intermediary states I'm not _that_ against them

noaccOS (changes requested):

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -450,12 +457,163 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
 
-      assert {:ok, deployment} = Ash.get(Deployment, deployment.id, tenant: tenant)
+      assert {:ok, release} = Ash.load(release, containers: [:image, :networks])
+      assert [container] = release.containers
+
+      deployment =
+        [tenant: tenant, device_id: device.id, release_id: release.id]
+        |> deployment_fixture()
+        |> Ash.Changeset.for_update(:set_status, %{status: :sent}, tenant: tenant)
+        |> Ash.update!()
```
noaccOS, line 472: I believe you can just

```suggestion
        [tenant: tenant, device_id: device.id, release_id: release.id, status: :sent]
        |> deployment_fixture()
```

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -450,6 +450,36 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
+      conn =
+        conn
+        |> put_req_header("astarte-realm", realm.name)
+        |> post(path, deployment_event)
+
+      assert response(conn, 200)
+
+      assert {:ok, deployment} == Ash.get(Deployment, deployment.id, tenant: tenant)
+    end
+
```
noaccOS: As this was heavily modified in the following commit, could this be squashed inside it?
TBF tests could go into the same commit as the change, so that could also be an option

lusergit: Yes, will do it :+1:

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
lusergit: Actually no since the deployment creation does not allow to set the initial status. I think this is the inteded behaviour since the status is an internal rapresentation that depends on the communication with the device. I'll keep it like this for now insted of creating a new `create` action for the deployment just for this test

backend/lib/edgehog/containers/deployment.ex
```diff
@@ -49,6 +49,7 @@ defmodule Edgehog.Containers.Deployment do
       end
 
       change manage_relationship(:device_id, :device, type: :append)
+      change set_attribute(:status, :created)
```
noaccOS, line 52: how about we instead set the default value to `:created` and disallow `nil`?

lusergit: I was thinking the same thing :+1:

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -124,7 +177,7 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
           if status == nil do
             Containers.delete_deployment(deployment)
           else
-            Containers.deployment_set_status(deployment, status, tenant: tenant)
+            Containers.deployment_update_satus(deployment)
```
noaccOS, line 180: This change should be applied to available deployments (probably also only on status: stopped but whatever)
Deployment events can only happen after we have sent a start, or if the device encountered an error during creation

In both circumstances, we want to set the state it's given us

Also, just noticed: `update_satus`

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
     |> Ash.create(tenant: tenant)
   end
 
+  defp handle_event(%IncomingData{interface: @available_images} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", image_id, "pulled"] ->
+        with {:ok, containers} <- Containers.containers_with_image(image_id, tenant: tenant),
+             {:ok, releases} <-
+               Enum.reduce_while(containers, {:ok, []}, &releases_from_container(&1, &2, tenant)),
```
noaccOS, line 117: https://github.com/edgehog-device-manager/edgehog/pull/724#discussion_r1846780287

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+      _ ->
+        {:error, :invalid_event_path}
+    end
+  end
+
+  defp handle_event(%IncomingData{interface: @available_networks} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", network_id, "pulled"] ->
+        with {:ok, relations} <- Containers.containers_with_network(network_id, tenant: tenant),
+             {:ok, relations} <- Ash.load(relations, [:container]),
```
noaccOS, line 132: https://github.com/edgehog-device-manager/edgehog/pull/724#discussion_r1850362629

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+      _ ->
+        {:error, :invalid_event_path}
+    end
+  end
+
+  defp handle_event(%IncomingData{interface: @available_containers} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", container_id, "status"] ->
+        with {:ok, relations} <- Containers.releases_with_container(container_id, tenant: tenant),
+             {:ok, relations} <- Ash.load(relations, :release),
```
noaccOS, line 150: https://github.com/edgehog-device-manager/edgehog/pull/724#discussion_r1850362629

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
   end
 
+  defp handle_event(%IncomingData{interface: @available_images} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", image_id, "pulled"] ->
+        with {:ok, containers} <- Containers.containers_with_image(image_id, tenant: tenant),
+             {:ok, releases} <-
+               Enum.reduce_while(containers, {:ok, []}, &releases_from_container(&1, &2, tenant)),
+             {:ok, deployments} <-
+               Enum.reduce_while(releases, {:ok, []}, &deployments_from_release(&1, &2, tenant)) do
```
noaccOS, line 119: https://github.com/edgehog-device-manager/edgehog/pull/724#discussion_r1846780287

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+          {:ok, Enum.map(deployments, &Containers.deployment_update_status(&1))}
```
noaccOS, line 120: Just a bit clearer

```suggestion
          {:ok, Enum.map(deployments, &Containers.deployment_update_status/1)}
```

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+  defp handle_event(%IncomingData{interface: @available_networks} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", network_id, "pulled"] ->
+        with {:ok, relations} <- Containers.containers_with_network(network_id, tenant: tenant),
+             {:ok, relations} <- Ash.load(relations, [:container]),
+             containers = relations |> Enum.map(& &1.container) |> Enum.uniq(),
+             {:ok, releases} <-
+               Enum.reduce_while(containers, {:ok, []}, &releases_from_container(&1, &2, tenant)),
+             {:ok, deployments} <-
+               Enum.reduce_while(releases, {:ok, []}, &deployments_from_release(&1, &2, tenant)) do
```
noaccOS, line 137: https://github.com/edgehog-device-manager/edgehog/pull/724#discussion_r1846780287

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+          {:ok, Enum.map(deployments, &Containers.deployment_update_status(&1))}
```
noaccOS, line 138: /1

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+  end
+
+  defp handle_event(%IncomingData{interface: @available_containers} = event, tenant, _realm_id, _device_id, _timestamp) do
+    case String.split(event.path, "/") do
+      ["", container_id, "status"] ->
+        with {:ok, relations} <- Containers.releases_with_container(container_id, tenant: tenant),
+             {:ok, relations} <- Ash.load(relations, :release),
+             releases = relations |> Enum.map(& &1.release) |> Enum.uniq(),
+             {:ok, deployments} <-
+               Enum.reduce_while(releases, {:ok, []}, &deployments_from_release(&1, &2, tenant)) do
```
noaccOS, line 153: https://github.com/edgehog-device-manager/edgehog/pull/724#discussion_r1846780287

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +109,56 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+          {:ok, Enum.map(deployments, &Containers.deployment_update_status(&1))}
```
noaccOS, line 154: /1

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -128,10 +181,17 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
           end
         end
 
+      ["", deployment_id, "message"] ->
+        message = event.value
+
+        with {:ok, deployment} <- Containers.fetch_deployment(deployment_id, tenant: tenant) do
+          Containers.deployment_set_status(deployment, deployment.status, message, tenant: tenant)
+        end
+
```
noaccOS, line 194: available deployments does not have a message. only deployment event does

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -205,4 +265,24 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+    end
+  end
+
+  defp deployments_from_release(release, {:ok, deployments}, tenant) do
+    case Containers.deployments_with_release(release.id, tenant: tenant) do
+      {:ok, new_deployments} -> {:cont, {:ok, Enum.uniq(deployments ++ new_deployments)}}
+      error -> {:halt, {:error, error}}
+    end
+  end
 end
```
noaccOS, line 288: as a rule of thumb if your function is not an abstraction, it should probably not be a function
of course this helps when your function grows too much, but it's usually better to look for other methods of refactoring.

just a small tip, I expect these functions to be reworked anyways as we don't really care about short circuiting in this instance. That is more important when we don't want to send excessive messages to astarte

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -475,6 +483,243 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
+      expect(AvailableDeploymentsMock, :get, fn _, _ ->
+        {:ok,
+         [
+           %DeploymentStatus{
+             id: deployment.id,
+             status: :stopped
+           }
+         ]}
+      end)
```
noaccOS, line 550: see if you can reuse this https://github.com/edgehog-device-manager/edgehog/pull/731/files#diff-20bdec863086b49e111b95ffa702843a39d08f2fbc2f1812c8ae5a1ab09da4b5R177

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -475,6 +483,243 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
+
```
noaccOS, line 630: see if you can reuse this https://github.com/edgehog-device-manager/edgehog/pull/731/files#diff-20bdec863086b49e111b95ffa702843a39d08f2fbc2f1812c8ae5a1ab09da4b5R177

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -475,6 +483,243 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
```
noaccOS, line 709: see if you can reuse this https://github.com/edgehog-device-manager/edgehog/pull/731/files#diff-20bdec863086b49e111b95ffa702843a39d08f2fbc2f1812c8ae5a1ab09da4b5R177

backend/lib/edgehog/devices/devices.ex
```diff
@@ -108,8 +108,10 @@ defmodule Edgehog.Devices do
 
       define :update_application, action: :update_application, args: [:from, :to]
 
-      define_calculation :available_images
-      define_calculation :available_containers
+      define_calculation :available_images, args: [:_record]
```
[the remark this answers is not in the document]
lusergit: Specifies the resource to use to perform the calculation. Without it the signature of the interface becomes `Devices.available_images/1` with only the optional `opts` parameter. I think that in that case the calculation is performed on all the resources that respect the query (which could be either all devices or no device, I have not checked but either way is not the desired behavior)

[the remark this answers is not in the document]
lusergit: This has to be done on all the other `available_*` code interfaces too

backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
```diff
@@ -46,7 +48,7 @@ defmodule Edgehog.Containers.ManualActions.SendDeployRequest do
            :ok <- send_create_container_requests(device, containers),
            :ok <- send_create_network_requests(device, networks),
            {:ok, _device} <- Devices.send_create_deployment_request(device, deployment) do
-        {:ok, deployment}
+        Containers.deployment_set_status(deployment, :sent, nil, tenant: tenant)
```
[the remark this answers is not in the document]
lusergit: I thought that too but thats not the case as `:set_status` action accepts both arguments, hence they are marked as required by default, we can put `{:optional, :message}` in the code action, but since sometimes we need to set the message, and some times the status itself I thought was a better idea to keep the interface and tweak the arguments as needed

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
lusergit: right I got confused, almost hit myself

lusergit: leaving the `deployment.message` because of [prior comment](https://github.com/edgehog-device-manager/edgehog/pull/724#discussion_r1856395142)

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
lusergit: I initially introduced these as the linter was complaining of eccessive nesting, I can them and duplicate code if needed

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -475,6 +483,125 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
+      {:ok, deployment} = Ash.get(Deployment, deployment.id, tenant: tenant)
+      assert deployment.status == :ready
+    end
+
+    test "AvailableContainers triggers update deployment status", context do
+      %{conn: conn, realm: realm, device: device, tenant: tenant} = context
+
+      release = release_fixture(containers: 1, tenant: tenant)
+
+      assert {:ok, release} = Ash.load(release, containers: [:image, :networks])
```
noaccOS, line 531: You assert something when you want to validate your code is working as expected. You do not want to test external libraries, they have their own tests

```suggestion
      release = release_fixture(containers: 1, tenant: tenant) |> Ash.load!(containers: [:image, :networks])
```

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -475,6 +483,125 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
+      path = Routes.astarte_trigger_path(conn, :process_event, tenant.slug)
+
+      conn =
+        conn
+        |> put_req_header("astarte-realm", realm.name)
+        |> post(path, deployment_event)
+
+      assert response(conn, 200)
+
+      {:ok, deployment} = Ash.get(Deployment, deployment.id, tenant: tenant)
```
noaccOS, line 562: don't be afraid of bangs, especially in tests

```suggestion
      deployment = Ash.get!(Deployment, deployment.id, tenant: tenant)
```

backend/test/edgehog_web/controllers/astarte_trigger_controller_test.exs
```diff
@@ -475,6 +483,125 @@ defmodule EdgehogWeb.Controllers.AstarteTriggerControllerTest do
...
+      set_resource_expectations(deployment)
+
+      path = Routes.astarte_trigger_path(conn, :process_event, tenant.slug)
+
+      conn =
+        conn
+        |> put_req_header("astarte-realm", realm.name)
+        |> post(path, deployment_event)
+
+      assert response(conn, 200)
```
noaccOS, line 599: `response` already raises if the response does not have the expected exit code

```suggestion
      conn
      |> put_req_header("astarte-realm", realm.name)
      |> post(path, deployment_event)
      |> response(200)
```

noaccOS (changes requested):
we're getting closer!
this round is mostly tests and a couple questions

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -106,6 +111,75 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+            |> Enum.flat_map(&Containers.releases_with_container!(&1.id, tenant: tenant))
+            |> Enum.map(&Ash.load!(&1, :release).release.id)
+            |> Enum.uniq()
+
+          deployments =
+            releases
+            |> Enum.flat_map(&Containers.deployments_with_release!(&1, tenant: tenant))
+            |> Enum.uniq_by(& &1.id)
+
+          {:ok, Enum.map(deployments, &Containers.deployment_update_status/1)}
```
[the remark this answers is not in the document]
lusergit: I thought that this way we can also report partial success to the frontend, but maybe the return is a bit unclear this way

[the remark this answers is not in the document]
noaccOS: It's only internal database stuff, not "we received invalid data from the device"
Let it crash

```suggestion
          {:ok, Enum.map(deployments, &Containers.deployment_update_status!/1)}
```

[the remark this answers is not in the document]
lusergit: Maybe better: log errors when updating status, but returning `{:ok, deployments}` in any case
```suggestion
        Enum.map(deployments, fn deployment ->
          with {:error, error} <- Containers.deployment_update_status(deployment, tenant: tenant) do
            Logger.error("Error while updating status status of deployment #{deployment.id}: #{inspect(error)}")
          end
        end)

        {:ok, deployments}
```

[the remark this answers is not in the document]
lusergit: I actually think this is ok, the second argument of the  `{:ok, ...}` tuple as return type of the `handle_event` callback is discarded anyway, so for now it does not really matter, this is coincise while doing the right thing (calling the `update_status` on all interested deployments)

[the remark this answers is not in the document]
noaccOS: honestly, this is just the callback of a trigger. IMO it is safe to just crash because in normal circumstances, `deployment_update_status` will not fail

if you insist on logging the errors, you could do something like this

```elixir
          {deployments, errors} = 
            deployments
            |> Enum.map(&{&1.id, Containers.deployment_update_status(&1)})
            |> Enum.split_with(&match?({_, {:ok, _}}, &1))
            
          for {deployment_id, {:error, reason}} <- errors do
            Logger.error("Error while updating status status of deployment #{deployment_id}: #{inspect(reason)}")
          end
            
          deployments = deployments |> Enum.map(fn {_, {:ok, deployment}} -> deployment end)
          {:ok, deployments}
```

so that you at least have a separation between program logic and error handling
(yes the double tuple is kind of ugly, this is just an example)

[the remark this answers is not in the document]
noaccOS: > I actually think this is ok, the second argument of the {:ok, ...} tuple as return type of the handle_event callback is discarded anyway

Then I'd argue you make it clear we're only interested in the side effects

```elixir
for deployment <- deployments do
  Containers.deployment_update_status(deployment)
end

{:ok, nil} # or {:ok, deployments} or even :ok (this I don't recommend)
```

backend/lib/edgehog/containers/deployment/changes/check_images.ex
```diff
@@ -0,0 +1,55 @@
...
+           Ash.load(deployment, [:device, release: [containers: [:image]]], reuse_values?: true),
+         {:ok, pulled_images_statuses} <-
+           Devices.available_images(deployment.device, tenant: tenant) do
+      pulled_images_ids =
+        Enum.map(pulled_images_statuses, & &1.id)
+
+      missing_images =
+        deployment.release.containers
+        |> Enum.map(& &1.image.id)
+        |> Enum.reject(&(&1 in pulled_images_ids))
```
noaccOS, line 43:
```suggestion
         {:ok, available_images_statuses} <-
           Devices.available_images(deployment.device, tenant: tenant) do
      available_images_ids =
        Enum.map(available_images_statuses, & &1.id)

      missing_images =
        deployment.release.containers
        |> Enum.map(& &1.image.id)
        |> Enum.reject(&(&1 in available_images_ids))
```

noaccOS (changes requested):
tiny leftovers

noaccOS (approved):
good job

PR #726, opened 2024-11-18
test(containers): fix deployment start tests

noaccOS:


backend/lib/edgehog/devices/device/device.ex
```diff
@@ -315,15 +315,11 @@ defmodule Edgehog.Devices.Device do
...
-        constraints instance_of: Release
+        constraints instance_of: Deployment
         allow_nil? false
         description "The new release of the application"
       end
 
-      validate {Edgehog.Containers.Validations.SameApplication, [release_a: :from, release_b: :to]}
-
-      validate {Edgehog.Containers.Validations.IsUpgrade, [from: :from, to: :to]}
-
```
lusergit, line 326: I think the validation logics should be kept here. If done in this action they are first class citizens. We can keep the logic implemented in `send_deployment_upgrade` but delegate the validations to the update action on the device

lusergit (changes requested):
Overall I think the code is very much ready, the only thing I think we would benefit in the mid/long term is to move the validations as they were to keep them first class citizens.

backend/lib/edgehog/devices/device/device.ex
noaccOS: I like validations where they are now, for a simple reason: ideally, you want to validate data only when receiving data from outside. This way you fail fast and in the rest of the program you can assume your data is valid.
While I agree it would be better to have first class citizen validations, I would argue it's mainly Ash's fault because it does not provide the tools to interact with its own ecosystem to a generic action.

If you really don't like validations as they are now, I propose changing the action to an update:
- We gain the ability to run normal validations on the data, granted we figure out how to pass them the required values (one parameter is the changeset, the other loaded from an argument).
- Our inputs/outputs get messed up. a lot. our api signature becomes `updgradeDeployment(id: ID!, input: %{to: ID})` and return type... one of the two deployments.

noaccOS:
Repurposed this pr to only  include the fix for the existing test. Will open a draft pr for the deployment upgrade api

lusergit (approved):

PR #727, opened 2024-11-18
fix(containers): `createRelease` relay translation

lusergit:
Fix the `application_id` being parsed as a raw id and not a relay id.

### DIFF
--- a/backend/test/edgehog_web/schema/mutation/create_release_test.exs
+++ b/backend/test/edgehog_web/schema/mutation/create_release_test.exs
@@ -1,0 +1,91 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


backend/test/edgehog_web/schema/mutation/create_release_test.exs
```diff
@@ -0,0 +1,90 @@
...
+    """
+
+    {tenant, opts} = Keyword.pop!(opts, :tenant)
+    {application_id, opts} = Keyword.pop!(opts, :application_id)
+
+    default_input = %{
+      "version" => Keyword.get(opts, :version, unique_release_version()),
+      "application_id" => application_id,
+      "containers" => []
+    }
```
noaccOS, line 65: Just adding a single comment because the rest looks good

usually we use pop! when we want to then use `Enum.into`, to make sure we remove our options but leave the user able to add extra parameters (which you couldn't do if you used `Keyword.take` to only keep your default values)

if you don't need to re-use opts this way, it's easier to understand if you just use `fetch!` instead.

```suggestion
    tenant = opts |> Keyword.fetch!(:tenant)

    default_input = %{
      "version" => Keyword.get(opts, :version, unique_release_version()),
      "application_id" => opts |> Keyword.fetch!(:application_id),
      "containers" => []
    }
```

Extra: TBF, in this specific example, you may consider using `take |> into`

```elixir
    tenant = opts |> Keyword.fetch!(:tenant)

    default_input = %{
      "version" => unique_release_version(),
      "containers" => []
    }

    input = 
      opts 
      |> Keyword.take([:application_id, :version, :containers]) 
      |> Enum.map(fn {key, value} -> {Atom.to_string(key), value} end) 
      |> Enum.into(default_input)
```

noaccOS (approved):

PR #728, opened 2024-11-19
feat(containers): deployment upgrade graphql api

noaccOS:
Previously #726 
After irl discussion, will be changed to an update which will accept the existing deployment and the new release, and then both deploy and upgrade the release.

Will need to figure out how to handle both deployments existing on the device, probably our best option would be to send a delete after the new version has been started.


backend/lib/edgehog/containers/manual_actions/send_deployment_upgrade.ex
```diff
@@ -0,0 +1,52 @@
...
+               deployment: %{device_id: device_id, release_id: target_id},
+               action_type: :upgrade_deployment,
+               action_arguments: %{upgrade_target_id: Ash.Changeset.get_data(changeset, :id)}
+             }
+           )
+           |> Ash.create(tenant: tenant),
+         result = action.deployment,
+         :ok <- Containers.send_deploy_request(result) do
+      {:ok, result}
+    end
```
lusergit, line 50: I *think* that what we also want to do here is to automagically send to the device the upgrade request, meaning

```suggestion

    current_id = Ash.Changeset.get_data(changeset, :id)

    with {:ok, action} <-
           DeploymentReadyAction
           |> Ash.Changeset.for_create(
             :create_deployment,
             %{
               deployment: %{device_id: device_id, release_id: target_id},
               action_type: :upgrade_deployment,
               action_arguments: %{upgrade_target_id: current_id}
             }
           )
           |> Ash.create(tenant: tenant),
         result = action.deployment,
         :ok <- Containers.send_deploy_request(result),
         :ok <- Devices.update_application(current_id, result.id) do
      {:ok, result}
    end
```

Or pherhaps we should wait for the new deployment to be ready to mark it as an update?

backend/lib/edgehog/containers/deployment_ready_action.ex
```diff
@@ -0,0 +1,67 @@
...
+    defaults [:read, :destroy]
+
+    create :create_deployment do
+      accept [:action_type]
+
+      argument :action_arguments, :map
+
+      argument :deployment, :map
+
+      change manage_relationship(:deployment, type: :create)
```
lusergit, line 38: this will call the primary `:create` action on the deployment, i.e., it will not automatically deploy the release on the device, to do so you have to manually call the deploy create action

```suggestion
      change manage_relationship(:deployment, on_no_match: {:create, :deploy}, on_match: :ignore)
```

lusergit (changes requested):
Overall LGTM, I think we also want to send the `update_application` request to the device once the new deployment is created but overall is ok

backend/lib/edgehog/containers/manual_actions/send_deployment_upgrade.ex
noaccOS: Yes, the whole point of the `DeploymentReadyAction`s is to have callbacks for when the deployment reaches the `:ready` state. we should not send a start/update request when the deployment is not ready yet

backend/lib/edgehog/containers/deployment_ready_action.ex
noaccOS: I was doing that manually [here](https://github.com/edgehog-device-manager/edgehog/pull/728/files#diff-040dec3573e5fdd58268cc6a8a20f9c53d3a78f00e238e72edbd7fa1d2e765efR48), but it may be a better idea to call the correct action on the deployment, right

PR #729, opened 2024-11-20
refactor(containers): send deploymentId on resource requests

noaccOS:
this is to be used by the device for error handling

waiting on updated interfaces


noaccOS:
marking as draft. we should wait until the device is ready

PR #731, opened 2024-11-20
feat(containers): run ready actions when the deployment is ready

noaccOS:
drafted because it depends on #724 and 724 is not ready, but 279515297e5f34665372476ac314bec7517906d4 is ready for review

### DIFF
--- a/backend/lib/edgehog/containers/manual_actions/run_ready_actions.ex
+++ b/backend/lib/edgehog/containers/manual_actions/run_ready_actions.ex
@@ -1,0 +1,45 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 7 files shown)


lusergit (approved):
lgtm

backend/lib/edgehog/containers/manual_actions/run_ready_actions.ex
```diff
@@ -0,0 +1,43 @@
...
+  def update(changeset, _opts, _context) do
+    deployment = changeset.data
+
+    with {:ok, deployment} <- Ash.load(deployment, :ready_actions) do
+      Enum.reduce_while(deployment.ready_actions, {:ok, deployment}, fn action, _acc ->
+        case Containers.run_ready_action(action) do
+          {:ok, _ready_action} -> {:cont, {:ok, deployment}}
+          {:error, reason} -> {:halt, {:error, reason}}
+        end
+      end)
```
lusergit, line 40: Maybe here we don't actually want to halt if one ready action fails, I think it would be better to just run the ready action and let them crash if something happens
```suggestion
      _ready_actions = Enum.map(deployment.ready_actions, &Containers.run_ready_action/1))
      {:ok, deployment}
```

lusergit (commented):
Overall approval, just a small comment on `run_ready_actions.ex`

PR #742, opened 2024-11-26
fix!(containers): remove ready status

lusergit:
- `:ready` and `:stopped` are redundant statuses, moved to using only `:stopped` status;
- the `check_deployment` check when the deployment is available sets the deployment status tio the device deployment status

### DIFF
--- a/backend/lib/edgehog/containers/deployment/changes/check_deployments.ex
+++ b/backend/lib/edgehog/containers/deployment/changes/check_deployments.ex
@@ -32,14 +32,14 @@ defmodule Edgehog.Containers.Deployment.Changes.CheckDeployments do
 
     with {:ok, :created_containers} <- Ash.Changeset.fetch_argument_or_change(changeset, :status),
          {:ok, deployment} <- Ash.load(deployment, :device),
-         {:ok, available_deployments} <-
+         {:ok, available_deployments_statuses} <-
            Devices.available_deployments(deployment.device, tenant: tenant) do
-      available_deployments =
-        Enum.map(available_deployments, & &1.id)
+      deployment_status =
+        Enum.find(available_deployments_statuses, &(&1.id == deployment.id))
 
-      if deployment.id in available_deployments do
+      if deployment_status do
         changeset
-        |> Ash.Changeset.change_attribute(:status, :ready)
+        |> Ash.Changeset.change_attribute(:status, deployment_status.status)
         |> Ash.Changeset.after_transaction(fn _changeset, transaction_result ->
           with {:ok, deployment} <- transaction_result do
             Containers.run_ready_actions(deployment)
--- a/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
+++ b/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
@@ -41,6 +41,15 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
   @ota_response "io.edgehog.devicemanager.OTAResponse"
   @system_info "io.edgehog.devicemanager.SystemInfo"
 
+  @initial_statuses [
+    :created,
+    :sent,
+    :created_images,
+    :created_networks,
+    :created_containers,
+    :created_deployment
+  ]
+
   @impl Ash.Resource.Actions.Implementation
   def run(input, _opts, _context) do
     realm_name = input.arguments.realm_name
@@ -198,8 +207,14 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
           # Skip Stopping if already Stopped
           {:ok, deployment}
 
-        _ ->
+        {_, "Error"} ->
+          # Errors have precedence
           Containers.deployment_set_status(deployment, status, message, tenant: tenant)
+
+        _ ->
+          if deployment.status in @initial_statuses,
+            do: Containers.deployment_update_status(deployment, tenant: tenant),
+            else: Containers.deployment_set_status(deployment, status, message, tenant: tenant)
       end
     end
   end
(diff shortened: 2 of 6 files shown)


backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -213,7 +213,7 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
           if status == nil do
             Containers.delete_deployment(deployment)
           else
-            Containers.deployment_set_status(deployment, status, deployment.message, tenant: tenant)
+            Containers.deployment_update_status(deployment, tenant: tenant)
```
noaccOS, line 216: I don't think this is correct: after the deployment hits what now is the `ready` state, no other change of state can happen from `deployment_update_status`, as it's basically a state machine which only operates if the deployment is in the state if thought it was in
I'd leave that as is, because its job is initialization, and instead port the required changes here

```elixir
cond do
  status == nil -> Containers.delete_deployment(deployment)
  deployment.status in @initial_states -> Containers.deployment_update_status(deployment, tenant: tenant)
  true -> Containers.deployment_set_status(deployment, status, deployment.message, tenant: tenant)
end
```

noaccOS (changes requested):
We're also missing the change on `deployment_event`: while we want to propagate a deployment `error` in any situation, we want to avoid setting `starting/stopping` while we're setting up the deployment, otherwise it will mess with the "state machine"

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -41,6 +41,14 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
   @ota_response "io.edgehog.devicemanager.OTAResponse"
   @system_info "io.edgehog.devicemanager.SystemInfo"
 
+  @initial_states [
```
noaccOS, line 44: probably not needed but to be extra safe:
```suggestion
  @initial_states [
    :created,
```

davidebriani (approved):
Just some nitpicks, but the PR is good

backend/lib/edgehog/containers/deployment/changes/check_deployments.ex
```diff
@@ -34,12 +34,12 @@ defmodule Edgehog.Containers.Deployment.Changes.CheckDeployments do
          {:ok, deployment} <- Ash.load(deployment, :device),
          {:ok, available_deployments} <-
            Devices.available_deployments(deployment.device, tenant: tenant) do
-      available_deployments =
-        Enum.map(available_deployments, & &1.id)
+      deployment_status =
+        Enum.find(available_deployments, &(&1.id == deployment.id))
```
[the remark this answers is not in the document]
davidebriani: Just noticed we already have a `deployment` in scope. Maybe `available_deployment` then?

davidebriani (approved):

backend/lib/edgehog/containers/deployment/changes/check_deployments.ex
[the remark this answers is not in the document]
lusergit: The fact is that that is not really a deployment but the status of a deployment, I changed the variable `available_deployments` to  `available_deployments_statuses` which maybe is clearer :+1:

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -198,8 +207,14 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
 
-        _ ->
+        {_, "Error"} ->
+          # Errors have precedence
           Containers.deployment_set_status(deployment, status, message, tenant: tenant)
+
+        _ ->
+          if deployment.status in @initial_statuses,
+            do: {:ok, deployment},
+            else: Containers.deployment_set_status(deployment, status, message, tenant: tenant)
```
davidebriani, line 217: If the device is publishing the Stopped status when all resources are ready, we should update the deployment status to `Stopped` instead of doing nothing.

You consider using a `cond` statement here too:

```elixir
with {:ok, deployment} <- Containers.fetch_deployment(deployment_id, tenant: tenant) do
  cond do
    status == "Starting" and deployment.status == :started ->
      # Skip Starting if already Started
      {:ok, deployment}

    status == "Stopping" and deployment.status == :stopped ->
      # Skip Stopping if already Stopped
      {:ok, deployment}
      
    status == "Error" ->
      # An error has occurred, set the deployment status immediately
      Containers.deployment_set_status(deployment, status, message, tenant: tenant)
      
    deployment.status in @initial_statuses ->
      # Besides errors, verify status of deployment resources before
      # blindly setting the deployment status. Indeed, a Stopped
      # status could be signaled before all resources are ready.
      Containers.deployment_update_status(deployment, tenant: tenant)
      
    true ->
      # Besides initial statuses, directly set the deployment status
      Containers.deployment_set_status(deployment, status, message, tenant: tenant)
  end
end
```

davidebriani (changes requested):

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
noaccOS: On deployment event we can only receive starting or stopping, not stopped
If all resources are ready and we receive a stopped on `available_deployments`, the `update_status` will correctly set the status to `stopped` and run the callbacks

davidebriani: Yes, I just realized the same thing.

We could still consider the `cond` statement if it makes the code clearer and less nested.

Regarding the `Stopped` status, it probably doesn't do any harm calling `Containers.deployment_update_status(deployment, tenant: tenant)` here too. It might serves us well if we get the device to also publish `Stopped` on DeploymentEvents as we were discussing the other day (because that would fix the fact that the Stopped status property on AvailableDeployments is not re-sent from Astarte and would let us avoid having the first two `cond` cases).

Just food for thought, the current code already works.

noaccOS (approved):

PR #743, opened 2024-11-26
Add is online check to deployments 

OmarBrbutovic:
- Added `isOnline` state to track the device's online status
- Passed `isOnline` as a prop to `AddAvailableApplications` and `DeployedApplicationsTable`
- Displayed an error message when the device is offline, preventing new deployments and restricting starting/stopping existing ones

<p>

image
image
image

</p>
</details>

### DIFF
--- a/frontend/src/components/DeployedApplicationsTable.tsx
+++ b/frontend/src/components/DeployedApplicationsTable.tsx
@@ -249,7 +249,7 @@ const DeployedApplicationsTable = ({
           onError: () => {
             setErrorFeedback(
               <FormattedMessage
-                id="components.AddAvailableApplications.startErrorFeedback"
+                id="components.DeployedApplicationsTable.startErrorFeedback"
                 defaultMessage="Could not Start the Deployed Application, please try again."
               />,
             );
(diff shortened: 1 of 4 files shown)


frontend/src/components/DeployedApplicationsTable.tsx
```diff
@@ -229,60 +231,78 @@ const DeployedApplicationsTable = ({
...
+          },
+        });
+      } else {
+        setErrorFeedback(
+          <FormattedMessage
+            id="components.DeployedApplicationsTable.startErrorOffline"
+            defaultMessage="The device is disconnected. You cannot start an application while it is offline."
+          />,
+        );
+      }
```
davidebriani: Pro tip for the future: consider early returns in case of error.
This way:
- the conditional has a very short code span, and the rest of the code continues the happy path without having to worry about the condition anymore.
- it isn't necessary to nest the code of the happy path, so the code diff is minimal and the review is easier.

```typescript
if (!isOnline) {
  return setErrorFeedback(
    <FormattedMessage
      id="components.DeployedApplicationsTable.startErrorOffline"
      defaultMessage="The device is disconnected. You cannot start an application while it is offline."
    />
  );
}

// happy path
startDeployment({ ... });
```

davidebriani (approved):

PR #744, opened 2024-11-26
UI: Handle new deployment statuses

davidebriani:
The code is adapted to handle the new schema and TypeScript types, handling all new initial deployment statuses as the synthetic "Deploying" status.


noaccOS (approved):
The change looks good but I'd wait for #742 before merging this one as that one will modify the generated graphql schema

noaccOS (approved):

PR #746, opened 2024-11-26
Fix calculations of Astarte values for container resources

davidebriani:
The calculations defined in the Devices domain do not correctly load the calculation's dependencies, so the calculations cannot be computed. The proposed change is to just load the calculations with Ash.load, which was already done in the addressed code anyway.

### DIFF
--- a/backend/lib/edgehog/containers/deployment/changes/check_containers.ex
+++ b/backend/lib/edgehog/containers/deployment/changes/check_containers.ex
@@ -22,22 +22,17 @@ defmodule Edgehog.Containers.Deployment.Changes.CheckContainers do
   @moduledoc false
   use Ash.Resource.Change
 
-  alias Edgehog.Devices
-
   @impl Ash.Resource.Change
-  def change(changeset, _opts, context) do
-    %{tenant: tenant} = context
+  def change(changeset, _opts, _context) do
     deployment = changeset.data
 
     with {:ok, :created_networks} <- Ash.Changeset.fetch_argument_or_change(changeset, :status),
          {:ok, deployment} <-
-           Ash.load(deployment, [:device, release: [:containers]], reuse_values?: true),
-         {:ok, available_containers} <-
-           Devices.available_containers(deployment.device, tenant: tenant) do
-      available_containers = Enum.map(available_containers, & &1.id)
+           Ash.load(deployment, device: :available_containers, release: [:containers]) do
+      available_container_ids = Enum.map(deployment.device.available_containers, & &1.id)
 
       missing_containers =
-        Enum.reject(deployment.release.containers, &(&1.id in available_containers))
+        Enum.reject(deployment.release.containers, &(&1.id in available_container_ids))
 
       if missing_containers == [] do
         Ash.Changeset.change_attribute(changeset, :status, :created_containers)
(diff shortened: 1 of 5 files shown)


backend/lib/edgehog/containers/deployment/changes/check_containers.ex
```diff
@@ -22,22 +22,17 @@ defmodule Edgehog.Containers.Deployment.Changes.CheckContainers do
...
+  def change(changeset, _opts, _context) do
     deployment = changeset.data
 
     with {:ok, :created_networks} <- Ash.Changeset.fetch_argument_or_change(changeset, :status),
          {:ok, deployment} <-
-           Ash.load(deployment, [:device, release: [:containers]], reuse_values?: true),
-         {:ok, available_containers} <-
-           Devices.available_containers(deployment.device, tenant: tenant) do
-      available_containers = Enum.map(available_containers, & &1.id)
+           Ash.load(deployment, [device: :available_containers, release: [:containers]], reuse_values?: true) do
```
noaccOS, line 31: this will not work as expected: if we use the same `load` call with `reuse_values?: true`, we won't load the updated resources from the device

```suggestion
           Ash.load(deployment, [device: :available_containers, release: [:containers]]) do
```

backend/lib/edgehog/containers/deployment/changes/check_networks.ex
```diff
@@ -22,27 +22,22 @@ defmodule Edgehog.Containers.Deployment.Changes.CheckNetworks do
...
     deployment = changeset.data
 
     with {:ok, :created_images} <- Ash.Changeset.fetch_argument_or_change(changeset, :status),
          {:ok, deployment} <-
-           Ash.load(deployment, [:device, release: [containers: [:networks]]], reuse_values?: true),
-         {:ok, available_networks} <-
-           Devices.available_networks(deployment.device, tenant: tenant) do
-      available_networks =
-        Enum.map(available_networks, & &1.id)
+           Ash.load(deployment, [device: :available_networks, release: [containers: [:networks]]], reuse_values?: true) do
```
noaccOS, line 31: same as above

noaccOS (changes requested):
by fixing the two comments I'm not against merging this, although honestly I don't understand why it wouldn't work as it is now

backend/lib/edgehog/containers/deployment/changes/check_images.ex
```diff
@@ -22,20 +22,15 @@ defmodule Edgehog.Containers.Deployment.Changes.CheckImages do
...
-    %{tenant: tenant} = context
+  def change(changeset, _opts, _context) do
     deployment = changeset.data
 
     with :sent <- deployment.status,
          {:ok, deployment} <-
-           Ash.load(deployment, [:device, release: [containers: [:image]]], reuse_values?: true),
-         {:ok, available_images_statuses} <-
-           Devices.available_images(deployment.device, tenant: tenant) do
+           Ash.load(deployment, [device: :available_images, release: [containers: [:image]]], reuse_values?: true) do
```
noaccOS, line 31: sorry there's also this one

noaccOS (approved):

PR #747, opened 2024-11-27
chore(containers): better create release error

lusergit:
- creating a new release with a version already in use complains about the `:version` and not about the `:application_id`
- added tests to check the error messages and related fields


backend/lib/edgehog/containers/release.ex
```diff
@@ -88,7 +88,7 @@ defmodule Edgehog.Containers.Release do
   end
 
   identities do
-    identity :application_version, [:application_id, :version]
+    identity :application_version, [:version, :application_id]
```
davidebriani, line 91: Does this require regenerating database migrations and snapshots?

lusergit: Added migrations to the last commit version

davidebriani (approved):

PR #754, opened 2024-12-04
chore(containers): refactor deployment

lusergit:
During the development of the `containers` feature we were prioritizing speed over project structure. This PR aims to refactor the existing code to be more comprehensible and understandable. The main changes are the new `*.Deployment` resources and the `containers` resources placement

## Containers structure
Each resources represented in the db has its own folder in the `backend/lib/edgehog/containers` folder, containing the changes, the manual actions and all the things needed for that particular resource. Each resource now has a `Deployment` module, representing the many to many relation between the resource and the devices.

## `Deployment`s
- Each resource has a `Deployment` module, which is the `many_to_many` relation between the resource and the device. 
- Each deployment has a `:state` attribute, representing the current state of the resource on the device. The state is updated trough the trigger handler and does not enforce a particular transition system between the states. This is done because we might lose some messages in the communication with the device and this way the backend can cope with the situation rather than throwing errors and not handling the situation correctly.
- The deployment is responsible, upon its creation, to send to astarte the appropriate messages to actually build the resources. The first two states of each deployment are the same for all resources: 
  - `:created` means the resources has been created in the database
  - `:sent` means that the backend has correctly sent to astarte the informations for the device to create the resource, will then be astarte responsible to actually send the informations to the device itself.


lusergit:
Tests will not pass for now as the infrastructure is not complete yet

PR #755, opened 2024-12-06
feat(ota): exposing optional update target of OTA

### LINKED ISSUE #356: Expose the associated UpdateCampaign (if any) from an OTA Operation
Right now we have a boolean flag `manual?` which indicates if an OTA Operation was started manually or from an Update Campaign. We should expose the Update Campaign too, so that we can link back to it in the frontend. At that point, the flag would become redundant too since it could just be desumed by the presence of a linked Update Campaign

lusergit:
Ota operations expose an optional `UpdateTarget`. It is a reference to the update target that created the managed ota operation

Closes #356

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10] - Unreleased
+### Added
+- Managed OTA operations expose the update target that created them in graphql ([#356](https://github.com/edgehog-device-manager/edgehog/issues/356).
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.10] - Unreleased
+### Added
+- Managed OTA operations expose the update target that created them in graphql.
```
davidebriani, line 9: A reference to the originating issue would be handy, if any

```suggestion
- Managed OTA operations expose the update target that created them in graphql ([#356](https://github.com/edgehog-device-manager/edgehog/issues/356).
```

davidebriani (changes requested):

davidebriani (approved):

PR #759, opened 2024-12-06
Allow to reach your edgehog instance from outside your host

### LINKED ISSUE #758: Support for external devices when using docker compose
When following Edgehog in 5 minutes, Edgehog is set up using Docker Compose.
With this method, connecting devices simulated on the same host machine running the Edgehog instance is possible.
However, some Edgehog devices are intended to run on external hardware.
This should be supported by the Docker compose setup to facilitate first-time users in trying out Edgehog.

matt-mazzucato:
Your edgehog (created via docker compose) can now be exposed on an arbitrary domain, making it possible for devices in the same subnet to interact with it.

Fix #758

### DIFF
--- a/docker-compose.without-astarte.yml
+++ b/docker-compose.without-astarte.yml
@@ -18,7 +18,6 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-version: "3.8"
 services:
   traefik:
     image: traefik:v2.10
--- a/.env
+++ b/.env
@@ -22,6 +22,14 @@ IPBASE_API_KEY=
 GOOGLE_GEOLOCATION_API_KEY=
 GOOGLE_GEOCODING_API_KEY=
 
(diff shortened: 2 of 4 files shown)


docker-compose.without-astarte.yml
```diff
@@ -18,7 +18,6 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-version: "3.8"
```
Pavinati, line 21: I don't think removing this version is related to the PR's description.
Is this required or a leftover from testing?

Pavinati: I agree, let's just put it in its own commit so we can describe why it was removed

matt-mazzucato: Done!

davidebriani (approved):

PR #764, opened 2024-12-09
fix(bucket_storage): image deletion

noaccOS:
previously, calling Ash.destroy! on a base image would crash the program because Edgehog.BaseImages.Uploaders.BaseImage.filename/2 would try to access the field `base_image_version` of a BaseImage, which does not exist


davidebriani:
PR is fine. Could we add a line of change to CHANGELOG.md?
It would be a good addition for a new `[0.9.3] - Unreleased` section.

lusergit (approved):

davidebriani (approved):

PR #766, opened 2024-12-09
docs(edgehog_in_5): fix renconciliation command

noaccOS:
the command was not updated after the transition to ash.

updates the doc to use the new functions to list and reconcile the tenants


lusergit (approved):

noaccOS:
superseeded by #767

PR #767, opened 2024-12-09
chore(seeds): reconcile tenant when seeding

lusergit:
- Reconciling tenant is done directly in seeds when it gets created
- Edgehog in 5 minutes does no longer instruct users to manually reconcile the tenant with astarte

### DIFF
--- a/backend/priv/repo/seeds.exs
+++ b/backend/priv/repo/seeds.exs
@@ -128,6 +128,11 @@ tenant =
 {status, realm_pk} =
   read_key!.("SEEDS_REALM_PRIVATE_KEY_FILE", "SEEDS_REALM_ORIGINAL_FILE", "realm_private")
 
+Astarte.create_realm!(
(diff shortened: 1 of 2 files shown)


backend/priv/repo/seeds.exs
```diff
@@ -142,4 +142,6 @@ Astarte.create_realm!(
   tenant: tenant
 )
 
+Edgehog.Tenants.reconcile_tenant(tenant)
```
noaccOS, line 145: As we already have all the data we need, I suggest adding a simple check so that we avoid reconciling the tenant when the user is using the default key (and thus is not associated with a real realm).
This is just to avoid the backend logging a bunch of errors. At least until the next reconciliation attempt.

```elixir
if status == :default do
  """
  You are using the default realm private key. \
  This instance will not be able to connect to Astarte.
  """
  |> String.trim_trailing("\n")
  |> Logger.warning()
else
  Edgehog.Tenants.reconcile_tenant(tenant)
end
```

PR #769, opened 2024-12-10
chore: update ash_graphql

noaccOS:
ash_graphql allows us to use the node query in the frontend, thanks to https://github.com/ash-project/ash_graphql/pull/246

However, there is not yet a tag with the feature, so this switches to using main at the moment


lusergit (approved):

davidebriani:
Closing the PR as not needed anymore. Ash GraphQL is already at a recent version that includes the features we need: there's no reason to update the dependency to a non-release commit.

PR #770, opened 2024-12-10
Fix doc page devices

sorru94:
Merge only after the related ESP32 API documentation [PR](https://github.com/edgehog-device-manager/edgehog-esp32-device/pull/108).


doc/mix.exs
```diff
@@ -66,7 +66,7 @@ defmodule Doc.MixProject do
       "pages/user/hardware_types.md",
       "pages/user/system_models.md",
       "pages/user/devices.md",
-      "pages/user/device_sdks_runtime.md",
+      "pages/user/devices_and_runtime.md",
```
harlem88, line 69: nitpick: what do you think about keep `sdks`?
```suggestion
      "pages/user/device_sdks_and_runtime.md",
```

sorru94: I removed the SDK suffix mostly because both the [ESP32](https://github.com/edgehog-device-manager/edgehog-esp32-device) and [Zephyr](https://github.com/edgehog-device-manager/edgehog-zephyr-device) ones are just called device.
If you think it's better to keep it I'm fine with adding it back.

harlem88 (approved):

davidebriani (approved):

PR #771, opened 2024-12-10
chore(update_channel): better errors

lusergit:
When creating and updating an update channel, target group ids are validated to be available, meaning:
- they are present in the DB
- they are not related to any other update channel

closes #529

PR #776, opened 2024-12-13
feat(update_channel) allow update channels to have 0 update groups

### LINKED ISSUE #339: [docs] Inconsistencies between  `Creating an Update Channel-Target Groups ` doc and actual behaviour
Currently in the  `Creating an Update Channel` section of `managed-ota-updates` page the `Target Groups` is declared as an optional field, but in the Fronted page (/update-channels/new) of Edgehog `Create Update Channel` the field is declared as `Required`
Screenshot_20230726_125831

### LINKED ISSUE #518: Allow Update Channels to have 0 target Device Groups
When creating or updating Update Channels, at least one target Device Group must be specified currently.

However, it may happen that a user wants to leave an update channel without target groups.

For example, consider a case where the user would like to "unlink" a target group from an update channel and link it to another update channel. If the first update channel has only 1 target group, the user will not be able to accomplish the operation.

lusergit:
Depends on #771

If merged  closes #339 and closes #518

### DIFF
--- a/backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
+++ b/backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
@@ -58,9 +58,6 @@ defmodule Edgehog.UpdateCampaigns.UpdateChannel do
         description """
         The IDs of the target groups that are targeted by this update channel.
         """
-
-        allow_nil? false
-        constraints min_length: 1
       end
 
       change Changes.RelateTargetGroups do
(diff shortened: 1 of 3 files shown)


backend/lib/edgehog/update_campaigns/update_channel/update_channel.ex
```diff
@@ -78,16 +75,13 @@ defmodule Edgehog.UpdateCampaigns.UpdateChannel do
...
       end
 
       # Needed because manage_relationship is not atomic
       require_atomic? false
 
-      change Changes.UnrelateCurrentTargetGroups do
-        where present(:target_group_ids)
-      end
+      # Always unrelate target groups
+      change Changes.UnrelateCurrentTargetGroups
```
davidebriani, line 84: Why always unrelating the target groups?
This action could be called just to update the name of the Update Channel, and you don't want to unrelate it from the associated target groups.

davidebriani (changes requested):

davidebriani (approved):

PR #777, opened 2024-12-13
chore(database): Add ssl support

### LINKED ISSUE #419: Expose Ecto SSL configuration
We should be able to connect to a Postgres database exposed over SSL, so we have to add some more environment variables.
The most basic one would just be `DATABASE_ENABLE_SSL`, then we could add the other ones when there's the need.

lusergit:
Adds ssl/tls support for the edgehog <-> database communication. 
Closes #419.

### DIFF
--- a/backend/config/runtime.exs
+++ b/backend/config/runtime.exs
@@ -179,10 +179,13 @@ end
 # any compile-time configuration in here, as it won't be applied.
 # The block below contains prod specific runtime configuration.
 if config_env() == :prod do
-  database_username = System.fetch_env!("DATABASE_USERNAME")
-  database_password = System.fetch_env!("DATABASE_PASSWORD")
-  database_hostname = System.fetch_env!("DATABASE_HOSTNAME")
-  database_name = System.fetch_env!("DATABASE_NAME")
+  database = %{
+    username: System.fetch_env!("DATABASE_USERNAME"),
+    password: System.fetch_env!("DATABASE_PASSWORD"),
+    hostname: System.fetch_env!("DATABASE_HOSTNAME"),
+    name: System.fetch_env!("DATABASE_NAME"),
+    ssl: Config.database_ssl_config()
+  }
 
   # The secret key base is used to sign/encrypt cookies and other secrets.
   # A default value is used in config/dev.exs and config/test.exs but you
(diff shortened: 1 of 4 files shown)


backend/config/runtime.exs
```diff
@@ -31,10 +31,86 @@ end
...
+
+          "false" ->
+            false
+
+          other ->
+            raise """
+            invalid database SSL configuration:
+            unknown value `#{other}` for variable DATABASE_USE_OS_CERTS.
+            """
+        end
```
noaccOS, line 59: short tip: [get_env](https://hexdocs.pm/elixir/1.17.3/System.html#get_env/2) accepts the default as a second parameter
long tip: you're manually doing quite a few type castings with validation. consider looking into [skogsra](https://github.com/gmtprime/skogsra). [`Edgehog.Config`](https://github.com/edgehog-device-manager/edgehog/blob/main/backend/lib/edgehog/config.ex) is already using it to define some environment variables

lusergit: The last commit uses skogsra and is consistent with rest of the configuration :+1:

backend/lib/edgehog/config.ex
```diff
@@ -94,6 +118,71 @@ defmodule Edgehog.Config do
...
+  Returns the ssl configuration for the ssl connection with the database.
+  """
+  @spec database_ssl_cert_config() :: {:cacerts, [term()]} | {:cacertfile, binary()}
+  def database_ssl_cert_config do
+    use_os_certs = database_use_os_certs?()
+
+    certfile = System.get_env("DATABASE_SSL_CACERTFILE")
+
+    case {certfile, use_os_certs} do
+      {nil, false} ->
```
noaccOS, line 149: I don't think we should raise regardless in this condition.
While it is a good idea to handle the situation, given that we can inform the user about the correct environment variables to set, the user may also just want to use `:verify_none`, in which case it does not need to set them

lusergit: Maybe I should have made this a private function, but this gets called only from line 172, so only if the user explicitly set `DATABASE_SSL_VERIFY=true`, in that case the configuration requires either a valid certificate or to use the os certificates

davidebriani (approved):

PR #782, opened 2025-01-08
chore(ash): update ash

lusergit:
- updating `ash`: 3.4.44 -> 3.4.52
- updating `ash_postgres` 2.4.15 -> 2.4.21
- updating `ash_graphql` 1.4.4 -> 1.4.7
- defers the load of the `realm_management_client` when reconciling all tenants. Previously we could ignore the tenant when loading because of a bug (see [ash changelog](https://github.com/ash-project/ash/blob/main/CHANGELOG.md#v3445-2024-12-10))


backend/lib/edgehog/tenants/reconciler.ex
```diff
@@ -73,7 +73,8 @@ defmodule Edgehog.Tenants.Reconciler do
       |> Ash.Query.load(:realm_management_client)
 
     Tenant
-    |> Ash.read!(load: [realm: global_realm_query])
+    |> Ash.read!()
+    |> Enum.map(&Ash.load!(&1, [realm: global_realm_query], tenant: &1))
```
davidebriani, line 77: Can't we just

```elixir
Tenant
|> Ash.read!(load: [realm: :realm_management_client])
|> Enum.each(&start_reconciliation_task(&1, tenant_to_trigger_url_fun))
```

lusergit: No because ash needs a tenant to load the realm of a tenant

davidebriani (approved):

Issue #779, opened 2024-12-20
Use server-side pagination for GraphQL queries that return lists


davidebriani:
All queries that return a possibly unbounded list of resources would be better served by API that support pagination.
This way API clients are free to request only a portion of the list at a time, without having to handle huge responses or put pressure on the server.

For instance, an initial attempt to convert the existing GraphQL queries to be paginated was started in [#542](https://github.com/edgehog-device-manager/edgehog/pull/542) which can be used as a reference for the backend work.

Ideally, the work can be split into several steps / PRs, so that both the backend and the frontend are updated at the same time: otherwise the API change would leave the frontend broken.

lusergit:
This is almost done, only one query is left, that query though relies on a manual relationship, not easily paginable. I think we can close this issue and open a new one ad-hoc for that specific query 

cc @davidebriani

davidebriani:
I agree, here's the remaining issue: #1117

PR #783, opened 2025-01-09
feat!(pagination): paginate device relationships

lusergit:
Part of #779
Paginates device's `:ota_operations` and `:tags` relationships with relay


davidebriani:
Since the PR introduces breaking changes that would leave the frontend broken, let's target the [feature/server-side-pagination](https://github.com/edgehog-device-manager/edgehog/tree/feature/server-side-pagination) branch instead of the main branch.

davidebriani (approved):

PR #800, opened 2025-01-15
Enable server-side pagination for `device_groups` and `target_groups`

OmarBrbutovic:
Updated queries for `device_groups`, and `target_groups` to support server-side pagination. This change ensures data fetching is paginated on the backend while maintaining the current client-side logic. A future update will optimize the respective tables to fully leverage pagination, avoiding an immediate rewrite of the client-side implementation.

Part of #779


frontend/src/pages/UpdateChannel.tsx
```diff
@@ -180,12 +189,14 @@ const UpdateChannelContent = ({
...
         );
-        const deviceGroups = root.getLinkedRecords("deviceGroups");
+
+        const deviceGroups = root.getLinkedRecord("deviceGroups");
         if (deviceGroups && targetGroupIds.size) {
-          deviceGroups.forEach((deviceGroup) => {
+          deviceGroups?.getLinkedRecords("edges")?.forEach((deviceGroup) => {
             if (targetGroupIds.has(deviceGroup.getDataID())) {
               deviceGroup.invalidateRecord();
             }
```
davidebriani, line 202: I am not sure if this works: you are iterating `deviceGroup` over `deviceGroups?.getLinkedRecords("edges")`, so `deviceGroup` is an edge and not the actual node.
`deviceGroup.getDataID())` and `deviceGroup.invalidateRecord()` may not give the intended behavior. Have you tested it?

davidebriani (changes requested):

frontend/src/pages/UpdateChannel.tsx
OmarBrbutovic: Thank you for pointing that out. You're absolutely right, in the old code, `deviceGroup` refers to an edge rather than the actual node, which could lead to unintended behavior. The updated code addresses this by explicitly extracting the node from the edge

davidebriani (approved):

PR #807, opened 2025-01-20
refactor(networks): make networks a property of the releases

noaccOS:
With this change, all the containers in a release will always have the same networks.

The GraphQL Schema is affected by this change, so if we decide to proceed with this the frontend should be updated as well


lusergit (approved):
lgtm

PR #809, opened 2025-01-27
refactor(containers): save network options as maps

noaccOS:
apply the same strategy as environment variables and volume options


backend/lib/edgehog/containers/network.ex
```diff
@@ -53,10 +55,9 @@ defmodule Edgehog.Containers.Network do
       public? true
     end
 
-    attribute :options, {:array, :string} do
+    attribute :options, :map do
       allow_nil? false
-      default []
-      public? true
+      default %{}
```
lusergit, line 60: Shouldn't we expose the network options to the user? I think `public? true` is the way to go here

noaccOS: Yes, it was removed by mistake. Fixed it

lusergit (approved):
:+1:

PR #810, opened 2025-01-27
feat(containers): send volumes to the device

noaccOS:
implement the logic for sending the volumes to the device.

the graphql api currently does not expose volumes, neither for creation nor queries.


backend/lib/edgehog/containers/container_volume.ex
```diff
@@ -0,0 +1,49 @@
...
+      attribute_type :uuid
+    end
+
+    belongs_to :volume, Edgehog.Containers.Volume do
+      primary_key? true
+      allow_nil? false
+      attribute_type :uuid
+    end
+  end
+
```
lusergit, line 44: Container <-> Volume should have a `mountpoint` attribute, so that when creating a new container we can send to the device the `binds` option and actually realte the container to the volume

lusergit (commented):
overall lgtm, just a minor tweak

lusergit (approved):

PR #812, opened 2025-02-05
Feat/image deployment

lusergit:
The current approach to check the state of a deployment is to query the device for each resource every time we react to a trigger. this PR introduces a `Deployment` caching layer that keeps track of the state of a particular resource on the device, reacting to device updates with triggers and avoiding re-querying the device.


backend/lib/edgehog/containers/image/changes/deploy_image_on_device.ex
```diff
@@ -0,0 +1,40 @@
...
+  use Ash.Resource.Change
+
+  alias Edgehog.Containers
+  alias Edgehog.Devices
+
+  @impl Ash.Resource.Change
+  def change(changeset, _opts, context) do
+    %{tenant: tenant} = context
+
+    Ash.Changeset.after_action(changeset, fn _changeset, deployment ->
```
davidebriani, line 32: We could be precise when naming variables to avoid any confusion: `image_deployment` instead of `deployment`.

```suggestion
    Ash.Changeset.after_action(changeset, fn _changeset, image_deployment ->
```

backend/lib/edgehog/containers/image/deployment.ex
```diff
@@ -0,0 +1,112 @@
...
+    belongs_to :image, Edgehog.Containers.Image do
+      attribute_type :uuid
+      public? true
+    end
+
+    belongs_to :device, Edgehog.Devices.Device
+  end
+
+  calculations do
+    calculate :ready?, :boolean, expr(state not in [:created, :sent, :error])
```
davidebriani, line 97: As rule of thumb, I would avoid defining a condition like this one by negation.
If we add more possible states in the future, we might forget to update this calculation, thus leading the ImageDeployment to result `ready` on unexpected states.

```suggestion
    calculate :ready?, :boolean, expr(state in [:pulled, :unpulled])
```

backend/lib/edgehog/containers/deployment/changes/create_deployment_on_device.ex
```diff
@@ -25,12 +25,11 @@ defmodule Edgehog.Containers.Deployment.Changes.CreateDeploymentOnDevice do
...
-  def change(changeset, _opts, _context) do
-    # After the transaction has been executed, i.e. all checks have passed
-    # and the deployment is in the data layer, start the deployment with
-    # the result.
+  def change(changeset, _opts, context) do
+    %{tenant: tenant} = context
+
     Ash.Changeset.after_action(changeset, fn _changeset, deployment ->
-      with :ok <- Containers.send_deploy_request(deployment) do
+      with :ok <- Containers.send_deploy_request(deployment, tenant: tenant) do
```
noaccOS, line 32: see https://github.com/edgehog-device-manager/edgehog/pull/813#discussion_r1943311539

backend/lib/edgehog/containers/image/deployment.ex
```diff
@@ -0,0 +1,111 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2024 SECO Mind Srl
```
noaccOS, line 4:
```suggestion
# Copyright 2025 SECO Mind Srl
```

davidebriani (approved):

PR #813, opened 2025-02-05
feat(containers): network deployment caching

lusergit:
Currently each time a device updates a property we recompute the whole state of a deployment by querying the device on related resource. The PR introduces a `Deployment` cache for the state of a network, so that the backend can just keep track of the updates from the device and update its state accordingly.


backend/lib/edgehog/containers/deployment/changes/check_networks.ex
```diff
@@ -22,28 +22,32 @@ defmodule Edgehog.Containers.Deployment.Changes.CheckNetworks do
...
     with {:ok, :created_images} <- Ash.Changeset.fetch_argument_or_change(changeset, :status),
          {:ok, deployment} <-
-           Ash.load(deployment, device: :available_networks, release: [containers: [:networks]]) do
-      available_network_ids =
-        Enum.map(deployment.device.available_networks, & &1.id)
+           Ash.load(deployment, device: [], release: [containers: [:networks]]) do
+      device = deployment.device
 
-      missing_networks =
+      deployments_ready? =
```
noaccOS, line 36:
```suggestion
      networks_ready? =
```

backend/lib/edgehog/containers/deployment/changes/create_deployment_on_device.ex
```diff
@@ -26,11 +26,8 @@ defmodule Edgehog.Containers.Deployment.Changes.CreateDeploymentOnDevice do
 
   @impl Ash.Resource.Change
   def change(changeset, _opts, _context) do
-    # After the transaction has been executed, i.e. all checks have passed
-    # and the deployment is in the data layer, start the deployment with
-    # the result.
     Ash.Changeset.after_action(changeset, fn _changeset, deployment ->
-      with :ok <- Containers.send_deploy_request(deployment) do
+      with :ok <- Containers.send_deploy_request(deployment, tenant: deployment.tenant_id) do
```
noaccOS, line 30:
```suggestion
      with :ok <- Containers.send_deploy_request(deployment) do
```

why would you need to specify the tenant? It would be a data error if it was different from the deployment's tenant in `send_deploy_request`

backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
```diff
@@ -52,22 +52,34 @@ defmodule Edgehog.Containers.ManualActions.SendDeployRequest do
...
+  defp deploy_networks(device, networks) do
+    networks =
+      networks
+      |> Enum.reject(&network_deployed?(&1, device))
+      |> Enum.uniq_by(& &1.id)
+
     Enum.reduce_while(networks, :ok, fn network, _acc ->
-      case Devices.send_create_network_request(device, network) do
-        {:ok, _device} -> {:cont, :ok}
+      case Containers.deploy_network(network.id, device.id, tenant: network.tenant_id) do
```
noaccOS, line 69: same as above

noaccOS (changes requested):

backend/lib/edgehog/containers/manual_actions/send_deploy_request.ex
lusergit: In this case if the tenant is not specified Ash throws an error:
```
** (Ash.Error.Invalid.TenantRequired) Queries against the Edgehog.Containers.Network.Deployment resource require a tenant to be specified
```

noaccOS: This is true but it makes me wonder: why do you not give the whole `network` object to `deploy_network`?
With you approach
- you need to re-fetch the data from the db
- you need to manually specify the tenant at this level

lusergit: The tenant here is required because it's creating a new resource. Resolving after some IRL discussion (broken glasses and shouting).

noaccOS (approved):

PR #814, opened 2025-02-05
feat!(containers): volume deployment caching

lusergit:
Currently each time a device updates a property we recompute the whole state of a deployment by querying the device on related resource. The PR introduces a Deployment cache for the state of a volume, so that the backend can just keep track of the updates from the device and update its state accordingly.

This also introduces a new deployment status `:created_volumes`, which affects the GraphQL Schema. If we decide to proceed with this change the frontend should be updated as well.


backend/lib/edgehog/containers/volume/deployment.ex
```diff
@@ -0,0 +1,107 @@
...
+  end
+
+  relationships do
+    belongs_to :volume, Edgehog.Containers.Volume do
+      attribute_type :uuid
+      public? true
+    end
+
+    belongs_to :device, Edgehog.Devices.Device
+  end
```
davidebriani, line 94: We can probably default to also using `attribute_public? false` on `belong_to` relationships that are `public`.

In this case I suspect we will just use these resources internally, and so we don't need these relationships to be public.

backend/lib/edgehog/containers/volume/deployment.ex
```diff
@@ -0,0 +1,107 @@
...
+
+  calculations do
+    calculate :ready?, :boolean, expr(state not in [:created, :sent, :error])
```
davidebriani, line 97:
```suggestion
    calculate :ready?, :boolean, expr(state in [:available, :unavailable])
```

PR #816, opened 2025-02-10
feat!(containers): deployment caching

lusergit:
Currently deployments are hand-checked each time a device trigger is called. This PR decouples the state of the deployment and the state of the underlying resources and caches the deployment state sent by the device. 

This change is breaking as it renames the old `status` attribute of a deployment into `state` (to use the same naming schema as the other caching layers, see #812 #813 #814 and #815) and introduces a new `resources_state` attribute to signal the fronted the state of the underlying resources related to the deployment. If this PR gets merged the frontend should be upgraded as well.


backend/lib/edgehog/containers/deployment/types/resources_state.ex
```diff
@@ -18,24 +18,14 @@
...
-      stopping: "The deploymen is stopping.",
-      created: "The deployment has been received by the backend and will be sent to the device.",
-      sent: "All the necessary resources have been sent to the device.",
-      # TODO: these are internal states that should not be exposed.
-      # Remove when reimplementing the deployment and its status as a state machine
       created_images: "The device has received the necessary image descriptions for the deployment.",
       created_networks: "The device has received all the network descriptions necessary for the deployment.",
       created_containers: "The device has received all the container descriptions necessary for the deployment.",
-      created_deployment: "The device has received the release description of the the deployment."
+      ready: "All the underlying resources needed for the deploymes have been received by the device."
```
davidebriani, line 28:
```suggestion
      ready: "All the underlying resources needed for the deployment have been received by the device."
```

davidebriani (changes requested):

noaccOS (approved):

PR #817, opened 2025-02-14
fix(containers): add missing triggers

lusergit:
Adding `available` triggers to edgehog, so that the conciliation process can correctly install the necessary interfaces for containers management


OmarBrbutovic (commented):
Since Trigger Delivery Policies are now supported by Edgehog, should we updated these triggers to use one?

lusergit:
> Since Trigger Delivery Policies are now supported by Edgehog, should we updated these triggers to use one?

Yes! I'll update

PR #820, opened 2025-02-21
Refactor auth token handling 

### LINKED ISSUE #819: Auth errors when cookie storage fails
The token used to authenticate API calls to the Edgehog backend is fetched directly from the session cookie.
This means that if the cookie cannot be set (for example due to browser settings) the frontend will stop to work even when a token is provided through normal means.
https://github.com/edgehog-device-manager/edgehog/blob/main/frontend/src/api/index.ts#L129

One way to fix this is to retrieve the token from the `Auth` context instead of relying on cookies for internal app data exchange.
https://github.com/edgehog-device-manager/edgehog/blob/main/frontend/src/contexts/Auth.tsx#L86

OmarBrbutovic:
### PR Update: Migration from `AuthConfig` to `Session` & Improved Session Handling  

This PR refactors the authentication and session handling mechanism by replacing `AuthConfig` with `Session`. Additionally, it enhances the `RelayEnvironment` initialization to support scenarios where no session is available.  

#### **Key Changes:**  
1. **Migration from `AuthConfig` to `Session`:**  
   - The `AuthConfig` context has been replaced with the `Session` context.

2. **Changes in `index.tsx`:**  
   - Introduced `SessionProvider` and `RelayProvider` to properly wrap components.
   - Ensured `RelayEnvironmentProvider` uses the dynamically created `relayEnvironment(session)`, allowing it to react to session changes.  

Closes #819

### DIFF
--- a/frontend/src/api/index.ts
+++ b/frontend/src/api/index.ts
@@ -1,7 +1,7 @@
 /*
   This file is part of Edgehog.
 
-  Copyright 2021-2023 SECO Mind Srl
+  Copyright 2021-2025 SECO Mind Srl
 
   Licensed under the Apache License, Version 2.0 (the "License");
   you may not use this file except in compliance with the License.
@@ -24,7 +24,7 @@ import type { FetchFunction, Variables, UploadableMap } from "relay-runtime";
 import type { TaskScheduler } from "relay-runtime";
 import ReactDOM from "react-dom";
 
-import { AuthConfig, loadAuthConfig } from "contexts/Auth";
+import type { Session } from "contexts/Session";
 
 const applicationMetatag: HTMLElement = document.head.querySelector(
   "[name=application-name]",
@@ -44,14 +44,15 @@ Please ensure that the 'BACKEND_URL' environment variable contains schema, e.g. 
 const fetchGraphQL = async (
   query: string | null | undefined,
   variables: Record<string, unknown>,
-  authConfig: AuthConfig,
+  tenantSlug: string,
+  authToken: string,
 ) => {
   const userLanguage = navigator.language; // TODO allow users to overwrite this
-  const apiUrl = new URL(`tenants/${authConfig.tenantSlug}/api`, backendUrl);
+  const apiUrl = new URL(`tenants/${tenantSlug}/api`, backendUrl);
   const response = await fetch(apiUrl.toString(), {
     method: "POST",
     headers: {
-      Authorization: `Bearer ${authConfig.authToken}`,
+      Authorization: `Bearer ${authToken}`,
       "Content-Type": "application/json",
       "Accept-Language": userLanguage,
     },
@@ -65,13 +66,14 @@ const uploadGraphQL = async (
   query: string | null | undefined,
   variables: Record<string, unknown>,
   uploadables: UploadableMap,
-  authConfig: AuthConfig,
+  tenantSlug: string,
+  authToken: string,
 ) => {
-  const apiUrl = new URL(`tenants/${authConfig.tenantSlug}/api`, backendUrl);
+  const apiUrl = new URL(`tenants/${tenantSlug}/api`, backendUrl);
   const request: RequestInit = {
     method: "POST",
     headers: {
-      Authorization: `Bearer ${authConfig.authToken}`,
+      Authorization: `Bearer ${authToken}`,
     },
   };
   const formData = new FormData();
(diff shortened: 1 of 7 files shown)


frontend/src/contexts/Auth.tsx
```diff
@@ -70,7 +70,7 @@ function saveAuthConfig(
   }
 }
 
-function loadAuthConfig(): AuthConfig | null {
+let currentAuthConfig: AuthConfig | null = (() => {
```
Pavinati, line 73: defining this as a mutable var is not the right way share date using react.
Here we should leverage `AuthProvider` to handle the `AuthContext` value properly by:
- initializing it (read previous session from cookies).
  Done [here](https://github.com/edgehog-device-manager/edgehog/blob/main/frontend/src/contexts/Auth.tsx#L107)
- exposing the token to the app so it can be used without checking cookies.
  This requires including the token in the [context value](https://github.com/edgehog-device-manager/edgehog/blob/main/frontend/src/contexts/Auth.tsx#L92)
- updating it after login/logout.
  Done [here](https://github.com/edgehog-device-manager/edgehog/blob/main/frontend/src/contexts/Auth.tsx#L112), but need to be updated to include updating the current token from previous point

Pavinati (changes requested):

frontend/src/contexts/Session.tsx
```diff
@@ -0,0 +1,97 @@
...
+  tenantSlug: string;
+  authToken: string | null;
+};
+
+const SESSION_CONFIG_VERSION = 1;
+
+type SessionContextValue = {
+  session: Session | null;
+  saveSession: (session: Session | null, persist?: boolean) => void;
+};
```
Pavinati, line 35: In my opinion, components using this context shouldn't care if the session is stored or not.
Also, a session should always be present. If there is no previous session, a new one will be created right away starting from scratch.

```suggestion
type SessionContextValue = {
  session: Session;
  updateSession: (session: Session) => void;
};
```

frontend/src/contexts/Session.tsx
```diff
@@ -0,0 +1,97 @@
...
+    session = null;
+  }
+  if (_.get(session, "_version") === SESSION_CONFIG_VERSION) {
+    return _.omit(session, "_version") as Session;
+  }
+  return null;
+};
+
+const SessionProvider = ({ children }: { children: React.ReactNode }) => {
+  const [session, setSession] = useState<Session | null>(loadSession());
```
Pavinati, line 55: Similarly, here we should try to restore the previous session. If none is found, we will initialize a new one with just the tenant slug

frontend/src/contexts/Session.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const saveSession = useCallback(
```
Pavinati, line 57: I would keep the "save to cookie" part of this function out of the component (as a module function).
This way if we change how we save stuff later we have a concise spot to edit.

Pavinati (changes requested):

frontend/src/index.tsx
```diff
@@ -24,21 +24,43 @@ import { RelayEnvironmentProvider } from "react-relay/hooks";
...
+import SessionProvider, { useSession } from "contexts/Session";
 import AuthProvider from "contexts/Auth";
 import I18nProvider from "i18n";
 import App from "./App";
+import LoginPage from "pages/Login";
 import "./index.scss";
 
+function AppOrLogin() {
+  const { session } = useSession();
+  if (session) {
```
Pavinati, line 36: session can never be null

frontend/src/contexts/Auth.tsx
```diff
@@ -157,12 +102,12 @@ const AuthProvider = ({ children, fetchGraphQL }: AuthProviderProps) => {
...
     let mounted = true;
     if (isValidatingInitialConfig) {
-      validateAuthConfig(authConfig).then((isValid) => {
+      validateSession(session).then((isValid) => {
         if (!mounted) {
           return;
         }
         if (!isValid) {
-          updateAuthConfig(null);
+          updateSession({ tenantSlug: session.tenantSlug, authToken: null });
```
Pavinati, line 110: considering this implementation, we might update or remove the following TODO comment

Pavinati (changes requested):

frontend/src/api/index.ts
```diff
@@ -153,11 +130,39 @@ const relayScheduler: TaskScheduler = {
   },
 };
 
-const relayEnvironment = new Environment({
-  network: Network.create(fetchRelay),
-  store: new Store(new RecordSource()),
-  scheduler: relayScheduler,
-});
+const relayEnvironment = (session: Session | null) => {
```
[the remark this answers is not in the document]
Pavinati: actually, thinking about it a bit more. There is one issue here.
Before login, we might not have a tenant slug and we'd like to initialize our app before the login phase.

Pavinati (approved):

PR #821, opened 2025-02-21
Expose associated UpdateCampaign from OTA Operation

### LINKED ISSUE #356: Expose the associated UpdateCampaign (if any) from an OTA Operation

OmarBrbutovic:
- Added a new column to display the associated UpdateCampaign name in
- Implemented a link to the `Update Campaign` edit page using

Closes #356

<p>

Old
image

New
image

</p>
</details>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.10] - Unreleased
 ### Added
-- Managed OTA operations expose the update target that created them in graphql ([#356](https://github.com/edgehog-device-manager/edgehog/issues/356).
+- Managed OTA operations expose the update target that created them in graphql ([#356](https://github.com/edgehog-device-manager/edgehog/issues/356)).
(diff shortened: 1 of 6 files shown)


frontend/src/pages/Device.tsx
```diff
@@ -181,6 +181,12 @@ const DEVICE_OTA_OPERATIONS_FRAGMENT = graphql`
       baseImageUrl
       status
       createdAt
+      updateTarget {
+        updateCampaign {
+          id
+          name
+        }
+      }
```
davidebriani, line 189: This shouldn't be needed, since some lines below we're already spreading the `OperationTable_otaOperations` fragment which defines the data requirements of our OperationTable component.
Indeed, the component actually defines and uses that fragment to access data.

davidebriani (changes requested):

davidebriani (approved):

PR #822, opened 2025-03-03
Update i18n extract output path

OmarBrbutovic:
- Updated the i18n extraction command to save translations directly to `src/i18n/langs/en.json`, eliminating the need for a manual move from `translations.generated.json`.
- Add script to update i18n translation files. This script ensures that all language JSON files stay in sync with the English master file (`en.json`). It performs the following updates:
  - Adds missing translation keys using the English default message.
  - Removes obsolete keys that are no longer in `en.json`.
  - Ensures existing keys include a `defaultMessage`, if missing.
  - Maintains the exact key order from `en.json` for consistency.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -76,7 +76,7 @@
     "pretest": "npm run init",
     "coverage": "TZ=UTC vitest run --coverage",
     "precoverage": "npm run init",
-    "i18n:extract": "formatjs extract 'src/**/*.ts*' --ignore '**/*.d.ts' --out-file translations.generated.json --id-interpolation-pattern '[sha512:contenthash:base64:6]'",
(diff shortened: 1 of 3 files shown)


davidebriani (approved):

davidebriani:
Yes!
PRs like this one are always welcome.
I especially liked that:
- it brings real value to the project, simplifying and speeding up our development practices.
- the PR was well described and presented, clearly communicating why it was needed and how the solution was implemented.

One suggestion for future additions would be that a JavaScript / NodeJS script would have been a better fit:
- Developers working on the frontend doesn't need to install and use Elixir.
- JavaScript has native and better support for working with JSON files.
- The JavaScript ecosystem and NPM probably already have some libraries for working with the format used in our i18n / translation files.

Of course that's just food for thought and an Elixir script is fine.
Nice work Omar!

PR #831, opened 2025-06-06
docs(admin): fix tenant provisioning example to use Edgehog.Tenants

### LINKED ISSUE #811: Issues with Edgehog.Provisioning alias.
Hello,
I started to playing with Edgehog and go through the documentation from here: https://docs.edgehog.io/0.9/deploying_with_kubernetes.html

At the at the tenant creation there is probably a typo in this line
[iex> alias Edgehog.Provisioning](https://github.com/edgehog-device-manager/edgehog/blob/fa032dc4fca80063f872bf759d17af40df6f2b1d/doc/pages/admin/deploying_with_kubernetes.md?plain=1#L783)

So after these steps I am getting still that message:
`** (UndefinedFunctionError) function Edgehog.Provisioning.provision_tenant/1 is undefined (module Edgehog.Provisioning is not available)`

Then I found that there is another alias named `Edgehog.Tenants` so after when I typed `alias Edgehog.Tenants` 
and tried to provision new tenant using `Tenants.provision_tenant()` method then tenant was provisioned.

davidebriani:
Update the Kubernetes deployment documentation to use the correct module and function for tenant provisioning (Edgehog.Tenants and Tenants.provision_tenant/1 instead of Edgehog.Provisioning).

Closes #811.

### DIFF
--- a/doc/pages/admin/deploying_with_kubernetes.md
+++ b/doc/pages/admin/deploying_with_kubernetes.md
@@ -836,7 +836,7 @@ The following commands will create a database entry representing the tenant, wit
 Astarte cluster and Realm.
 
 ```elixir
-iex> alias Edgehog.Provisioning
+iex> alias Edgehog.Tenants
 iex> tenant_name = "<TENANT-NAME>"
 iex> tenant_slug = "<TENANT-SLUG>"
 iex> tenant_public_key = """
@@ -847,7 +847,7 @@ iex> realm_name = "<ASTARTE-REALM-NAME>"
 iex> realm_private_key = """
 <ASTARTE-REALM-PRIVATE-KEY>
 """
-iex> {:ok, tenant} = Provisioning.provision_tenant(
+iex> {:ok, tenant} = Tenants.provision_tenant(
   %{
     name: tenant_name,
     slug: tenant_slug,


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+### Fixed
+- Documentation: Fix tenant provisioning example to use `Edgehog.Tenants` and `Tenants.provision_tenant/1` instead of the incorrect `Edgehog.Provisioning` ([#811](https://github.com/edgehog-device-manager/edgehog/pull/811)).
```
noaccOS, line 9: I would not include this change in the changelog at all
IMO `Fixes` should be for functionalities of the application and not Docs/Chores, as this should be a changelog for the end user

noaccOS (changes requested):

CHANGELOG.md
davidebriani: Agreed, I removed the changelog.

noaccOS (approved):

PR #843, opened 2025-08-07
feat: Add GraphQL API to delete releases with cascading cleanup

OmarBrbutovic:
Implements release deletion API that automatically handles related resource cleanup and deployment management.

- Added destroy action to Release resource with custom change handler
- Created HandleRelatedResourcesDeletion change to manage cleanup logic
- Handles ReleaseContainers and ContainerNetwork join table cleanup


backend/lib/edgehog/containers/release/changes/cleanup_containers.ex
lusergit, line 1: overall good approach! you can demand more things to the data layer but it's ok, I think the `cleanup_containers.ex`can be something on the lines of
```elixir
defmodule Edgehog.Containers.Release.Changes.CleanupContainers do
  @moduledoc false

  use Ash.Resource.Change
  alias Edgehog.Containers

  @impl Ash.Resource.Change
  def change(changeset, _opts, context) do
    release = changeset.data
    %{tenant: tenant} = context

    with {:ok, release} <- Ash.load(release, :containers, tenant: tenant) do
    Ash.Changeset.after_action(changeset, fn changeset, _deleted_release ->
      containers = release.containers

      for container <- containers do
        case Containers.destroy_container_if_dangling(container, tenant: tenant) do
          {:ok, _} -> :ok
          # What to do here? for the moment logging on errors might be enough
          {:error, changeset} -> Logger.warning(...)
        end
      end

      {:ok, changeset}
    end)
    end
  end
end
```

backend/lib/edgehog/containers/container/manual_actions/destroy_if_dangling.ex
```diff
@@ -0,0 +1,50 @@
...
+      if container.dangling? do
+        # Container is dangling, destroy it and trigger image cleanup
+        image_id = container.image_id
+
+        with {:ok, destroyed_container} <- Ash.destroy(changeset) do
+          # Try to destroy the image if it's also dangling
+          _ = Containers.destroy_image_if_dangling(image_id, tenant: tenant)
+
+          {:ok, destroyed_container}
+        end
```
lusergit, line 43: I think we want to wrap this in a [`Ash.Changeset.after_action/3`](https://hexdocs.pm/ash/3.5.33/Ash.Changeset.html#after_action/3) (or after transaction, depends whether you want to possibly revert changes or just handle deletion of dangling resources separately)

lusergit (changes requested):
Almost there! I have suggested a couple of changes, but you're on track 👍

PR #844, opened 2025-08-07
feat: Add GraphQL API to delete application with cascading cleanup

ArnelaL:
- Added destroy action to Application resource with custom change handler
- Created DestroyRelatedReleases change to manage cleanup logic

depends on: #843


backend/lib/edgehog/containers/changes/destroy_related_releases.ex
```diff
@@ -0,0 +1,58 @@
...
+  def change(changeset, _opts, _context) do
+    application =
+      Ash.load!(
+        changeset.data,
+        [:releases],
+        reuse_values?: true
+      )
+
+    Enum.reduce_while(application.releases, changeset, fn release, changeset_acc ->
+      case Ash.destroy(release, tenant: changeset.tenant) do
```
lusergit, line 34: I don't think that's what we want to do here. This would make inconsistent deletions: some releases get deleted while some others not.

I suggest to use the same approach as #843, first make a change that checks if an application can be deleted (are all its releases deletable?) then a change to run post-transaction hooks to garbage collect its releases (take a look at #856 for this passage)

backend/lib/edgehog/containers/changes/destroy_related_releases.ex
```diff
@@ -0,0 +1,50 @@
...
+        %{valid?: valid} = Ash.Changeset.for_destroy(release, :destroy)
+        valid
+      end)
+
+    if undeletable_releases == [] do
+      Enum.each(application.releases, fn release ->
+        Ash.destroy!(release, tenant: tenant)
+      end)
+
+      changeset
```
lusergit, line 39: This has to be wrapped in an `after_action`, we do not want to delete all the releases and then something goes wrong with the deletion of the application or vice versa, leaving in an inconsistent state

```suggestion
Ash.Changeset.after_action(changeset, fn _changeset, _application ->
  with {:ok, _} <- result do 
      Enum.each(application.releases, fn release ->
        Ash.destroy!(release, tenant: tenant)
      end)
  end
  result
end)
```

lusergit (changes requested):
almost there! just a small thing

PR #846, opened 2025-08-08
feat(containers): optionally sepcify a system model for an application

lusergit:
Supports specifying a system model in the definition of an application.

Part of a broader series of PRs aimed at supporting system model integration with container management


lusergit:
Superseded by #847

PR #847, opened 2025-08-08
feat(containers): support optional system model for application

lusergit:
Adds support for users to specify a system model for an aplpication. Edgehog will also prevent the user from deploying an application on incompatible devices


OmarBrbutovic (approved):
:rocket:

Dam-99 (approved):

noaccOS (approved):

PR #848, opened 2025-08-11
fix(containers): report correct states of deployments on Device page

Emina-M:
- Update GraphQL schema on frontend.
- Update frontend to reflect changes in deployment attributes. Rename `status` to `state` and add two new states: CREATED and SENT.
- Display resources state alongside deployment state in Deployed Applications table.

<details>
<summary>
Screenshots
</summary>

<kbd>Screenshot from 2025-08-12 13-37-43

<kbd>Screenshot from 2025-08-12 13-38-11

<kbd>Screenshot from 2025-08-12 13-38-21

</details>

### DIFF
--- a/frontend/src/components/DeployedApplicationsTable.tsx
+++ b/frontend/src/components/DeployedApplicationsTable.tsx
@@ -25,6 +25,7 @@ import semver from "semver";
 
 import type { DeployedApplicationsTable_PaginationQuery } from "api/__generated__/DeployedApplicationsTable_PaginationQuery.graphql";
 import type {
+  ApplicationDeploymentResourcesState,
   ApplicationDeploymentState,
   DeployedApplicationsTable_deployedApplications$key,
 } from "api/__generated__/DeployedApplicationsTable_deployedApplications.graphql";
@@ -53,6 +54,7 @@ const DEPLOYED_APPLICATIONS_TABLE_FRAGMENT = graphql`
         node {
           id
           state
+          resourcesState
           release {
             id
             version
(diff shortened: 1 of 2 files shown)


frontend/src/components/DeployedApplicationsTable.tsx
```diff
@@ -196,48 +252,89 @@ const statusMessages = defineMessages<DeploymentStatus>({
...
+  resourcesState: DeploymentResourcesState;
+}) => (
+  <div className="d-flex align-items-center">
+    <Icon
+      icon={resourcesState === "INITIAL" ? "spinner" : "circle"}
+      className={`me-2 ${resourcesStateColors[resourcesState]} ${
+        resourcesState === "INITIAL" ? "fa-spin" : ""
+      }`}
+    />
+    <FormattedMessage id={resourcesStateMessages[resourcesState].id} />
```
OmarBrbutovic, line 315: This is more of a backend-logic question, so we should also get @lusergit  input.

From my understanding, the *Deployment Resources State* goes through all the states in sequence. For the UI, it might be better to show a spinner for states like `CREATED_IMAGES`, `CREATED_NETWORKS`, etc., instead of a solid color, signaling to the user that something is still happening in the backend.

image



Alternatively, once a step is completed, we could display the next step in an “-ING” form (e.g., `CREATING_IMAGES` instead of `CREATED_IMAGES`) until it finishes, but that might be to much

OmarBrbutovic, line 315: This is more of a backend-logic question, so we should also get @lusergit  input.

From my understanding, the *Deployment Resources State* goes through all the states in sequence. For the UI, it might be better to show a spinner for states like `CREATED_IMAGES`, `CREATED_NETWORKS`, etc., instead of a solid color, signaling to the user that something is still happening in the backend.

image



Alternatively, once a step is completed, we could display the next step in an “-ING” form (e.g., `CREATING_IMAGES` instead of `CREATED_IMAGES`) until it finishes, but that might be to much

lusergit: Yes, the backend goes trough all the states in the sequence, so only when the state is `ready` the underlying resources of a deployment are actually ready,  I agree with your point of displaying a spinner to signal users that some things are still happening in the background :+1:

OmarBrbutovic (approved):

PR #851, opened 2025-08-12
refactor(container): remove 'image' from 'CreateContainerRequest'

Dam-99:
The `image` key is redundant, as edgehog already sends the `image_id` in the `CreateContainerRequest`


lusergit (approved):
:fire:

PR #852, opened 2025-08-13
fix(containers): correctly display resources state

lusergit:
correctly handles deployment by demanding device requests to `after_transaction` correctly configures the network of containers, without the default separated network

Previously deployng applications resulted in race conditions in the cashing of `deployment` resources:
- edgehog was sending to the device the request for creating resources _before_ the end of the database transaction
- the device responded to edgehog with the state of its underlying resource, this resulted in an error
- the creation of the  `deployment` resource on the device was also wrapped in a confusing set of actions, mimicking the `after_transaction/3` behavior
- `CreateImageRequest` interface used the wrong ttl


backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -281,6 +281,12 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
 
         "Error" ->
           Containers.mark_deployment_as_errored(deployment, message, tenant: tenant)
+
+        "Deleting" ->
+          Containers.mark_deployment_as_deleting(deployment, tenant: tenant)
+
+        "Updating" ->
+          Containers.mark_deployment_as_deleting(deployment, tenant: tenant)
```
OmarBrbutovic, line 289: is there a state called "Updating" ?

lusergit: Yes, That's one of the possible states in the [`DeploymentEvent` interface](https://github.com/edgehog-device-manager/edgehog/blob/feature/application-management/backend/priv/astarte_resources/interfaces/io.edgehog.devicemanager.apps.DeploymentEvent.json), we are not showing to the user such state directly at the moment, as the update procedure consists in 

- deploying the new release of an application
- once the new release is up and running stop the old one
- once the old one is stopped, delete it

However we're not displaying the `updating` state of a deployment to the user for the moment, and since what will eventually happen is the old deployment deletion for the moment we can mark it as in deletion, until we'll have a better representation of such state

Dam-99 (changes requested):

PR #864, opened 2025-08-19
chore(containers): adds errors to astarte device communication

lusergit:
Adds support for more detailed errors in the communication with astarte devices, distinguishing between 404 errors (the device is offline) and other types of errors (for now generic errors only, more can be added).

### DIFF
--- a/backend/lib/edgehog/error/astarte_api_error.ex
+++ b/backend/lib/edgehog/error/astarte_api_error.ex
@@ -20,7 +20,7 @@
 
 defmodule Edgehog.Error.AstarteAPIError do
   @moduledoc "Used when Astarte replies with an APIError"
-  use Splode.Error, fields: [:status, :response], class: :invalid
+  use Splode.Error, fields: [:status, :response, :device_id, :interface], class: :invalid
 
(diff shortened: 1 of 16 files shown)


backend/lib/edgehog/error/astarte_api_error.ex
```diff
@@ -32,13 +32,21 @@ defmodule Edgehog.Error.AstarteAPIError do
     error_details =
       case error.response do
         %{"error" => %{"detail" => error_message}} -> error_message
-        response -> Jason.encode(response)
+        %{"errors" => %{"detail" => error_message}} -> error_message
+        response -> Jason.encode!(response)
       end
 
+    # TODO: change old usages of `AstarteAPIError` to include device_id and
+    # interface
```
Dam-99, line 40: this operation is also done in `Edgehog.Error`, is teh plan to remove the direct usage of `AstarteAPIError` and substitute it with `Edgehog.Error`? in which case, it shouldn't be necessary to compute `error_details` again here and could be removed in the future

Dam-99 (approved):

noaccOS (approved):

PR #865, opened 2025-08-20
feat(containers): improve Release page layout

ArnelaL:
- Restyled the Release page for better readability and presentation of container data

[Screencast from 2025-08-20 13-56-15.webm](https://github.com/user-attachments/assets/23eabfae-2a74-477f-89c0-070b253f9abd)

### DIFF
--- a/frontend/src/components/ContainersTable.tsx
+++ b/frontend/src/components/ContainersTable.tsx
@@ -18,8 +18,13 @@
   SPDX-License-Identifier: Apache-2.0
 */
 
+import React, { useState } from "react";
 import { FormattedMessage } from "react-intl";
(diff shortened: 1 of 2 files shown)


davidebriani (changes requested):
That's a good overall improvement.
I would evaluate these changes:
1. Collapsible are ok, but it's better if the user don't have to click around too much. I would have the collapsible elements open by default, so that the user just need to scroll the page to review the whole release definition.
2. I would remove the first `Containers` collapsible, it seems unnecessary.
3. Instead of `Container 1`, `Container 2`, etc., we can use the image references (e.g. `docker.io//minio/minio:latest`) from the container definition.

PR #868, opened 2025-08-20
refactor: campaigns

lusergit:
Based on #864 

This refactor of `UpdateCampaign` infrastructure (namely, supervisors, registries and the resumer) has the benefit of not being dependent on the type of campaign one might want to run.


backend/lib/edgehog/campaigns/supervisor.ex
```diff
@@ -42,7 +44,19 @@ defmodule Edgehog.UpdateCampaigns.Supervisor do
...
+    _other ->
+      defp children, do: @base_children ++ [update_campaigns_resumer()]
+
+      defp update_campaigns_resumer do
+        update_campaigns_stream =
+          UpdateCampaign
+          |> Ash.Query.for_read(:read_all_resumable)
+          |> Ash.stream!()
+
+        {Resumer, update_campaigns_stream}
```
davidebriani, line 59: `Resumer` is not aliased in the file, the whole module name is needed:

```suggestion
        {Edgehog.Campaigns.Resumer, update_campaigns_stream}
```

davidebriani (changes requested):

PR #873, opened 2025-08-21
feat: Filter available applications by device system model

JasminaPiric:
- Added application filtering to show only apps supported by the selected device
- Ensured apps are displayed if they require the same system model as the device or if no system model requirement exist

<p>
All existing applications:
image

All devices:
image

Supported apps on device with system model:
image

Apps with no specified system model:
image
</p>
</details>

### DIFF
--- a/frontend/src/pages/Device.tsx
+++ b/frontend/src/pages/Device.tsx
@@ -281,6 +281,9 @@ const DEVICE_DEPLOYED_APPLICATIONS_FRAGMENT = graphql`
     device(id: $id) {
       id
       online
+      systemModel {
+        name
+      }
       ...DeployedApplicationsTable_deployedApplications
     }
   }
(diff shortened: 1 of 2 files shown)


frontend/src/pages/Device.tsx
```diff
@@ -281,6 +281,9 @@ const DEVICE_DEPLOYED_APPLICATIONS_FRAGMENT = graphql`
     device(id: $id) {
       id
       online
+      systemModel {
+        name
+      }
```
davidebriani, line 286: The logic is there, but it's better to use the ID of the system model instead of its name to do comparisons and filters.

davidebriani (changes requested):

frontend/src/pages/Device.tsx
davidebriani: If the filter expects an Int, it's wrong and should be changed.
The API returns IDs in the form of strings, so it does not make sense to accept Int in a filter input (and it's probably never been used, so we can change the API).

Since we should verify whether we can easily change the backend to accept strings, you're solution is a good one and we can proceed with merging the PR as it is now.

davidebriani (approved):

PR #896, opened 2025-09-04
chore(containers): replicate update campaign behavior for deployments

lusergit:
Replicates the update campaign behavior for the deployment campaigns.

This is part of the effort in creating deployment campaigns, bringing the idea of update campaigns to the application management features of edgehog


backend/lib/edgehog/deployment_campaigns/deployment_target.ex
```diff
@@ -68,11 +70,81 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentTarget do
...
+      This is useful when resuming an deployment campaign to know which
+      targets need to setup a retry timeout.
+      """
+
+      argument :deployment_campaign_id, :integer, allow_nil?: false
+
+      prepare build(load: [deployment: :status])
+
+      filter expr(deployment_campaign_id == ^arg(:deployment_campaign_id))
+      filter expr(not is_nil(deployment) and deployment.status == :pending)
```
davidebriani, line 86: Deployments currently have a `state`, not a `status`.

Besides, we could evaluate to choose one between `status` or `state` and use that phrasing for every resource that use a similar concept, so that we don't keep confusing the two. Food for though for our future selves

backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/executor.ex
```diff
@@ -33,8 +54,566 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Executor do
...
+    new_data = %{
+      data
+      | failed_count: failed_count,
+        in_progress_count: in_progress_count,
+        available_slots: available_slots
+    }
+
+    timeout_actions =
+      tenant_id
+      |> Core.list_targets_with_pending_deployment(deployment_campaign_id)
```
davidebriani, line 184: I have a small doubt here, which might also be valid for Update Campaigns since they share similar code/concepts.

When resuming a campaign, we get the `in_progress_count` of targets that are in the `in_progress` status.
So I would expect that these are the targets that should be resumed and retried, occupying available slots in the campaign.
However, we actually retrieve the actual targets by looking at those with a pending deployment (`Core.list_targets_with_pending_deployment`), which could be a different thing.

In addition, retrieving targets in this way doesn't account for the fact that a previously "in progress" target might have already progressed to a state like `sent` or `started` (different from the default `pending`). So we could potentially had "in progress" targets that are now ignored when resuming the campaign (even though they account for the `in_progress_count` number, thus reducing available slots). I fear we might leave these targets dangling without the ability to retry deployment in case of error.

backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/executor.ex
```diff
@@ -33,8 +54,566 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Executor do
...
+            {:next_state, :campaign_success, data}
+        end
+    end
+  end
+
+  def handle_event(:internal, {:deployment_target, target}, :deployment, data) do
+    # We occupy a slot since we're rolling out an update
+    new_data = occupy_slot(data)
+
+    case Core.deploy(target, new_data.release, new_data.deployment_mechanism) do
```
davidebriani, line 255: Note that `Core.deploy` just creates the Deployment resource once (which automatically tries to deploy itself to the device, if I remember correctly).
However, it will not wait for successful deployment to the device, nor it will retry it in case of failure.

The fact is that `Core.deploy` returns `{:ok, :already_deployed}` after the first invocation, but that doesn't mean that the deployment was successfully deployed to the device.
However `handle_event(:internal, {:already_deployed, target}, :deployment, data)` handles the "already deployed" case by marking the target as successful.

backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/executor.ex
```diff
@@ -33,8 +54,566 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Executor do
...
+  def handle_event(:info, {:deployment_updated, deployment}, _state, data) do
+    # Event generated from PubSub when a Deployment is updated
+    additional_actions =
+      case Core.deployment_ready?(deployment) do
+        :ready ->
+          [internal_event({:deployment_success, deployment})]
+
+        {:not_ready, _missing_resources} ->
+          []
+      end
```
davidebriani, line 429: So `Core.deployment_ready?(deployment)` now returns a boolean.
Hence, we either change the function or we change the `case` cases.

davidebriani (changes requested):

Dam-99 (approved):

PR #898, opened 2025-09-04
feat(containers): use delivery policy on all Astarte triggers

davidebriani:
Ensure that all triggers that need to deliver messages from Astarte to Edgehog employ the delivery policy defined by Edgehog. This ensures that delivery of messages is retried in case of errors in receiving them.


lusergit (approved):
🚀

PR #899, opened 2025-09-04
feat(containers): release multiple system models - UI

Dam-99:
Update the frontend to manage the changes in Applications, Releases and System Models:
- SystemModels no longer visible when creating an application
- multiple SystemModels requirements for a Release
- Application deployment on Devices only for compatible Releases

image

image
image

image


frontend/src/components/AddAvailableApplications.tsx
```diff
@@ -137,9 +144,17 @@ const AddAvailableApplications = ({
...
+      value: release.id,
+      label: release.version,
+      // release should be enabled if:
+      // systemModelName is not null/undefined || release.systemModels.length != 0 || release.systemModels contains systemM │ odelName
+      disabled: !(
+        !(systemModelName === null || systemModelName === undefined) ||
+        !release.systemModels.length ||
+        (systemModelName !== undefined &&
+          systemModelName in release.systemModels)
+      ),
```
osmanhadzic, line 157:
```suggestion
  return selectedApplication.node.releases.edges.map(({ node: release }) => {
    const systemModelNames = release.systemModels?.map((sm) => sm.name) ?? [];

    const hasSystemModel = Boolean(systemModelName);
    const matchesSystemModel =
      hasSystemModel && systemModelNames.includes(systemModelName);
    const appliesToAll = systemModelNames.length === 0;

    const enabled = matchesSystemModel || appliesToAll;

    return {
      value: release.id,
      label: release.version,
      disabled: !enabled,
    };
```

osmanhadzic: This code is a bit hard to read and maintain. Let’s apply these suggestions to improve it.

osmanhadzic (changes requested):

frontend/src/components/AddAvailableApplications.tsx
Dam-99: Yeah I agree it was pretty unreadable, I was trying to make it simple but I was too fixated it keeping the code as contained as possible

osmanhadzic: I get what you mean. Just keep in mind you’re not the last person who’ll work on this code, so making it readable and maintainable will really help in the future.

osmanhadzic (approved):

ArnelaL (approved):

PR #901, opened 2025-09-08
feat(containers): handle Astarte triggers on available device mappings

davidebriani:
Add support for handling incoming trigger events about device mappings when the device acknowledges their definition and assert whether each device mapping is present or not.

This change also adds the Astarte trigger template that Edgehog will automatically installs in Astarte to ensure that trigger events are delivered to Edgehog.


Dam-99 (approved):

lusergit (approved):
:rocket:

PR #903, opened 2025-09-08
fix(containers): typo in Astarte message for container creation

davidebriani:
Fix a casing issue that resulted in Astarte failing to handle messages on the container creation Astarte interface.


lusergit (approved):
🐛

PR #920, opened 2025-09-12
feat(contianers): aggregated deployment readines

lusergit:
`ready?` calculation on deployment resources is more descriptful, reporting, when not ready, the non-ready resources.

deployment campaigns when relying on the backend to compute the readyness of a deployment (and mark the target as successful/failed/etc.) can have a more descriptive view of what's happening, with missing resources to handle.


OmarBrbutovic (approved):

backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/executor.ex
```diff
@@ -420,19 +420,21 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Executor do
...
-        {_, :error} -> [internal_event({:deployment_failure, deployment})]
-        {_, _} -> []
+      case ready? do
+        :ready ->
+          [internal_event({:deployment_success, deployment})]
+
+        {:not_ready, missing} ->
+          if errors?(missing),
+            do: [internal_event({:deployment_failure, deployment})],
+            else: []
```
davidebriani, line 437: This block of code made me think.

Maybe we could find a better way to expose the status of the whole deployment from the Deployment resource.

It also seems a bit convoluted to build a single list of different resources, where some elements could be a `{_deployment, :error}` tuple.

Let's start from scratch and let's look at what we really need to know here in the executor:
1. if a deployment has stopped progressing because it succeeded
2. if a deployment has stopped progressing because it encountered an error
3. if a deployment is still progressing

Note that we don't need to know more than that here, like the exact resources that are still missing device acknowledgement.

Besides, my suspicion is that when rolling out a deployment we will not receive any error by the device regarding specific resources, since the only Astarte interface where the device can report errors is the `io.edgehog.devicemanager.apps.DeploymentEvent` interface.
During rollout, the device just acknowledges the received messages, so it's likely we only get errors when:
- The device signals an error about the deployment. It could be due to an error on a specific container or network, but it is advertised as a deployment error.
- The device does not reply, and we make the rollout operation time out.

On one hand we know when individual resources (including Deployment) are successfully rolled out, on the other hand we receive a single error to know when something breaks. Incidentally, this error is saved on the Deployment resource by setting its state to `:error` (+ a `:last_error_message`)

So it seems to me that unfortunately we are mixing things up a bit, since the Deployment resource could be successfully rolled out and yet we put it in an error state because another resource (e.g. Container) encountered an error during rollout. In this sense, we are using the Deployment `state` as both the state of the single Deployment resource and to report a general error state for all of its resources.

Additionally, it would be nice to start distinguishing between:
- the fact that the device acknowledges a resource definition, an operation that typically reaches a terminal state (e.g. `:pending` -> `:sent` -> `:received`)
- the fact that the resource is in a certain state on the device: e.g. `:stopped` -> `:starting` -> `:started` -> `:stopping` -> `:stopped`.
 
Given these considerations, let's go back to our three-parts goal in the Campaign executor.
We could compute that with a `rollout_state` calculation on the Deployment resource, which returns:
- `:rolled_out`, if all the resources (including Deployment) are successfully rolled out.
- `:progressing` if not all resources are rolled out, and the deployment is not reporting an error.
- `:error` if not all resources are rolled out, and the deployment is reporting an error (currently, `:state = :error`).

We could even type this well with a `RolloutState` enum struct.

To keep it simple, we probably don't need to compute list of missing resources, or have complex `:ready?` manual calculations. It probably suffices to have `:rolled_out?` calculations on resources derived from their `:state`. Ideally, as pointed out, in the future on each resource we should really have a `:rollout_state` that ultimately transitions to the terminal rollout state `:rolled_out`. And have a `:rollout_error` field on each resource instead of conflating different kind of errors in the same Deployment field `:state = :error`. In this case, the calculation that we need to use for the executor would just need to gather the RolloutState of all resources and combine them to get a summarized RolloutState.

davidebriani (changes requested):

PR #922, opened 2025-09-12
chore(tooling): configure development tooling to quickly setup an Edgehog instance and connect a simulated device

davidebriani:
This change aims to do a small improvement to our tooling to make local development and testing simpler and faster.

The PR configures the Docker Compose environment with default resources, used to setup a sample tenant, and some tooling like pgAdmin to quickly inspect Edgehog's database.

A `justfile` is introduced to easily run tasks using the [Just](https://github.com/casey/just) task, like connecting a simulated device by cloning, compiling and running the [edgehog-device-runtime](https://github.com/edgehog-device-manager/edgehog-device-runtime).


lusergit (approved):
:rocket:

PR #926, opened 2025-09-15
feat(containers): UI to support mounting device files on containers

Emina-M:
From Edgehog's UI, when creating a release, it is possible to optionally specify device files that need to be mounted on a container. The files are displayed under `Device mappings` section.

<details>
<summary>
Screenshots
</summary>

Screenshot from 2025-09-15 09-36-21

Screenshot from 2025-09-15 09-37-10

Screenshot from 2025-09-15 09-36-45

</details>

### DIFF
--- a/frontend/src/forms/CreateRelease.tsx
+++ b/frontend/src/forms/CreateRelease.tsx
@@ -30,6 +30,7 @@ import type { CreateRelease_NetworksOptionsFragment$key } from "api/__generated_
 import type { CreateRelease_VolumesOptionsFragment$key } from "api/__generated__/CreateRelease_VolumesOptionsFragment.graphql";
 import type { CreateRelease_SystemModelsOptionsFragment$key } from "api/__generated__/CreateRelease_SystemModelsOptionsFragment.graphql";
 import {
+  ContainerCreateWithNestedDeviceMappingsInput,
   ContainerCreateWithNestedNetworksInput,
   ContainerCreateWithNestedVolumesInput,
   ReleaseCreateRequiredSystemModelsInput,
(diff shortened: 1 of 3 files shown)


frontend/src/forms/CreateRelease.tsx
```diff
@@ -1402,6 +1453,130 @@ const ContainerForm = ({
...
+                    })
+                  }
+                  disabled={!canAddDeviceMapping}
+                >
+                  <FormattedMessage
+                    id="forms.CreateRelease.addDeviceMappingButton"
+                    defaultMessage="Add Device Mapping"
+                  />
+                </Button>
+              </div>
```
OmarBrbutovic, line 1575: **Suggestion for a later improvement:**
When adding new device mappings, it might be cleaner if the input field labels (e.g., *Path in container*, *Path on host*, *Container group permissions*) were shown only once at the top. The “Add” button could then insert new rows with just the input fields, without duplicating the labels.

Other components, such as *Networks*, *Volumes*, and similar sections, could also be refactored later to follow this unified approach.

OmarBrbutovic (approved):

frontend/src/forms/CreateRelease.tsx
davidebriani: I agree with Omar.

In addition, displaying the device mappings on the release review page takes a lot of space.
Maybe we can find a leaner way to list the existing mappings. Or, if we want to keep a tabular view, it could have a similar visualization to the one on the release creation page, so that the user is not confused by the UI differences and quickly see the table of device mappings he defined earlier. In that case, we could evaluate a single DeviceMappingsTable used in both pages (but readonly in the release review page).

As small improvements on the current PR, I think we could:
- use the whole "Path on Host" value as the label for the mapping, if we really need a label.
- align the Trash icons horizontally with the input fields. Right now they look off to me.

Anyway, the PR is in a good enough shape to merge it and secure your work @Emina-M.
I propose to continue improving it with additional PRs.

davidebriani (approved):

PR #942, opened 2025-09-22
chore: relate deployment resources on creation

lusergit:
This commit separates concerns of `create` and `send` actions in deployments. `create` creates the resource, realting all underlying resources, while `send` sends the description to the device. SInce the second one is idempotent it can be repeated any number of times.

This also accidentally fixes a bug: `fetch_image_deployment` did not work before (probably because of a race condition). Now it does

### DIFF
--- a/.github/workflows/backend-test.yaml
+++ b/.github/workflows/backend-test.yaml
@@ -200,25 +200,34 @@ jobs:
           --health-interval 10s
           --health-timeout 5s
           --health-retries 5
-      minio:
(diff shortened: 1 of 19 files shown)


Dam-99 (approved):
lgtm! just a question: what are the "resources" mentioned in the description for the `send_deployment` update actions for image, volume, network and device_mapping?

OmarBrbutovic (approved):

davidebriani (approved):

PR #953, opened 2025-09-24
chore!: aggregate readiness

lusergit:
Based on #952 

Deployment events are treated as events, rather than something that should update the state of the deployment.

The UI can log these and leave to the user the choice on what to do.

BREAKING CHANGE: `error` states no longer exist, instead the `Deployment` resource exposes an `events` relationship with a list of events, sortable per creation


davidebriani (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+## Changed
```
davidebriani:
```suggestion
### Changed 
```

davidebriani (approved):

PR #955, opened 2025-09-25
chore: handle deployments deletion

lusergit:
The device publishes a `nil` value on `Available*` interfaces when removing a resource. This should trigger the associated `.Deployment` resource's deletion: fixed.

### DIFF
--- a/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
+++ b/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
@@ -120,10 +120,10 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
         image_deployment =
           Containers.fetch_image_deployment!(image_id, device.id, tenant: tenant)
 
-        if event.value do
-          Containers.mark_image_deployment_as_pulled!(image_deployment, tenant: tenant)
-        else
-          Containers.mark_image_deployment_as_unpulled!(image_deployment, tenant: tenant)
+        case event.value do
+          true -> Containers.mark_image_deployment_as_pulled!(image_deployment, tenant: tenant)
+          false -> Containers.mark_image_deployment_as_unpulled!(image_deployment, tenant: tenant)
+          nil -> Containers.destroy_image_deployment!(image_deployment, tenant: tenant)
         end
 
         containers = Containers.containers_with_image!(image_id, tenant: tenant)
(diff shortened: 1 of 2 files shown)


backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -154,9 +154,16 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
+        case event.value do
+          true ->
+            Containers.mark_network_deployment_as_available(network_deployment, tenant: tenant)
+
+          false ->
+            Containers.mark_network_deployment_as_unavailable(network_deployment, tenant: tenant)
+
+          nil ->
+            Containers.destroy_network_deployment!(network_deployment, tenant: tenant)
+        end
```
davidebriani, line 166: Great, now thanks to `Containers.destroy_network_deployment!` we should see when the backend fails to purge the network deployment.

However, in the first two cases, errors are silently ignored.
It would be better to use the `!` versions of the Ash-generated functions.

The same is true for the changes on the similar events for the other resources (volumes and device mappings).

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -349,16 +377,17 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
               Containers.mark_deployment_as_started(deployment, tenant: tenant)
               Containers.deployment_update_resources_state(deployment, tenant: tenant)
 
             "Stopped" ->
               Containers.mark_deployment_as_stopped(deployment, tenant: tenant)
               Containers.deployment_update_resources_state(deployment, tenant: tenant)
+
+            # The device is removing the deployment, remove it!
+            nil ->
+              Containers.destroy_deployment!(deployment, tenant: tenant)
```
davidebriani, line 390: Contrary to the previous comment, here the the value of the function calls are directly used as the return value of the event handler.
Since the `handle_event` function should return an `{:ok, value}` tuple, we should respect that here and avoid using the `!` function variant as the last function call. Otherwise, we produce a match exception within the `run/3` function that will return a 500 error to the API client.

```suggestion
          case state do
            "Started" ->
              Containers.mark_deployment_as_started!(deployment, tenant: tenant)
              |> Containers.deployment_update_resources_state(tenant: tenant)

            "Stopped" ->
              Containers.mark_deployment_as_stopped!(deployment, tenant: tenant)
              |> Containers.deployment_update_resources_state(tenant: tenant)

            # The device is removing the deployment, remove it!
            nil ->
              Containers.destroy_deployment(deployment, tenant: tenant)
          end
```

davidebriani (changes requested):

PR #956, opened 2025-09-25
feat: expose underlying resources to the user

lusergit:
Based on #959 

Two changes are introduced:
The first one is about exposing to the API the deployment resources. Users can with this new exposure better understand the current state of the deployment, without relying on a misterious `resources_state` attribute that has to be updated every time.

The second one is about tests:
Making a deployment fixture does not really look at the release it should
deploy. Instead it just creates a `Deployment` resource with the correct release.
This might however be a curse when trying to look at related resources.

`deployment_fixture` should create a valid deployment for the release,
which includes relationships.

### DIFF
--- a/backend/lib/edgehog/containers/deployment.ex
+++ b/backend/lib/edgehog/containers/deployment.ex
@@ -38,6 +38,8 @@ defmodule Edgehog.Containers.Deployment do
 
   graphql do
     type :deployment
+
+    paginate_relationship_with container_deployments: :relay
   end
 
   actions do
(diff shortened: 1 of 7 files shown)


backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -195,7 +195,10 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
     case String.split(event.path, "/") do
       ["", volume_id, "created"] ->
         volume_deployment =
-          Containers.fetch_volume_deployment!(volume_id, device.id, tenant: tenant)
+          Containers.fetch_volume_deployment!(volume_id, device.id,
+            tenant: tenant,
+            load: [container_deployments: :deployments]
+          )
```
Dam-99, line 201: Why is this needed, and why only on volume deployments?

OmarBrbutovic (approved):

Dam-99 (approved):

davidebriani (approved):

PR #961, opened 2025-09-26
fix: ignore prefetch_count in comparison of delivery policies

davidebriani:
The current comparison did not properly make a reasonable comparison.


lusergit (approved):
:rocket:

PR #985, opened 2025-10-14
Update versions of elxir and erlang

osmanhadzic:
Updated Elixir version from 1.17.2 → 1.18.4 in .tool-versions
Updated Erlang version from 27.0 → 28.1 in .tool-versions
Refactor function `handle_cleanup` 
Add migration to include UUID v7 functions
Updated corresponding versions in the Dockerfile
Removed unused functions from repo.ex:
`transact/1`
`fetch/2`
`fetch_by/2`

### DIFF
--- a/frontend/src/components/AddAvailableApplications.tsx
+++ b/frontend/src/components/AddAvailableApplications.tsx
@@ -29,7 +29,7 @@ import type { AddAvailableApplications_DeployRelease_Mutation } from "api/__gene
 
 const GET_APPLICATIONS_WITH_RELEASES_QUERY = graphql`
   query AddAvailableApplications_GetApplicationsWithReleases_Query(
-    $filter: ApplicationFilterInput
+    $filter: ApplicationFilterInput = {}
   ) {
     applications(first: 10000, filter: $filter) {
       edges {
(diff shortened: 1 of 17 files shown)


lusergit (commented):
Huge work @osmanhadzic ! There are some errors in tests right now, but I think this is solvable 👍 

Basically what happens is that at tenant cleanup some resources should be handled differently than other. Ash introduced [Prevention on concurrent writes](https://hexdocs.pm/ash/prevent-concurrent-writes.html) which interferes (I think actually for the good) in our cleanup process. I see the following solution for this problem:

- in `multitenant_resource.ex` (L52-L56) macro, add a new default behavior for deletion
  ```elixir
      if unquote(Keyword.get(opts, :auto_cleanup, true)) do
        references do
          reference :tenant, on_delete: :delete
        end
      end
  ```
This is the current behavior, but we want to disable it fro three resources that require special handling in deletion: `SystemModel`s `BaseImage`s and managed `OTAOperation`s. In these resources change the `MultitenantResource` macro to disable the automatic cleanup:
```elixir
# Example: in base image
defmodule Edgehog.BaseImages.BaseImage do
  @moduledoc false
  use Edgehog.MultitenantResource,
    domain: Edgehog.BaseImages,
    extensions: [
      AshGraphql.Resource
    ],
	auto_cleanup: false # <- Add this one !

```
and at the same time specify to ignore deletion when deleting a tenant
```elixir
  postgres do
	  # ...

      references do
        reference :tenant, on_delete: :nilify # <- this sets the foreign key to nil when deleting a tenant 
      end
  end
``` 

I think in the `HandleCleanup` change for the tenant these resources might still be stale in the `after_transaction` handling, but now you can retrieve them by their ID and handle the deletion as before.

lusergit (changes requested):

lusergit (approved):

PR #993, opened 2025-10-17
fix: reconcile backend snapshots

ArnelaL:
Ensures snapshots are created sequentially (each based on the previous one) instead of multiple snapshots being derived from the same parent, which caused the penultimate snapshot’s changes to be overwritten.


lusergit (approved):
:rock:

PR #1004, opened 2025-10-21
feat: add validation for deployment campaign operation type requirements

OmarBrbutovic:
Add validation to ensure that deployment campaigns are created with the required arguments based on their operation type. Specifically, upgrade operations now require a `target_release_id` that references a valid, compatible release.


lusergit (commented):
lgtm 👍

Dam-99 (approved):

PR #1005, opened 2025-10-21
feat: Add default image for devices without system model image

JasminaPiric:
Devices without a system model image now show a default image instead of a grey placeholder, improving user visual experience
Old:
image

New:
image


OmarBrbutovic (approved):

lusergit (approved):
:rocket:

PR #1008, opened 2025-10-22
feat: Add container binds

OmarBrbutovic:
Adds attributes to the containers, allowing for the specification of `binds`, along with necessary validations. Each bind must follow the format `host-path:container-path[:options]`.


lusergit (commented):
Overall lgtm 🔥 

One small nit based on the interface description

PR #1011, opened 2025-10-23
refactor: Add modular operation type support to deployment campaigns 

OmarBrbutovic:
Refactor the lazy deployment executor to support multiple operation types (deploy, upgrade, start, stop, delete) through a modular architecture.

This enables independent development of operation logic in parallel, while keeping the core state machine stable. Currently, only the `deploy` operation is fully implemented, other types include placeholder handlers to facilitate future extensions.

Depends on #1004

### DIFF
--- a/backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/executor.ex
+++ b/backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/executor.ex
@@ -115,7 +115,7 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Executor do
       tenant_id
       |> Core.get_deployment_campaign!(deployment_campaign_id)
       |> Ash.load!(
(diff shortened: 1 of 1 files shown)


backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/executor.ex
```diff
@@ -107,12 +115,15 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Executor do
       tenant_id
       |> Core.get_deployment_campaign!(deployment_campaign_id)
       |> Ash.load!(
-        [release: [containers: [:networks, :volumes, :image]]],
+        [:target_release, release: [containers: [:networks, :volumes, :image]]],
```
lusergit:
```suggestion
        [target_release: [], release: [containers: [:networks, :volumes, :image]]],
```
(ash would otherwise complain at runtime)

lusergit (changes requested):
overall lgtm 👍

PR #1013, opened 2025-10-24
feat: remove device upon deletion finished in astarte

lusergit:
Based on #1007 

This feature allows the astarte-edgehog binomy to handle devices seamlessly: once devices are removed in astarte they are also removed from edgehog.

### DIFF
--- a/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
+++ b/backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
@@ -29,11 +29,13 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
   alias Edgehog.Devices.Device
   alias Edgehog.OSManagement
   alias Edgehog.Triggers.DeviceConnected
+  alias Edgehog.Triggers.DeviceDeletionFinished
   alias Edgehog.Triggers.DeviceDisconnected
   alias Edgehog.Triggers.DeviceRegistered
   alias Edgehog.Triggers.IncomingData
   alias Edgehog.Triggers.TriggerPayload
 
+  require Ash.Query
   require Logger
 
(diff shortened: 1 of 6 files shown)


ArnelaL (approved):

backend/lib/edgehog/triggers/handler/manual_actions/handle_trigger.ex
```diff
@@ -87,6 +89,16 @@ defmodule Edgehog.Triggers.Handler.ManualActions.HandleTrigger do
...
   end
 
+  defp handle_event(%DeviceDeleted{}, tenant, realm_id, device_id, _timestamp) do
+    device_query = Ash.Query.filter(Device, device_id == ^device_id and realm_id == ^realm_id)
+
+    # here uniqueness is guaranteed by :unique_realm_device_id identity on
+    # Devices
+    with {:ok, device} <- Ash.read_one(device_query, tenant: tenant) do
+      Ash.destroy(device)
+    end
```
davidebriani: `Ash.read_one` can return `{:ok, nil}` by default.
We can use the [:not_found_error?](https://hexdocs.pm/ash/Ash.html#read_one/2-options) option to avoid that.

```elixir
    with {:ok, device} <- Ash.read_one(device_query, tenant: tenant, not_found_error?: true) do
      Ash.destroy(device)
    end
```

backend/priv/astarte_resources/trigger_templates/edgehog-deletion.json.eex
```diff
@@ -0,0 +1,18 @@
+{
+    "name": "edgehog-deletion",
```
davidebriani: Here too we could be more specific: `edgehog-device-deletion-finished`, or `edgehog-deletion-finished`

davidebriani (changes requested):

PR #1016, opened 2025-10-24
chore: garbage collect dangling resources

lusergit:
`Deployment` resources can be dangling. This happens when a device nilifies one of the `Available*` interfaces without nilifying the other resources. This commit introduces `destroy_if_dangling` and `danlging?` deletion action and calculation (respectively), to trigger a deletion when a resource is dangling.


OmarBrbutovic (changes requested):
Consider adding logging where child deletion is attempted and when it fails for visibility.

PR #1029, opened 2025-10-31
feat: add upgrade operation support to deployment campaigns

ArnelaL:
Implement support for the `upgrade` operation type in deployment campaigns. The `upgrade` operation checks if a deployment we want to upgrade exists and is in an appropriate state before upgrading it, with proper error handling for cases like missing deployments or deployments in transitional states.


lusergit (approved):
👍

PR #1031, opened 2025-10-31
feat: add `delete` operation support to deployment campaigns

Dam-99:
Implement support fo the start operation type in deployment campaigns.
The `delete` operation checks if a deployment exists and is in an appropriate
state before deleting it.


backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/core.ex
```diff
@@ -373,6 +373,25 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Core do
...
+    - The deployment target struct associated with the device ID and campaign ID.
+  """
+  def get_target_for_deployment_by_device!(tenant_id, deployment_campaign_id, device_id) do
+    DeploymentCampaigns.fetch_target_by_device_and_campaign!(
+      device_id,
+      deployment_campaign_id,
+      tenant: tenant_id
+    )
+  end
+
```
OmarBrbutovic: here as well, it can replace the `get_target_for_deployment`

OmarBrbutovic (changes requested):

backend/lib/edgehog/deployment_campaigns/deployment_mechanism/lazy/core.ex
```diff
@@ -668,6 +666,49 @@ defmodule Edgehog.DeploymentCampaigns.DeploymentMechanism.Lazy.Core do
...
+    - release: The release struct referenced in the deployment to be deployed.
+    - mechanism: The mechanism settings.
+
+  ## Returns
+    - `{:ok, target}` if the delete command is successfully sent.
+    - `{:error, :deployment_transitioning}` if the deployment is in a transitional state.
+    - `{:error, :deployement_not_found}`, if the deployment doesn't exist n the target.
+    - `{:error, reason}` for any other errors.
+  """
+  def delete(target, release, _mechanism) do
```
OmarBrbutovic: There’s no need to pass `mechanism` as a prop here since it’s not used, and there’s no other delete function with the same arity that would cause ambiguity.

OmarBrbutovic (changes requested):

PR #1035, opened 2025-11-03
ci: compile first, test later

lusergit:
We used to run tests and linting independently. It was sound, but it meant that the CI re-compiled the code (and dependencies) for each job (which meant ~10jobs per PR).

## Alembic CI workflows
Taking inspiration from Ash's CIs themselves, we can use https://github.com/team-alembic/staple-actions. These workflow allow for parallel working and take care of handling cache themselves

## New CI workflows
- `Build DEV` and `Build TEST` separated, the two environments are needed for different jobs
- `Dialyzer`, `Credo` and `Code formatting` as separated jobs. they all depend on compilation and run after that
- `Test` matrix and `Coverage` separated. `Coverage` is its own job running as before with postgres 18. both depend on test compilation
- `Integration Test*` didn't change much, only compilation is demanded to the test build

## Benefits
Devs can focus on what's wrong since dyalizer, credo and format are on different jobs. Testing still fails if one of them fails, but we have more control on coverage reports.

### DIFF
--- a/.github/workflows/backend-test.yaml
+++ b/.github/workflows/backend-test.yaml
@@ -24,18 +24,18 @@ on:
   # Run when pushing to stable branches
   push:
     paths:
-    - 'backend/**'
-    - '.tool-versions'
-    - '.github/workflows/backend-test.yaml'
+      - "backend/**"
+      - ".tool-versions"
+      - ".github/workflows/backend-test.yaml"
     branches:
-    - 'main'
-    - 'release-*'
+      - "main"
+      - "release-*"
   # Run on pull requests matching apps
   pull_request:
     paths:
-    - 'backend/**'
-    - '.tool-versions'
-    - '.github/workflows/backend-test.yaml'
+      - "backend/**"
+      - ".tool-versions"
+      - ".github/workflows/backend-test.yaml"
 
 env:
   MIX_ENV: test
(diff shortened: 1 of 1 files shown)


.github/workflows/backend-test.yaml
```diff
@@ -43,93 +43,73 @@ env:
...
+          working-directory: backend
+          mix-env: test
 
-    - name: Run dialyzer
-      run: mix dialyzer --format github
+  dialyzer:
+    name: Dialyzer
+    runs-on: ubuntu-latest
+    needs:
+      - build-dev
```
OmarBrbutovic: I'm not sure how it works in CI, but locally we usually run Dialyzer with the test environment. Using dev normally passes, while test can surface additional errors.

.github/workflows/backend-test.yaml
```diff
@@ -43,93 +43,73 @@ env:
...
+          mix-env: test
+          args: "--warnings-as-errors"
 
-    # Create PLTs if no cache was found
-    - name: Create PLTs
-      if: steps.plt_cache.outputs.cache-hit != 'true'
-      run: mix dialyzer --plt
+  build-dev:
+    name: Build DEV
+    runs-on: ubuntu-22.04
```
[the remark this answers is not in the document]
lusergit: `ubuntu-latest` should be the one we use. I forgot to replace it everywhere

.github/workflows/backend-test.yaml
lusergit: On a second thought using the test environment everywhere is either sound or better. I think the `build-dev` step can be avoided and run all checks in `build-test`

PR #1039, opened 2025-11-07
Refactor: centralize load constants and create generic FormRow component 

JasminaPiric:
- Created common constants RECORDS_TO_LOAD_FIRST and RECORDS_TO_LOAD_NEXT and replaced entity specific load variables with them across the app
- Introduced a reusable FormRow component shared across multiple files


lusergit (approved):
took a while to review 😅  but makes sense 👍

PR #1041, opened 2025-11-10
feat: allow users to re-send deployment messages

lusergit:
Allows users to re-send deployment messages to the device. A new button triggers
a `send_deployment` action in the backend, re-sending the necessary messages to
the device.

This is accompanied by a validation in the `send_deployment` action, allowing it to be run only when the device is not ready.

Screenshot From 2025-11-13 11-28-50

### DIFF
--- a/frontend/src/components/DeployedApplicationsTable.tsx
+++ b/frontend/src/components/DeployedApplicationsTable.tsx
@@ -27,6 +27,7 @@ import Select, { SingleValue } from "react-select";
 import type { DeployedApplicationsTable_PaginationQuery } from "api/__generated__/DeployedApplicationsTable_PaginationQuery.graphql";
 import type { DeployedApplicationsTable_deployedApplications$key } from "api/__generated__/DeployedApplicationsTable_deployedApplications.graphql";
 
+import type { DeployedApplicationsTable_sendDeployment_Mutation } from "api/__generated__/DeployedApplicationsTable_sendDeployment_Mutation.graphql";
 import type { DeployedApplicationsTable_startDeployment_Mutation } from "api/__generated__/DeployedApplicationsTable_startDeployment_Mutation.graphql";
 import type { DeployedApplicationsTable_stopDeployment_Mutation } from "api/__generated__/DeployedApplicationsTable_stopDeployment_Mutation.graphql";
 import type { DeployedApplicationsTable_deleteDeployment_Mutation } from "api/__generated__/DeployedApplicationsTable_deleteDeployment_Mutation.graphql";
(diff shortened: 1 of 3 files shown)


frontend/src/components/DeployedApplicationsTable.tsx
```diff
@@ -622,6 +679,21 @@ const DeployedApplicationsTable = ({
...
+              id: "components.DeployedApplicationsTable.sendButtonTitle",
+              defaultMessage: "Re-send the release to the device",
+            })}
+            onClick={() => handleSendDeployedApplication(getValue().id)}
+          >
+            <Icon
+              className={getValue().isReady ? "text-muted" : "text-success"}
+              icon={"rotate"}
+            />
+          </Button>
```
OmarBrbutovic: Instead of adding a new button that would be visible to users all the time, but remain disabled 99% of the time when everything is working as expected, we could extend the existing `ActionButtons` component to dynamically show the **Re-deploy** button only when the deployment is not ready.

In that case, it would replace the existing action buttons, which wouldn’t be usable anyway while the deployment is still in progress.

Here’s an example implementation:

```ts
// Action buttons with play, stop, and redeploy icons
const ActionButtons = ({
  intl,
  state,
  onStart,
  onStop,
  onRedeploy,
  isReady,
}: {
  intl: ReturnType<typeof useIntl>;
  state: DeploymentState;
  onStart: () => void;
  onStop: () => void;
  onRedeploy: () => void;
  isReady: boolean | null;
}) => (
  <div className="d-flex align-items-center">
    {!isReady ? (
      <Button
        className="btn p-0 border-0 bg-transparent icon-click"
        title={intl.formatMessage({
          id: "components.DeployedApplicationsTable.sendButtonTitle",
          defaultMessage: "Redeploy Application",
        })}
        onClick={onRedeploy}
      >
        <Icon className="text-dark" icon="rotate" />
      </Button>
    ) : state === "STOPPED" || state === "ERROR" ? (
      <Button
        onClick={onStart}
        className="btn p-0 text-success border-0 bg-transparent icon-click"
        title={intl.formatMessage({
          id: "components.DeployedApplicationsTable.startButtonTitle",
          defaultMessage: "Start Deployment",
        })}
      >
        <Icon icon="play" className="text-success" />
      </Button>
    ) : state === "STARTED" ? (
      <Button
        onClick={onStop}
        className="btn p-0 text-danger border-0 bg-transparent icon-click"
        title={intl.formatMessage({
          id: "components.DeployedApplicationsTable.stopButtonTitle",
          defaultMessage: "Stop Deployment",
        })}
      >
        <Icon icon="stop" className="text-danger" />
      </Button>
    ) : (
      <Button className="btn p-0 border-0 bg-transparent" disabled>
        <Icon
          icon={state === "STARTING" || state === "DEPLOYING" ? "play" : "stop"}
          className="text-muted"
        />
      </Button>
    )}
  </div>
);
```

One small consideration: right after a deployment, the `isReady` flag will still be `false`, which could lead to users accidentally triggering a re-deploy too soon.
To mitigate that, we could add a short **30-second cooldown timer** that temporarily disables the button immediately after deployment.

OmarBrbutovic (approved):

PR #1044, opened 2025-11-12
ci: expand docs workflow triggers to include admin API and Ash resources

davidebriani:
The docs workflow now monitors additional backend paths to ensure documentation is regenerated when the GraphQL schema changes due to modifications in:
- Admin API modules that define GraphQL types and operations
- Ash resources and domains that are exposed through the schema

This prevents outdated API documentation when backend changes are made outside the previously monitored schema directory.

Also updates copyright year to 2025.


lusergit (approved):
🪨

PR #1052, opened 2025-11-17
fix: Use Node.js 22 for SpectaQL to avoid compatibility issues

OmarBrbutovic:
SpectaQL 3.0.5 has compatibility issues with Node.js 24, resulting in "arrangeData is not a function" errors. This change temporarily switches to Node.js 22 specifically for the SpectaQL documentation generation steps, while keeping Node.js 24 for the rest of the workflow.


lusergit (approved):
:droplet:

PR #1058, opened 2025-11-19
feat: Prevent deployment actions during conflicting campaigns

OmarBrbutovic:
Add validation to block deployment actions `start`, `stop`, `delete`, `upgrade`, when the deployment is part of an in-progress campaign with a different operation type


lusergit (approved):
:+1:

ArnelaL (approved):

PR #1060, opened 2025-11-20
feat: add deployment details

ArnelaL:
- Implemented DeploymentDetails component that displays detailed information about resource deployment status and events reported by the device
- Added an action button to the Action column that redirects to a page showing deployment details

https://github.com/user-attachments/assets/013b2c0b-6da1-4334-8843-017cde173c3b


lusergit (commented):
:+1:

PR #1063, opened 2025-11-21
chore: just develop

lusergit:
Since we provide `just` tasks to setup prod-like test environment it would have been useful to also provide tasks to provision a development environment. This change introduces a handful of new tasks:
- `just provision-realm` provisions an astarte realm
- `just dev-backend` sets up a development environment for the backend. Setting up an astarte instance with a realm and simultaneously initiating an edgehog tenant. This task does not seed the database as in a development environment the setup of the backend state is left to the developer. A `iex -S mix phx.server` shell is spawned which allows devs to inspect in real time state of the backend. It also allows for hot code reloading
- `just dev-frontend` sets up a development environment for the frontend, opening a `npm run start` shell which allowshot code reloading and real-time compilation
- dashboard opening is also split and can be triggered separately: `just open-astarte-dashboard` and `just open-edgehog-dashboard`

### DIFF
--- a/backend/config/dev.exs
+++ b/backend/config/dev.exs
@@ -20,6 +20,10 @@
 
 import Config
 
+url_host = System.get_env("URL_HOST", "localhost")
+url_port = System.get_env("URL_PORT", "4000")
+url_scheme = System.get_env("URL_SCHEME", "http")
+
 config :azurex, Azurex.Blob.Config,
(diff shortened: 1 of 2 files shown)


Dam-99 (approved):
:rocket:

Dam-99 (approved):
yesyesyes

PR #1065, opened 2025-11-21
chore: e2e test edgehog -- Gleam

lusergit:
# E2E test edgehog -- [Gleam]
The following PR adds a new e2e tool to test edgehog. Starting a painful, yet important task: ensuring things work without human testing.

## Integration in the CI
Ideally this tool can be compiled once and run in CI at each backend change. It could test a specific path in the backend in an end to end fashion: it makes a request, expects some side effects and a specific response from the server.

The tool runs each test in parallel, aggregating the result of each test and prompting the user what went wrong where.

## Different integration with graphql
libraries for a gleam graphql client are not there yet. Integrating with an existing schema is a bit painful. Instead gleam could have other advantages over rust:

- better integration in the BEAM: we could poke into the BEAM runtime and verify edgehog properties while testing, This however doesn't seem useful for an end to end test.
- ?

### DIFF
--- a/.github/workflows/e2e-test.yaml
+++ b/.github/workflows/e2e-test.yaml
@@ -1,0 +1,79 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2025 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


lusergit:
Superseded by #1064

PR #1084, opened 2025-12-01
feat: collapsible headers in container form

Dam-99:
Allows to collapse singular sections for each container, and the container
itself, **when** creating a new application release that includes containers.

https://github.com/user-attachments/assets/36c71529-5f7c-4d51-90d8-b96be317aaff

### DIFF
--- a/frontend/src/forms/ContainerForm.scss
+++ b/frontend/src/forms/ContainerForm.scss
@@ -1,0 +1,64 @@
+/*
+  This file is part of Edgehog.
+
+  Copyright 2025 SECO Mind Srl
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


frontend/src/forms/CreateRelease.tsx
```diff
@@ -715,25 +738,70 @@ const CreateRelease = ({
...
+                      index={index}
+                      register={register}
+                      errors={errors}
+                      remove={handleRemoveContainer}
+                      imageCredentials={imageCredentialsOptions}
+                      networks={networkOptions}
+                      volumes={volumeOptions}
+                      control={control}
+                      isImported={isContainerImported(index)}
+                      isModified={
```
ArnelaL: These two aren't used anymore in the ContainerForm, so we should remove them from ContainerFormProps, and not pass them anymore

Dam-99: Ops, ctrl+z'd a once to many times

frontend/src/forms/ContainerForm.scss
```diff
@@ -0,0 +1,61 @@
...
+        content: "";
+        flex: 1;
+        height: $border-width;
+        margin-left: 1rem;
+        margin-right: 1rem;
+        background-color: $border-color;
+      }
+    }
+  }
+}
```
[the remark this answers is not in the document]
Dam-99: I think this could work, I'll try it out

ArnelaL (changes requested):

PR #1199, opened 2026-01-28
feat: run devices reconciliation at tenant provisioning

lusergit:
## feat: devices reconciliation

Expanding on #1164 , this change runs reconciliations with astarte in a separate task, allowing it to fail without dragging the application tree down with it if something fails. This also makes uses of the `stream` functionality of the astarte client introduced in https://github.com/astarte-platform/astarte-client-elixir/pull/34, making use of pagination to query the full list of devices, which might be very big.

## Checklist


ArnelaL (approved):

OmarBrbutovic (approved):
Nit: Since these resources are now no longer in use, they can be safely removed:

```elixir
define :fetch_realms, action: :read
```

and 

```elixir
define :fetch_all, action: :read
```

Emina-M (approved):

PR #1215, opened 2026-02-03
ci: do not consider changelogs in typos and format

lusergit:
## Make `typos`, `dprint` and `prettier` ignore frontend changelog.

Frontend CI checked the CHANGELOG file, which is auto-generated from commits, including typos there. It also checked formatting of such file, which was wrongfully marked as unformatted.

## Checklist

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -43,3 +43,7 @@ PN = "PN"
 
 [default.extend-identifiers]
 OTPs = "OTPs"
+
+# Exclude changelogs: they're auto generated based on commits which are linted
+[files]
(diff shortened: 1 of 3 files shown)


.typos.toml
```diff
@@ -43,3 +43,7 @@ PN = "PN"
 
 [default.extend-identifiers]
 OTPs = "OTPs"
+
+# Exclude changelogs: they're auto generated based on commits which are linted
```
Dam-99: Technically no spell check and only linting, will look into other options for that

Dam-99 (approved):

OmarBrbutovic (approved):

PR #1255, opened 2026-02-27
refactor: rework file transfer to use presigned URLs

ArnelaL:
## Rework file transfer to use presigned URLs

- In the previous code, file uploads and downloads were handled by the backend. Now, the backend only generates presigned URLs for uploads and downloads, and the actual file transfer will be done directly between the client and the storage

## Checklist


Dam-99 (approved):

backend/lib/edgehog/files/file_download_request/manual_actions/create_presigned_url.ex
```diff
@@ -18,33 +18,21 @@
...
-
-  def storage_dir(_version, {_file, scope}) do
-    %{tenant_id: tenant_id, repository_id: repository_id} = scope
-
-    "uploads/tenants/#{tenant_id}/repositories/#{repository_id}/files"
-  end
-
-  def filename(_version, {_file, scope}) do
-    scope.file_name
+    Edgehog.Storage.create_presigned_urls(file_path)
```
lusergit: Same for the read, this can be a calculation too

lusergit (changes requested):
Overall huge work 😭 

I have a couple of observations feel ofc free to add your opinion on this

lusergit (approved):
:+1:

PR #1296, opened 2026-03-19
refactor: Optimize and improve validation on Files domain

ArnelaL:
## refactor: Optimize and improve validation on Files domain
- removed unused fields on File Resource
- added default values where needed
- improved repository fields naming
- removed unused `read_file_download_request_presigned_url` mutation
- added validation to :create_presigned_url on File resource to check whether repository with passed repository_id exists
- Added new create action for File Download Request called :managed so we can distinguish between requests created using ephemeral files and using existing files from repositories. This is useful for handling deletion of ephemeral files from storage upon request completion, so that when a manged file download request is made the managed data does not get deleted.
- Handle File resource deletion on tenant removal
- Use different mutations when creating file download request using direct upload and file from repository
- Rename fields to match the updated backend implementation
## Checklist


Dam-99 (approved):

OmarBrbutovic (approved):

backend/lib/edgehog/files/file/file.ex
```diff
@@ -98,32 +101,6 @@ defmodule Edgehog.Files.File do
...
-      public? true
-      # root
-      default 0
-    end
-
-    attribute :url, :string do
-      description "Full URL where file is stored. Set by the uploader."
-      public? true
-    end
-
```
lusergit: I think #1299 is actually using these 😳

lusergit (approved):
A couple minor nits.

It's a big PR, for me we can do them later 👍

PR #1477, opened 2026-05-29
chore(backend): auto-write FGA tuples on actions

Dam-99:
Add changes for updating the FGA store when creating or destroying a resource.

The `alias` can be used for querying tuples in a more human-friendly way for
resources that do not use their primary id in FGA, and the changes create and
delete the tuples defining the alias.

The `owner` changes define the user who performed the create action as its
owner, and remove the related tuple from the FGA service on destroy.

Create transformers for the `Ash.FGA` extension that add the `alias` and `owner`
changes on the appropriate actions of resources where FGA is applied.

#### Additional documentation e.g. usage docs, diagrams, reviewer notes, etc.:

---

<details>
<summary><i>Thanks for sending a pull request! If this is your first time, here are some tips for you:</i></summary>

1. You can take a look at our [developer guide] for an introduction on Edgehog development!
2. Make sure to read [CONTRIBUTING.md] and [CODE_OF_CONDUCT.md]
3. If the PR is unfinished or you're actively working on it, mark it as draft

When fixing existing issues, use [github's syntax to link your pull request] to it

> `fixes #<issue number>`

We also have a syntax to signal dependencies to other open pull requests

> `depends on #<pr number>`
> `depends on https://github.com/...`

In case of stacked PRs, you may add the PR number in the last commit's title instead:

> ```mermaid
> gitGraph
>     commit id: "Current master"
>     branch feat1
>     checkout feat1
>     commit id: "feat: add something"
>     commit id: "feat: add something else (#100)"
>     branch feat2
>     checkout feat2
>     commit id: "refactor: do something"
>     commit id: "fix: solve issue"
>     commit id: "feat: add a feature (#101)"
>     branch feat3
>     checkout feat3
>     commit id: "feat: feat without pr number"
> ```

</details>

[github's syntax to link your pull request]: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests
[developer guide]: https://docs.edgehog.io/snapshot/edgehog_just_in_time.html
[CONTRIBUTING.md]: ../CONTRIBUTING.md
[CODE_OF_CONDUCT.md]: ../CODE_OF_CONDUCT.md

### DIFF
--- a/backend/lib/edgehog/auth/changes/erase_owner.ex
+++ b/backend/lib/edgehog/auth/changes/erase_owner.ex
@@ -1,0 +1,85 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+
+defmodule Edgehog.Auth.Changes.EraseOwner do
+  @moduledoc """
+  A change to remove the owner of a destroyed object in the FGA provider.
+
+  Opts:
+  - fga_type     :: atom. the type assigned to the resource in the FGA model
+
+  Example tuple:
+  ```fga
+    {user:bob, owner, device:Simpl3D3vice1d}
+  ```
+  """
+  use Ash.Resource.Change
+
+  alias Edgehog.Auth.FGAService
+
+  @needed_keys [:fga_type, :primary_key]
+
+  @impl Ash.Resource.Change
+  def init(opts) do
+    {_, errors} = Enum.reduce(@needed_keys, {opts, []}, &error_if_not_present/2)
+
+    if Enum.empty?(errors),
+      do: {:ok, opts},
+      else: {:error, errors}
+  end
+
+  @impl Ash.Resource.Change
+  def change(%{valid: false} = changeset, _opts, _context), do: changeset
+
+  @impl Ash.Resource.Change
+  def change(%{action_type: :destroy} = changeset, opts, _context) do
+    Ash.Changeset.after_action(changeset, &erase_owner(&1, &2, opts))
+  end
+
+  defp erase_owner(_changeset, record, opts) do
+    type = Keyword.get(opts, :fga_type)
+    primary_key = Keyword.get(opts, :primary_key)
+
+    id = Map.get(record, primary_key)
+    fga_id = "#{type}:#{id}"
+
+    with {:ok, user_list} <- FGAService.list_users(fga_id, "owner", "user") do
+      # There should only be one owner for a resource, and it shouldn't be a userset nor a wildcard.
+      # Handle None provider separately
+      user =
+        case user_list do
+          :all -> "user:anon"
+          [sub] -> "user:#{sub}"
+        end
+
+      with {:ok, _result} <- FGAService.delete(user, "owner", fga_id) do
+        {:ok, record}
+      end
+    end
+  end
+
+  defp error_if_not_present(key, {opts, errors}) do
+    error = "Missing key #{inspect(key)}."
+
+    if Keyword.has_key?(opts, key),
+      do: {opts, errors},
+      else: {opts, [error | errors]}
+  end
+end
--- a/backend/lib/edgehog/auth/changes/write_alias.ex
+++ b/backend/lib/edgehog/auth/changes/write_alias.ex
@@ -1,0 +1,136 @@
+#
+# This file is part of Edgehog.
+#
+# Copyright 2026 SECO Mind Srl
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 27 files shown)


backend/lib/edgehog/auth/changes/write_alias.ex
```diff
@@ -0,0 +1,138 @@
...
+    type = Keyword.get(opts, :fga_type)
+
+    id = Map.get(record, :id)
+    fga_id = "#{type}:#{id}"
+    alias = Map.get(record, alias_attr)
+    fga_alias = "#{type}:#{alias}"
+
+    if type == "device" do
+      dbg({fga_alias, "alias", fga_id})
+    end
```
ArnelaL: maybe a leftover?

backend/lib/edgehog/auth/changes/erase_owner.ex
```diff
@@ -0,0 +1,82 @@
...
+    id = Map.get(record, :id)
+    fga_id = "#{type}:#{id}"
+
+    # There should only be one owner for a resource.
+    # Handle None provider separately
+    user =
+      case FGAService.list_objects(fga_id, "owner", "user") do
+        {:ok, :all} -> "user:anon"
+        {:ok, [sub]} -> "user:#{sub}"
+      end
```
ArnelaL: Should we handle the cases where the list is empty or an error tuple is returned?

Dam-99: yesn't? an empty list would mean that the `write_owner` change didn't complete successfully, but if that's the case the creation of the object should've not happened, which sounds like a broken state between db and fga.
an error means something went wrong with the querying of fga, which should error out to invalidate the action and rollback the transaction, to avoid leaving an inconsistent state.
_that said..._ looking at how i wrote the code, this is not what happens (since the after_actions always return `:ok`), so i'll change it accordingly

backend/lib/edgehog/auth/providers/none.ex
```diff
@@ -56,6 +56,13 @@ defmodule Edgehog.Auth.Providers.None do
     {:ok, :all}
   end
 
+  @impl Behaviour
+  def list_users({subj, rel, type}, _context) do
+    Logger.debug("Filtering users for: {#{subj}, #{rel}, #{type}}")
```
lusergit:
```suggestion
    Logger.debug("Listing users for: {#{subj}, #{rel}, #{type}}")
```

backend/lib/edgehog/auth/providers/openfga.ex
```diff
@@ -154,4 +181,17 @@ defmodule Edgehog.Auth.Providers.OpenFGA do
...
+    case user do
+      {:object, %Openfga.V1.Object{type: type, id: id}} ->
+        "#{type}:#{id}"
+
+      {:userset, %Openfga.V1.UsersetUser{type: type, id: id, relation: rel}} ->
+        "#{type}:#{id}##{rel}"
+
+      {:wildcard, %Openfga.V1.TypedWildcard{type: type}} ->
+        "#{type}:*"
+    end
```
lusergit, line 193: Here we should just extract ids, we know from the request the actual type:

```suggestion
    case user do
      {:wildcard, %Openfga.V1.TypedWildcard{type: type}} -> :wildcard
      {_, other} -> Map.fetch!(other, :id)
    end
```

backend/lib/edgehog/auth/providers/openfga.ex
```diff
@@ -103,6 +103,33 @@ defmodule Edgehog.Auth.Providers.OpenFGA do
...
+      relation: rel,
+      object: %Openfga.V1.Object{
+        type: subj_type,
+        id: subj_id
+      }
+    }
+
+    with {:ok, %Openfga.V1.ListUsersResponse{users: users}} <-
+           Stub.list_users(channel, request) do
+      users = Enum.map(users, &extract_user_id/1)
```
lusergit: If you make the changes referenced here [`f7dd5ae` (this PR)](https://github.com/edgehog-device-manager/edgehog/pull/1477/changes/f7dd5aeffe30c9d1cc17cdf3a3f667446a69eb37#r3354554786)

maybe it's also useful to collapse wildcards to `{:ok, :all}`:

```suggestion
      users = users |> Enum.map(&extract_user_id/1) |> collapse_wildcard()

# ...

defp collapse_wildcard(users) do
  if :wildcard in users, do: :all, else: users
end
```

ArnelaL (approved):

backend/lib/edgehog/auth/changes/write_alias.ex
```diff
@@ -0,0 +1,136 @@
...
+        alias = Map.get(record, alias_attr)
+        fga_alias = "#{type}:#{alias}"
+        old_fga_alias = "#{type}:#{old_alias}"
+
+        if fga_alias != old_fga_alias do
+          FGAService.delete(old_fga_alias, "alias", fga_id)
+          FGAService.delete(fga_id, "alias", old_fga_alias)
+
+          FGAService.write(fga_alias, "alias", fga_id)
+          FGAService.write(fga_id, "alias", fga_alias)
```
lusergit, line 122: Nit, these are 4 calls when this could be done with 1 call.

maybe we can think of a bulk write

maybe in another PR

lusergit (commented):
Hi I have nitpicks 🤓 

And a comment (about bulk writes, no need to change it now but maybe a `TODO` can be added)

backend/lib/edgehog/auth/providers/openfga.ex
Dam-99: i think `userset` should still be handled separately, as it means "all objects in relation `relation` with object `<type>:<id>`": this means that they potentially are of different types than the one requested.
unfortunately the OpenFGA docs are not that clear on how the filtering works

lusergit: hmm ok makes sense

nonetheless we might want to colelct ids here? and not the whole `<type>:<id>` string

lusergit: AAAAaaaand you did 👍 

Approving

lusergit (approved):

PR #1589, opened 2026-08-12
refactor(backend): migrate resource depl. provisioners to macro

Dam-99:
Add a macro that implements a provisioner for a deployment resource,
alongside a macro that implements the `Core` module for a provisioner.

Migrate existing provisioners to use that macro.

This only includes the provisioners for the underlying resources.
Container deployments and full release deployments take a bit more effort and
should be migrated in the future.

#### What this PR does / why we need it:
Code duplication was very painful

#### Additional documentation e.g. usage docs, diagrams, reviewer notes, etc.:

---

<details>
<summary><i>Thanks for sending a pull request! If this is your first time, here are some tips for you:</i></summary>

1. You can take a look at our [developer guide] for an introduction on Edgehog development!
2. Make sure to read [CONTRIBUTING.md] and [CODE_OF_CONDUCT.md]
3. If the PR is unfinished or you're actively working on it, mark it as draft

When fixing existing issues, use [github's syntax to link your pull request] to it

> `fixes #<issue number>`

We also have a syntax to signal dependencies to other open pull requests

> `depends on #<pr number>`
> `depends on https://github.com/...`

In case of stacked PRs, you may add the PR number in the last commit's title instead:

> ```mermaid
> gitGraph
>     commit id: "Current master"
>     branch feat1
>     checkout feat1
>     commit id: "feat: add something"
>     commit id: "feat: add something else (#100)"
>     branch feat2
>     checkout feat2
>     commit id: "refactor: do something"
>     commit id: "fix: solve issue"
>     commit id: "feat: add a feature (#101)"
>     branch feat3
>     checkout feat3
>     commit id: "feat: feat without pr number"
> ```

</details>

[github's syntax to link your pull request]: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#about-linked-issues-and-pull-requests
[developer guide]: https://docs.edgehog.io/snapshot/edgehog_just_in_time.html
[CONTRIBUTING.md]: ../CONTRIBUTING.md
[CODE_OF_CONDUCT.md]: ../CODE_OF_CONDUCT.md


backend/lib/edgehog/containers/provisioner.ex
```diff
@@ -0,0 +1,448 @@
...
+      @type_atom unquote(type_atom)
+      @string_type unquote(string_type)
+      @type_string_camel unquote(type_string_camel)
+      @type_string_lower unquote(type_string_lower)
+      @deployment_string_lower unquote(deployment_string_lower)
+      @available_atom unquote(available_atom)
+
+      defmodule unquote(core_module) do
+        use Provisioner.Core, status_verb_key: unquote(status_verb_key)
+      end
```
lusergit: As discussed IRL let the user of the macro define the core module himself, even if it's just a 

```elixir
defmodule Core do
  use Edgehog.Provisioner.Core, ...
end
```

backend/lib/edgehog/containers/provisioner.ex
```diff
@@ -0,0 +1,448 @@
...
+          |> Keyword.put(unquote(deployment_atom), resource_deployment)
+          |> Keyword.put(:deployment, deployment)
+          |> Keyword.put(:tenant, tenant)
+
+        child_spec = Supervisor.child_spec({__MODULE__, args}, id: resource_deployment.id)
+
+        with {:error, {:already_started, pid}} <- DynamicSupervisor.start_child(@sup, child_spec) do
+          {:ok, pid}
+        end
+      end
```
lusergit: This is good ! But it's an architectural decision where to spawn the provisioner. 

This is fine as a default implementation, but I'd let the user define it otherwise. `defoverridable`?

backend/lib/edgehog/containers/provisioner.ex
```diff
@@ -0,0 +1,448 @@
...
+        problem with astarte
+      - an exponential backoff timeout is started
+      - A :timeout hits the server, it retries to send the #{unquote(type_string_lower)} information to the
+      device
+
+      TODOs, shortcomigs:
+      The logic to handle send succeeding but no message coming back from astarte is
+      not there yet
+      """
+
```
lusergit: As discussed IRL, this is sound with the code that was there, but I'd rather not copy this specific thing into the new code

- let's have the `@moduledoc` for the `Edgehog.Containers.Provisioner` macro, stating what it does and the possible keys
- in the single provisioners their `@moduledoc` should state something like
```
This is a provisioner. For more info refer to the `Edgehog.Containers.Provisioner` documentation.

This particular instance handles ... # <- let the users decide documentation on this matter
```

backend/lib/edgehog/containers/provisioner.ex
lusergit, line 1: A couple of nits! 

Overall it's sound and makes code happen in one place instead of 10.

My fear is that we might be generalizing a bit too much. Some things are internal policies (e.g., deployment module naming (with the `.Deployment` suffix)  and might be different in other resources.

For the time being I'd say this is ok, and we can adjust for different scenarios later 👍  (except for the docs nit and the core module 😳 )

lusergit (changes requested):

backend/lib/edgehog/containers/provisioner.ex
lusergit: my bad it's already overridable

backend/lib/edgehog/containers/provisioner.ex
ArnelaL: I agree with @lusergit , I don't like that the code relies so heavily on string parsing. I would prefer it to rely on the resource type solely. Maybe we could make use of Protocols here, similar to how it was done for campaigns, but I am not sure if that would make the code all over the place again.

backend/lib/edgehog/containers/provisioner/core.ex
```diff
@@ -0,0 +1,302 @@
...
+        end
+      end
+
+      def maybe_early_terminate(%{device_online?: device_online?} = state, next_step) do
+        if device_online? do
+          next_step
+        else
+          {:stop, {:shutdown, :device_offline}, state}
+        end
+      end
```
ArnelaL, line 147: Functions defined in Core module should be used to manipulate data and not to manage the lifecycle of GenServer process.

ArnelaL (changes requested):

backend/lib/edgehog/containers/provisioner.ex
Dam-99: Also agree on the string manipulation to extracting modules and atoms being kinda convoluted, and it would be a problem if we ever stray from our conventions in the future. For now however, I believe that the reduced boilerplate is worth it. I'll add a note to remind us of this point in the future.

About the Protocols: it was the first option that I considered, but after some thinking and irl discussion with luca, I decided to prefer Behaviours. At the end of the day, we just need for provisioner modules to adhere to a contract, but we don't care about how, and the way a protocol allows for a default implementation (aka delegates) kinda clashes with us wanting to override some function if needed. Also: from what I understood from [this thread on the difference between the two](https://groups.google.com/g/elixir-lang-talk/c/S0NlOoc4ThM/m/J2aD2hKrtuoJ) Josè Valim kinda makes a point that we should prefer Behaviours over Protocols wherever possible, which also guided this decision (feel free to tell me if I misinterpreted it though)

backend/lib/edgehog/containers/provisioner/core.ex
Dam-99: This is true. I was honestly just adapting the code that was already there, but we can address this on a separate PR

lusergit: Yeah since we're still in a feature branch I'd rather merge this change and then update the macro iteratively.

For the moment we're still merging code into a single place, but that's a change we have to do for sure

backend/lib/edgehog/containers/provisioner.ex
ArnelaL: We can stick with the Behaviours, but I’m wondering why we are pushing towards having a single Core module for all resources. Why not implement the behaviour separately for each resource? That would allow us to avoid relying so heavily on string manipulation and make the code more type oriented.

Also, I think all the `unquote`s currently in the Provisioner code should be removed by adding the necessary functions to the Core module. For example, the Core module could expose a function that returns the subscription topic for a given resource, or even handle the subscribing and unsubscribing itself.
This way, even if we decide to keep the string manipulation approach, we could simply pass the resource atom to the Core module from the Provisioner and move all of that string manipulation logic out of the Provisioner. This would keep the Provisioner code much cleaner and more maintainable.

This doesn't need to be implemented in this PR, I am saying this just so we can discuss it.

Dam-99: Yeah moving as much stuff as possible to the Core module sounds very reasonable, and we will likely refactor in the future. This PR is mostly for condensing the duplicated code into a single point

lusergit: I agree with you @ArnelaL :saluting_face: that's the goal for sure
```
resource
  +- resource.ex
  +- provisioner.ex # <- `use Provisioner`. gives deafult implementation of functions, the user can customize them
  +- provisioner
    +- core.ex # <- `use Provisioner.Core`, gives default implementation of functions and requires the user to provide others (where no default implementation can be provided)
```

I think this should be the goal structure to keep. This way we reliably understand what's going on

ArnelaL (approved):

lusergit (approved):

